PR #5, opened 2024-10-17
Rust duration type

adespawn:
This is an implementation of one of the datatype used by the database. The current approch keeps the internal data of the type in object created in rust, while all the processing of this object is done in JS part (it usses mostly unchanged datastax code).  Appart from diffrent behaviour in case of providing unexpected type in the constructor, everything should work the same way as in old implementation. If not, treat it as a bug.

This is a example implementation. The goal is to have some kind of example how to implement datastax custom types.
Current approach keeps the data in the rust object, but leaves all the processing that is done in the js part unchanged.
As this type is analog to ``CqlDuration`` used by the rust driver, there should be a way to create ``Duration`` object from ``CqlDuration``, which is implemented. 

Both comments about this specific implementation, and general approach are welcomed.

Tests expected to pass:
[There is new, temporary workflow with just the tests expected to pass]
``test/unit/duration-type-tests.js``

### DIFF
--- a/lib/new-utils.js
+++ b/lib/new-utils.js
@@ -1,0 +1,24 @@
+"use strict";
+
+const Long = require("long");
+
+/**
+ * Converts from bigint provided by napi into Long type used by the datastax library
+ * @param {bigint} from
+ * @returns {Long}
+ */
+function bigint_to_long(from) {
+  return Long.fromString(from.toString());
+}
+
+/**
+ * Converts from Long type used by the datastax library into bigint used by napi
+ * @param {Long} from
+ * @returns {bigint}
+ */
+function long_to_bigint(from) {
+  return BigInt(from.toString());
+}
+
+exports.bigint_to_long = bigint_to_long;
+exports.long_to_bigint = long_to_bigint;
--- a/lib/new-utils.js
+++ b/lib/new-utils.js
@@ -1,8 +1,10 @@
 "use strict";
 
 const Long = require("long");
+
+// maxInt value is based on how does Long splits values between internal high and low fields.
 const maxInt = BigInt(0x100000000);
-const minuOne = BigInt(-1);
+const minusOne = BigInt(-1);
 
 /**
  * Converts from bigint provided by napi into Long type used by the datastax library
--- a/lib/new-utils.js
+++ b/lib/new-utils.js
@@ -1,14 +1,21 @@
 "use strict";
 
 const Long = require("long");
+const maxInt = BigInt(0x100000000);
+const minuOne = BigInt(-1);
 
 /**
  * Converts from bigint provided by napi into Long type used by the datastax library
+ * BigInt is the way napi handles values too big for js Number type,
+ * while Long is the way datastax parts of the code handle 64-bit integers.
  * @param {bigint} from
  * @returns {Long}
  */
-function bigint_to_long(from) {
-  return Long.fromString(from.toString());
+function bigintToLong(from) {
+  let lo = from % maxInt;
+  let hi = from / maxInt;
+  if (lo < 0) hi += minusOne;
+  return Long.fromValue({ low: Number(lo), high: Number(hi), unsigned: false });
 }
 
 /**
@@ -16,9 +23,13 @@ function bigint_to_long(from) {
  * @param {Long} from
  * @returns {bigint}
  */
-function long_to_bigint(from) {
-  return BigInt(from.toString());
+function longToBigint(from) {
+  let lo = BigInt(from.low);
+  let hi = BigInt(from.high);
+  let r = lo + maxInt * hi;
+  if (lo < 0) r += maxInt;
+  return r;
 }
 
-exports.bigint_to_long = bigint_to_long;
-exports.long_to_bigint = long_to_bigint;
+exports.bigintToLong = bigintToLong;
+exports.longToBigint = longToBigint;
--- a/lib/types/duration.js
+++ b/lib/types/duration.js
@@ -3,10 +3,11 @@ const Long = require("long");
 const util = require("util");
 const utils = require("../utils");
 const rust = require("../../index");
-const { bigint_to_long, long_to_bigint } = require("../new-utils");
+const { bigintToLong, longToBigint } = require("../new-utils");
 
 /** @module types */
 
+// Reuse the same buffers that should perform slightly better than built-in buffer pool
 const reusableBuffers = {
   months: utils.allocBuffer(9),
   days: utils.allocBuffer(9),
@@ -44,7 +45,7 @@ function Duration(months, days, nanoseconds) {
     this.internal = rust.DurationWrapper.new(
       months,
       days,
-      long_to_bigint(nanoseconds),
+      longToBigint(nanoseconds),
     );
   else if (isNumber(nanoseconds))
     this.internal = rust.DurationWrapper.new(months, days, BigInt(nanoseconds));
@@ -70,7 +71,7 @@ Duration.prototype.equals = function (other) {
  * @returns {Buffer}
  */
 Duration.prototype.toBuffer = function () {
-  let nanoseconds = bigint_to_long(this.internal.getNanoseconds());
+  let nanoseconds = bigintToLong(this.internal.getNanoseconds());
   const lengthMonths = VIntCoding.writeVInt(
     Long.fromNumber(this.internal.months),
     reusableBuffers.months,
@@ -99,7 +100,7 @@ Duration.prototype.toBuffer = function () {
  * @return {string}
  */
 Duration.prototype.toString = function () {
-  let nanoseconds = bigint_to_long(this.internal.getNanoseconds());
+  let nanoseconds = bigintToLong(this.internal.getNanoseconds());
   let value = "";
   function append(dividend, divisor, unit) {
     if (dividend === 0 || dividend < divisor) {
@@ -146,7 +147,6 @@ Duration.prototype.toString = function () {
  * @returns {Duration}
  */
 Duration.fromBuffer = function (buffer) {
-  if ((!buffer) instanceof Buffer) throw new TypeError("Expected Buffer type");
   const offset = { value: 0 };
   const months = VIntCoding.readVInt(buffer, offset).toNumber();
   const days = VIntCoding.readVInt(buffer, offset).toNumber();
--- a/lib/types/duration.js
+++ b/lib/types/duration.js
@@ -30,172 +30,179 @@ const iso8601WeekRegex = /P(\d+)W/;
 const iso8601AlternateRegex =
   /P(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})/;
 
-/**
- * Creates a new instance of {@link Duration}.
- * @classdesc
- * Represents a duration. A duration stores separately months, days, and seconds due to the fact that the number of
- * days in a month varies, and a day can have 23 or 25 hours if a daylight saving is involved.
- * @param {Number} months The number of months.
- * @param {Number} days The number of days.
- * @param {Number|Long} nanoseconds The number of nanoseconds.
- * @constructor
- */
-function Duration(months, days, nanoseconds) {
-  if (nanoseconds instanceof Long)
-    this.internal = rust.DurationWrapper.new(
-      months,
-      days,
-      longToBigint(nanoseconds),
+class Duration {
+  /**
+   * Creates a new instance of {@link Duration}.
+   * @classdesc
+   * Represents a duration. A duration stores separately months, days, and seconds due to the fact that the number of
+   * days in a month varies, and a day can have 23 or 25 hours if a daylight saving is involved.
+   * @param {Number} months The number of months.
+   * @param {Number} days The number of days.
+   * @param {Number|Long} nanoseconds The number of nanoseconds.
+   * @constructor
+   */
+  constructor(months, days, nanoseconds) {
+    if (nanoseconds instanceof Long)
+      this.internal = rust.DurationWrapper.new(
+        months,
+        days,
+        longToBigint(nanoseconds),
+      );
+    else if (typeof nanoseconds === "number")
+      this.internal = rust.DurationWrapper.new(
+        months,
+        days,
+        BigInt(nanoseconds),
+      );
+    else
+      throw new TypeError(
+        `Invalid nanosecond argument type: ${typeof nanoseconds}`,
+      );
+  }
+
+  equals(other) {
+    if (!(other instanceof Duration)) {
+      return false;
+    }
+    return (
+      this.internal.months === other.internal.months &&
+      this.internal.days === other.internal.days &&
+      this.internal.getNanoseconds() === other.internal.getNanoseconds()
     );
-  else if (typeof nanoseconds === "number")
-    this.internal = rust.DurationWrapper.new(months, days, BigInt(nanoseconds));
-  else
-    throw new TypeError(
-      `Invalid nanosecond argument type: ${typeof nanoseconds}`,
+  }
+
+  /**
+   * Serializes the duration and returns the representation of the value in bytes.
+   * @returns {Buffer}
+   */
+  toBuffer() {
+    let nanoseconds = bigintToLong(this.internal.getNanoseconds());
+    const lengthMonths = VIntCoding.writeVInt(
+      Long.fromNumber(this.internal.months),
+      reusableBuffers.months,
     );
+    const lengthDays = VIntCoding.writeVInt(
+      Long.fromNumber(this.internal.days),
+      reusableBuffers.days,
+    );
+    const lengthNanoseconds = VIntCoding.writeVInt(
+      nanoseconds,
+      reusableBuffers.nanoseconds,
+    );
+    const buffer = utils.allocBufferUnsafe(
+      lengthMonths + lengthDays + lengthNanoseconds,
+    );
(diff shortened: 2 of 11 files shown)


lib/types/duration.js
```diff
@@ -406,7 +473,7 @@ Builder.prototype.addMillis = function(millis) {
  * @param {String|Long} micros
  * @return {Builder}
  */
-Builder.prototype.addMicros = function(micros) {
+Builder.prototype.addMicros = function (micros) {
   const value = typeof micros === 'string' ? Long.fromString(micros) : micros;
   this._validateOrder(9);
```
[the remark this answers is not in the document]
adespawn: @Stapox35 Added JS formatter in #30.

lib/types/duration.js
```diff
@@ -102,15 +96,15 @@ Duration.prototype.toString = function () {
     value += dividend.divide(divisor).toString() + unit;
     return dividend.modulo(divisor);
   }
-  if (this.months < 0 || this.days < 0 || this.nanoseconds.isNegative()) {
+  if (this.internal.months < 0 || this.internal.days < 0 || nanoseconds.isNegative()) {
```
wprzytula, line 99: why did `this.nanoseconds` change into `nanoseconds`? Are both equivalent in JS?

lib/types/duration.js
```diff
@@ -102,15 +96,15 @@ Duration.prototype.toString = function () {
...
   }
-  let remainder = append(Math.abs(this.months), monthsPerYear, "y");
+  let remainder = append(Math.abs(this.internal.months), monthsPerYear, "y");
   append(remainder, 1, "mo");
-  append(Math.abs(this.days), 1, "d");
+  append(Math.abs(this.internal.days), 1, "d");
 
-  if (!this.nanoseconds.equals(Long.ZERO)) {
-    const nanos = this.nanoseconds.isNegative() ? this.nanoseconds.negate() : this.nanoseconds;
+  if (!nanoseconds.equals(Long.ZERO)) {
```
wprzytula, line 106: ditto

lib/types/duration.js
```diff
@@ -127,6 +121,8 @@ Duration.prototype.toString = function () {
  * @returns {Duration}
  */
 Duration.fromBuffer = function (buffer) {
+  if (!buffer instanceof Buffer)
+    throw new TypeError('Expected Buffer type');
```
wprzytula, line 125: Is the goal of this an earlier incompatible type detection?

src/result.rs
```diff
@@ -181,6 +183,14 @@ impl CqlValueWrapper {
     }
   }
 
+  #[napi]
+  pub fn get_duration(&self) -> napi::Result<DurationWrapper> {
+    match self.internal.as_cql_duration() {
+      Some(r) => Ok(DurationWrapper::from_cql_duration(r)),
+      None => Err(Error::new(Status::GenericFailure, "Error")),
+    }
+  }
```
wprzytula, line 192: Of course the error message must be more informative. Maybe `Error: the underlying CQL value is not of Duration type`?

wprzytula: Also, please check which `Status` variant is the best suit here. Perhaps it's `GenericFailure` indeed, but let's make sure.

src/types/duration.rs
```diff
@@ -0,0 +1,59 @@
...
+impl DurationWrapper {
+  #[napi]
+  pub fn new(months: i32, days: i32, ns_bigint: BigInt) -> napi::Result<Self> {
+    let ns = ns_bigint.get_i64();
+    if !ns.1 {
+      return Err(Error::new(
+        Status::GenericFailure,
+        "Invalid use: Should not happen?",
+      ));
+    }
```
wprzytula, line 21: What can trigger this error? Please reflect the possible cause in the error message.

src/types/duration.rs
```diff
@@ -0,0 +1,59 @@
...
+  pub days: i32,
+  pub nanoseconds: i64,
+}
+
+#[napi]
+impl DurationWrapper {
+  #[napi]
+  pub fn new(months: i32, days: i32, ns_bigint: BigInt) -> napi::Result<Self> {
+    let ns = ns_bigint.get_i64();
+    if !ns.1 {
```
wprzytula, line 16: It's not clear at all what is `ns.1` and `ns.0`.

Let's aid readability by using deconstruction instead of `.1` access.
```suggestion
    let (_, ns) = ns_bigint.get_i64(); // Assuming that a two-element tuple is returned
    if ns {
```

wprzytula (changes requested):

src/result.rs
adespawn: Here is list of all possible ``Status`` values: [code](https://github.com/napi-rs/napi-rs/blob/de3a60dfc1e59b3a0a493cf8d955506f7ab39d86/crates/napi/src/status.rs#L7-L35) I can't find any appropriate documentation, but it seems all other types are for receiving invalid type of argument.

lib/types/duration.js
```diff
@@ -85,7 +78,8 @@ Duration.prototype.toBuffer = function () {
  * @return {string}
  */
 Duration.prototype.toString = function () {
-  let value = '';
+  let nanoseconds = bigint_to_long(this.internal.getNanoseconds());
+  let value = ''
```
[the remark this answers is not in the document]
adespawn: In most cases, in JS it doesn't matter if semicolon is present. For the purpose of clean code it's best to use them, and I just missed this one.

lib/types/duration.js
adespawn: this.internal.nanoseconds refers to the value in rust object. In JS it's of bigint type. This datastax code treats nanoseconds as Long, so we need to convert the type, which we keep in local variable nanoseconds.

lib/types/duration.js
adespawn: Sort of. But as this function is unchanged from original, it may make more sense to don't do it.

src/result.rs
```diff
@@ -171,47 +181,47 @@ impl CqlValueWrapper {
     // Is this correct?
     match self.internal.as_counter() {
       Some(r) => Ok(r.0.into()),
-      None => Err(Error::new(Status::GenericFailure, "Error")),
+      None => Err(Self::generic_error("counter")),
     }
```
wprzytula, line 185: Regarding the comment: what's the doubt here?

src/types/duration.rs
```diff
@@ -12,15 +12,15 @@ pub struct DurationWrapper {
...
-    let ns = ns_bigint.get_i64();
-    if !ns.1 {
+    let (ns_value,is_lossless) = ns_bigint.get_i64();
+    if !is_lossless {
       return Err(Error::new(
         Status::GenericFailure,
-        "Invalid use: Should not happen?",
+        "Nanoseconds cannot overflow i64",
       ));
     }
```
wprzytula, line 21: Where does this guarantee come from? Is it some kind of runtime checks (where?) that prevents values out of `i64` range to be passed here?
In general, let's include such explanation upon each situation that we assume something and fail if it's not true.

lib/new-utils.js
```diff
@@ -0,0 +1,25 @@
...
+
+
+/**
+ * Converts from bigint provided by napi into Long type used by the datastax library
+ * @param {bigint} from
+ * @returns {Long}
+ */
+function bigint_to_long(from){
+    return Long.fromString(from.toString());
+}
```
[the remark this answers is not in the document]
adespawn: Your concerns are justified. I updated the conversion a bit, but it also may be nice to test it a bit more when it comes to performance, depending how much this function will be used.

lib/new-utils.js
```diff
@@ -2,23 +2,34 @@
 
 const Long = require('long');
 
+const maxInt = BigInt(0x100000000);
+const minuOne = BigInt(-1);
```
wprzytula, line 6: minuOne -> minusOne

lib/new-utils.js
```diff
@@ -2,23 +2,34 @@
 
 const Long = require('long');
 
+const maxInt = BigInt(0x100000000);
```
wprzytula, line 5: Is it really the maximum possible Int? Isn't e.g. `0x100000001` possible?

lib/new-utils.js
```diff
@@ -2,23 +2,34 @@
...
  */
-function bigint_to_long(from){
-    return Long.fromString(from.toString());
+function bigint_to_long(from) {
+    let lo = from % maxInt;
+    let hi = from / maxInt;
+    if (lo < 0)
+        hi += minuOne
+    return Long.fromValue({ low: Number(lo), high: Number(hi), unsigned: false });
 }
```
wprzytula, line 19: Looks much better than strings :)

src/types/duration.rs
adespawn: This is more of an expectation, than a guarantee. In the JS version, the nanoseconds are kept in i64, and if there is bigger value provided, it quietly changes the value. (not even in c style: by taking modulo 2^k). So by throwing this error we change a behavior in (improper use / undefined situation) of Duration type

lib/new-utils.js
adespawn: This is reversed engineered from how the Long type keeps it's values split between high and low

lib/new-utils.js
adespawn: That took much longer than it looks like 😅

src/result.rs
adespawn: I wasn't able to find any usage of Counter type in the js driver. Probably will need to address, but maybe not in this PR

lib/new-utils.js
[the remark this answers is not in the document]
adespawn: Bigint is the way values unfit for js Number type are handled by napi and Long is the datastax way of handling 64 bit integers. 
Is this a clearer explanation than the current one?

src/types/duration.rs
wprzytula: So maybe `Nanoseconds must not overflow i64` would convey the message better? As it's our expectation/limitation that we impose on the input and we forbid the user to pass something different.

lib/new-utils.js
wprzytula: OK, good. Then please explain this in a comment.

lib/new-utils.js
wprzytula: That's perfectly understandable. Diffs rarely reflect the actual amount of work needed to get sth done.

lib/types/duration.js
```diff
@@ -37,40 +38,32 @@ const iso8601AlternateRegex = /P(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})/
...
-   * Gets the number of nanoseconds represented as a <code>int64</code>.
-   * @type {Long}
-   */
-  this.nanoseconds = typeof nanoseconds === 'number' ? Long.fromNumber(nanoseconds) : nanoseconds;
+  if (nanoseconds instanceof Long)
+    this.internal = rust.DurationWrapper.new(months, days, long_to_bigint(nanoseconds));
+  else if (isNumber(nanoseconds))
+    this.internal = rust.DurationWrapper.new(months, days, BigInt(nanoseconds));
+  else
+    throw new TypeError(`Invalid nanosecond argument type: ${typeof(nanoseconds)}`);
```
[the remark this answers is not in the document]
adespawn: If this constructor is used correctly (meaning nanoseconds are one of mentioned types), then yes.
If nanosecond is of a diffrent type, then my code fails immidietly, when the old code may partialy work, (ex. the duration type may be created, but will fail ((throw error)) when calling .toString() on it).
So this changes how the driver handles invalid usage. The correct usage remains the same.

lib/new-utils.js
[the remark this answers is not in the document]
wprzytula: Yes, this definitely adds the needed context. Thanks.

lib/types/duration.js
wprzytula: My bad, I didn't notice the variable. Nvm.

src/result.rs
wprzytula: So the doubt arises because there is no way to check the logic with the JS driver, because it's not triggered anywhere? Did I get it right?

adespawn: Sort of. It's possibile this type is not used by the DS library so we will be adding it instead of updating as f.ex. duration type

adespawn:
There was a lot ot comments, and unless I missed some, I responded to all of them.

lib/types/duration.js
```diff
@@ -161,6 +156,7 @@ Duration.fromBuffer = function (buffer) {
  * @returns {Duration}
  */
 Duration.fromString = function (input) {
+  // This does a lot of stuff. Is it nesesery to implement it in rust, or should we just keep it as is?
   const isNegative = input.charAt(0) === '-';
   const source = isNegative ? input.substr(1) : input;
   if (source.charAt(0) === 'P') {
```
wprzytula, line 162: Does the question still stand, or have we decided that we're keeping this logic in JS?

lib/types/result-set.js
```diff
@@ -148,6 +148,8 @@ function getCqlObject(field) {
       return field.getCounter();
     case rust.CqlTypes.Double:
       return field.getDouble();
+    case rust.CqlTypes.Duration:
+      return field.getDuration();
     case rust.CqlTypes.Float:
       return field.getFloat();
```
wprzytula, line 154: A fundamental question: why were other types already present here, and Duration was not? What makes Duration different in this context?

lib/types/duration.js
adespawn: The question still stands

lib/types/result-set.js
adespawn: Those types were added in branch ``types-support``

adespawn:
Brought the linear history to this branch!  No pesky history graphs will hide.

wprzytula:
But tests are failing... why?

adespawn:
Because of changes in simple query and types support, not all features of Client class and execute function are currently present, which fails examples.

lib/new-utils.js
```diff
@@ -0,0 +1,37 @@
...
+const minusOne = BigInt(-1);
+
+/**
+ * Converts from bigint provided by napi into Long type used by the datastax library
+ * BigInt is the way napi handles values too big for js Number type,
+ * while Long is the way datastax parts of the code handle 64-bit integers.
+ * @param {bigint} from
+ * @returns {Long}
+ */
+function bigint_to_long(from) {
```
ZuzaOsa, line 16: Shouldn't we keep the convention of the Datastax driver to write functions in camelCase?

wprzytula: We could as well use a different case to clearly distinguish Datastax procedures and entities from ours. This is an open discussion.

adespawn:
I've changed from prototypes into class as discussed today, and cleaned up the commit history a little bit. If there is no new comments, I will merge this branch.

lib/new-utils.js
```diff
@@ -0,0 +1,37 @@
...
+ * while Long is the way datastax parts of the code handle 64-bit integers.
+ * @param {bigint} from
+ * @returns {Long}
+ */
+function bigintToLong(from) {
+  let lo = from % maxInt;
+  let hi = from / maxInt;
+  if (lo < 0) hi += minusOne;
+  return Long.fromValue({ low: Number(lo), high: Number(hi), unsigned: false });
+}
```
wprzytula, line 21: We're using `Long` only because Datastax driver exposes it in its API, right? In the office we concluded that `BigInt` is available for us in the minimum supported NodeJS version, that's why I'm asking if Long is needed at all.

lib/types/duration.js
```diff
@@ -28,173 +30,179 @@ const iso8601WeekRegex = /P(\d+)W/;
...
-   * @type {Number}
-   */
-  this.days = days;
+  constructor(months, days, nanoseconds) {
+    if (nanoseconds instanceof Long)
+      this.internal = rust.DurationWrapper.new(
+        months,
+        days,
+        longToBigint(nanoseconds),
+      );
```
wprzytula, line 50: So this is an API-compatible modernisation of Datastax API as suggested by @Chlor87?

adespawn: Yes, it's cleaner way of declaring class which was suggested by Kuba.

lib/new-utils.js
adespawn: Not only because of that. Currently datastax in it's code uses Long for all internal i64 calculations, and changing that would require not insignificant amount of effort, thanks to functions like this: https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/6cc023e07142e8ea6c8d8a7d49b52a40d3784a75/lib/types/duration.js#L84-L87
that assume the long is provided. Also the Long has different semantic for even basic operations like shown here:
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/6cc023e07142e8ea6c8d8a7d49b52a40d3784a75/lib/types/duration.js#L119-L120

wprzytula: Understood. Then please create an issue about possible future replacement of Long with BigInt in the whole codebase, and how much effort it's going to be and why.

wprzytula (approved):

wprzytula:
Don't forget to remove the temporary commit with new CI workflow.

adespawn:
I thought it would be better to remove this step later, either in types support or simple query when the normal CI step would be working again

ZuzaOsa (approved):

adespawn:
As discussed today, I reverted this changes

PR #18, opened 2024-10-18
Add examples running to CI

adespawn:
ScyllaDB setup is based on rust driver


examples/DataStax/util.js
```diff
@@ -0,0 +1,8 @@
+function getClientArgs() {
+    return {
+        contactPoints: [process.env.SCYLLA_URI?.split(':')[0] ?? '172.17.0.2'],
+        localDataCenter: process.env.DATACENTER?.split(':')[0] ?? 'datacenter1'
```
wprzytula, line 4: Why is this split needed?

.github/workflows/CI.yml
```diff
@@ -76,6 +76,8 @@ jobs:
         shell: bash
       - name: Run tests
         run: npm test
+      - name: Run examples
+        run: npm run examples
```
wprzytula: IIUC, examples contain no assertions, so they don't test anything using assertions. What do we get by running examples in CI? We only (but it's still some benefit) know that they run without throwing exceptions, correct?

.github/workflows/CI.yml
```diff
@@ -77,7 +81,17 @@ jobs:
       - name: Run tests
         run: npm test
       - name: Run examples
-        run: npm run examples
+        run: SCYLLA_URI=172.42.0.2:9042 npm run examples
```
wprzytula: It is recommended to pass all 3 nodes as contact points to the driver, to make tests more realistic. This way it's always the first node who hosts the control connection; it's better to allow more varied scenarios.

.github/workflows/CI.yml
adespawn: That's correct.

.github/workflows/CI.yml
adespawn: Currently the examples don't use more than one node. I left creating three Scylla nodes as is in rust driver, as it is very likely we will add more examples/tests that use more than one node in the future.

wprzytula: What do you mean by using more nodes? Scylla, being a distributed system, uses all nodes transparently to serve requests.

wprzytula: Even if examples only provide one initial contact point (address of one node) to the driver, the driver soon recognizes all nodes and contacts them.

adespawn: Ok. I didn't know that. So then the current examples just uses single entrypoint. My goal with this PR was just to only add those examples to CI and we can add more examples showing things like multiple entry points alongside developing the driver.

examples/DataStax/README.md
```diff
@@ -1,11 +1,15 @@
 # DataStax Node.js Driver usage samples
 
-This examples are from https://github.com/datastax/nodejs-driver/tree/master repository
+This examples are from the [datastax driver repository](https://github.com/datastax/nodejs-driver/tree/master).
 This folder contains examples on how to use some features of the DataStax Node.js Driver.
 
+They assume you have a running a ScyllaDB / Cassandra at the default port.
+You can provide ip address of the database by setting ``SCYLLA_URI`` env variable. Any port provided alongside ip will be ignored.
```
wprzytula, line 7: Now I don't understand why we strip the port. It's part of configuration; Rust driver also expects `IP:PORT` format, but if PORT is not present, then default 9042 is used. IMO we should pass the port down the code, too.

adespawn: Previously, when I tested it, it didn't work if the port was provided. But I must have made a mistake, because it works now, so there is no need to remove the port.

adespawn: Previously, when I tested it, it didn't work if the port was provided. But I must have made a mistake, because it works now, so there is no need to remove the port.

examples/DataStax/util.js
```diff
@@ -0,0 +1,8 @@
+function getClientArgs() {
+    return {
+        contactPoints: [process.env.SCYLLA_URI ?? '172.17.0.2'],
```
wprzytula, line 3: last nit: let's add port here, too, so that a user gets a correct hint that it's expected to be passed.
```suggestion
        contactPoints: [process.env.SCYLLA_URI ?? '172.17.0.2:9042'],
```

wprzytula (changes requested):

wprzytula (approved):
LGTM, good job.

ZuzaOsa (approved):

PR #31, opened 2024-10-20
Mark not supported functions

ZuzaOsa:


lib/datastax/graph/structure.js
```diff
@@ -3,142 +3,45 @@
...
-  (function adaptProperties(self) {
-    if (properties) {
-      const keys = Object.keys(properties);
-      for (let i = 0; i < keys.length; i++) {
-        const k = keys[i];
-        self.properties[k] = properties[k].value;
-      }
-    }
-  })(this);
+  throw new ReferenceError(`This class (Edge) is not supported by our driver`);
```
wprzytula, line 27: One thought: If we'd like to change the error message at some point in the future, we will have to replace the string in all occurences. I suggest defining the message in one place and using it everywhere, so that changes in one place in code are necessary in such situation.

wprzytula: The text could be parametrised by kind of the entity (`class`, `function`, etc.) and name of the entity (`Edge`).

wprzytula (changes requested):

lib/utils.js
```diff
@@ -1033,6 +1033,16 @@ function whilst(condition, fn, callback) {
...
 
+/**
+ * 
+ * @param {String} entity 
+ * @param {String} name 
+ * @returns 
+ */
+function markAsNotSupported(entity, name) {
+  return "This ${entity} (${name}) is not supported by our driver";
+}
```
wprzytula, line 1044: nit: `This` -> `The`

wprzytula (approved):
One nit, otherwise LGTM

PR #34, opened 2024-10-22
Not supported cleanup

ZuzaOsa:
This PR makes graph and geometry functions that will not be supported by our driver throw a ReferenceError with a relevant error message. Code from index.d.ts in this modules was also deleted to cause TypeScript to throw compilation error if someone wants to use those functions. Datastax tests for graph and geometry methods were deleted.


lib/datastax/graph/result-set.js
```diff
@@ -1,141 +1,56 @@
...
- * @example
- * const vertex = result.first();
- * @param {ResultSet} result
- * @param {Function} [rowParser]
- * @alias module:datastax/graph~GraphResultSet
- * @constructor
+ * @deprecated
  */
 function GraphResultSet(result, rowParser) {
+  throw new ReferenceError(utils.markAsNotSupported("class", "GraphResultSet"));
```
adespawn, line 9: I think we also considered adding some way of stoping typescript compilation from finishing. Did you considered that / checked it would work as expected?

adespawn: (Work as expected: would trigger compilation error only if used)

lib/datastax/graph/structure.js
```diff
@@ -1,144 +1,47 @@
...
-  /**
-   * Gets the element id.
-   */
-  this.id = id;
-  /**
-   * Gets the element label.
-   * @type {String}
-   */
-  this.label = label;
+  throw new ReferenceError(utils.markAsNotSupported("class", "Element"));
```
adespawn, line 9: Style difference. Sometimes you use ``,`` and split it to 3 lines and sometimes you keep it in one line.

lib/utils.js
```diff
@@ -1033,6 +1033,16 @@ function whilst(condition, fn, callback) {
...
 }
 
+/**
+ *
+ * @param {String} entity
+ * @param {String} name
+ * @returns
+ */
+function markAsNotSupported(entity, name) {
+  return "The ${entity} (${name}) is not supported by our driver";
```
adespawn, line 1043: Why not just throw error here?

lib/utils.js
```diff
@@ -1033,6 +1033,16 @@ function whilst(condition, fn, callback) {
   }
 }
 
+/**
+ *
+ * @param {String} entity
+ * @param {String} name
+ * @returns
+ */
+function markAsNotSupported(entity, name) {
```
adespawn, line 1042: Maybe consider moving this to ``new-utils.js`` like in [duration PR](https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/f7fd151c5e44dc836a17a15030e37dddcfd54b3a/lib/new-utils.js), to differentiate between new and old code? Later we could merge those two files.

lib/datastax/graph/result-set.js
```diff
@@ -1,141 +1,56 @@
...
- * for (let vertex of result} { ... }
- * @example
- * const arr = result.toArray();
- * @example
- * const vertex = result.first();
- * @param {ResultSet} result
- * @param {Function} [rowParser]
- * @alias module:datastax/graph~GraphResultSet
- * @constructor
+ * @deprecated
```
adespawn, line 6: Consider adding some message to the deprecated tag.

adespawn (changes requested):

lib/datastax/graph/result-set.js
wprzytula: Right, I forgot about this requirement. Let's ensure it's true, if it's doable.

lib/utils.js
wprzytula: I like the idea, it's going to bring more clarity to the code. I think _later_ in this context is going to be quite distant future :)

wprzytula:
I can see that tests are failing, which makes sense because we deleted the tested functions. Two things:
1. Let's take a look at https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/actions/runs/11462837301/job/31895202380?pr=34 and verify that what we is what would like the user to see.
2. Let's disable (delete) the tests related to the functionality removed in this PR, because we want the CI to pass for this PR and subsequent ones. After that, that's the point of having CI: its failures should be meaningful, signalling that's something's wrong with the PR.

lib/datastax/graph/structure.js
ZuzaOsa: This is actually Prettier thing, maybe this should be marked as an issue

lib/datastax/graph/result-set.js
```diff
@@ -1,141 +1,56 @@
...
-    }
-  };
-}
-
-if (typeof Symbol !== "undefined" && typeof Symbol.iterator === "symbol") {
-  // Make iterable
-  GraphResultSet.prototype[Symbol.iterator] = function getIterator() {
-    return this.values();
+    throw new ReferenceError(
+      utils.markAsNotSupported("function", "GraphResultSet.getTraversers"),
```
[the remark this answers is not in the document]
ZuzaOsa: Same here, Prettier thing.

lib/datastax/graph/structure.js
adespawn: Yes, but sometimes you added ``,`` and than it splits into 3 lines, and when you don't it keeps it as one line

adespawn: I see... @Stapox35 Can you look into it?

lib/new_utils.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @param {String} entity
+ * @param {String} name
+ */
+function markAsNotSupported(entity, name) {
+  throw new ReferenceError(
+    "The ${entity} (${name}) is not supported by our driver",
+  );
+}
```
wprzytula, line 9: Now the semantics of this function would be better described by the name: `throw[Error]NotSupported[Entity]`, where chunks in brackets are optional IMO.

lib/new_utils.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @param {String} entity
+ * @param {String} name
+ */
+function markAsNotSupported(entity, name) {
+  throw new ReferenceError(
+    "The ${entity} (${name}) is not supported by our driver",
```
adespawn, line 7: If you want the error to contain value of entity and name I would recommend using ` instead of " ;)

adespawn (changes requested):

lib/new_utils.js
wprzytula: This proves that we need at least a few unit tests that check behaviour of those nonsupported functions when called.

test/unit/geometry/line-string-tests.js
```diff
@@ -1,166 +1,13 @@
 "use strict";
 const assert = require("assert");
 const helper = require("../../test-helper");
-const utils = require("../../../lib/utils");
-const Point = require("../../../lib/geometry/point");
 const moduleName = "../../../lib/geometry/line-string";
```
adespawn, line 4: Do tests use helper?

test/unit/graph/graph-result-tests.js
```diff
@@ -1,203 +1,12 @@
...
-});
-
-/**
- * @param {Array} rows
- * @returns {ResultSet}
- */
-function getResultSet(rows) {
-  return new ResultSet({ rows: rows }, null, null, null);
-}
+});
```
adespawn, line 12: No new line

test/unit/geometry/point-tests.js
```diff
@@ -1,122 +1,13 @@
 "use strict";
 const assert = require("assert");
 const helper = require("../../test-helper");
```
adespawn, line 3: Do tests use helper?

test/unit/graph/graph-result-tests.js
```diff
@@ -1,203 +1,12 @@
 "use strict";
 
 const assert = require("assert");
-const utils = require("../../../lib/utils");
-const ResultSet = require("../../../lib/types/result-set");
 const GraphResultSet = require("../../../lib/datastax/graph/result-set");
```
adespawn, line 4: consistency: sometimes test use ``moduleName``, sometimes include modules directly

adespawn (changes requested):

test/unit/geometry/polygon-tests.js
```diff
@@ -1,218 +1,13 @@
...
-      ].forEach(function (item) {
-        assert.throws(function () {
-          Polygon.fromString(item);
-        }, TypeError);
-      });
-    });
+  it("tests not supported error", function () {
+    assert.throws(function () {
+      new Polygon();
+    }, ReferenceError);
```
wprzytula, line 9: These tests are OK, but I would check at least in one unit test that the output string of the `ReferenceError` is equal to its expected value.

adespawn:
Remember to update this readme: 
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/c42672cd2856672dfe003ed9a6bd59f9f5eb4c79/README.md?plain=1#L64-L73

adespawn:
Remember to rebase onto main ;)
Also maybe consider to move all files to some compatibility catalog (what Kuba suggested today)?
Do you want to squash whole PR, or to edit the history of this PR?

adespawn (dismissed):

README.md
```diff
@@ -56,18 +56,24 @@ Files in .gitignore (auto-generated)
...
 
 ```js
-function function_name()) {
+function functionName()) {
     throw new Error(`TODO: Not implemented`);
 }
 ```
 
-And for functions that will be not implemented:
+And for functions or classes that will not be implemented:
```
wprzytula, line 64: Are all (at present moment and in the future) unimplemented entities classes or function? Are there some other entities?

README.md
```diff
@@ -56,18 +56,24 @@ Files in .gitignore (auto-generated)
...
-  throw new ReferenceError(`This function is not supported by our driver`);
+function functionName() {
+  throw new ReferenceError(`The function (functionName) is not supported by our driver`);
 }
+
+/**
+ * @deprecated Not supported by the driver. Usage will throw an error.
+ */
+function ClassName() {
+  throw new ReferenceError(`The class (ClassName) is not supported by our driver`);
```
wprzytula, line 78: Are classes in JS functions, too??

lib/client.js
```diff
@@ -603,53 +601,15 @@ Client.prototype.execute = function (query, params, options, callback) {
 };
 
 /**
- * Executes a graph query.
- * <p>It returns a <code>Promise</code> when a <code>callback</code> is not provided.</p>
- * @param {String} query The gremlin query.
```
wprzytula, line 608: What is gremlin??? Nvm, I don't have to know that.

lib/client.js
```diff
@@ -603,53 +601,15 @@ Client.prototype.execute = function (query, params, options, callback) {
...
-    if (typeof callback === "function") {
-        parameters = typeof parameters !== "function" ? parameters : null;
-        return promiseUtils.toCallback(
-            this._graphExecutor.send(query, parameters, options),
-            callback,
-        );
-    }
-
-    return this._graphExecutor.send(query, parameters, options);
+    throwNotSupported("function", "Client.executeGraph");
```
wprzytula, line 612: I suggest another layer of helper functions:
`throwNotSupportedFunction(functionName)` := `throwNotSupportedEntity("function", functionName)`
`throwNotSupportedClass(className)` := `throwNotSupportedEntity("class", className)`
This way we type less for each call, leaving less space for typos.

lib/new_utils.js
```diff
@@ -0,0 +1,13 @@
...
+ * @param {String} entity
+ * @param {String} name
+ */
+function throwNotSupported(entity, name) {
+    throw new ReferenceError(
+        `The ${entity} (${name}) is not supported by our driver`,
+    );
+}
+
+exports.throwNotSupported = throwNotSupported;
```
wprzytula, line 13: As suggested, we could export only concrete variants:
`throwNotSupportedFunction`, `throwNotSupportedClass`, etc.

wprzytula (changes requested):

lib/client.js
wprzytula: The change should be easy to be done in all places by Search + Replace All.
- `throwNotSupported("function", ` -> `throwNotSupportedFunction(`

README.md
ZuzaOsa: I deleted the entity definition. It can be derived from the naming convention in JS (camelCase for functions and PascalCase for classes).

README.md
ZuzaOsa: In the older versions yes, now changed to new syntax.

README.md
```diff
@@ -61,19 +61,13 @@ function functionName()) {
...
+And for entities that will not be implemented:
 
 ```js
 /**
  * @deprecated Not supported by the driver. Usage will throw an error.
  */
-function functionName() {
-  throw new ReferenceError(`The function (functionName) is not supported by our driver`);
+function entityName() {
+  throw new ReferenceError(`entityName is not supported by our driver`);
```
adespawn, line 71: Why don't show ``throwNotSupported()`` here?

ZuzaOsa: I think that in README we want to show how it is handled and not overcomplicate it with driver logic.

lib/client.js
wprzytula: This is no longer relevant.

wprzytula (approved):

adespawn (approved):

PR #37, opened 2024-10-23
Update result set

adespawn:
Added some metadata to ResultSet constructor (info field). Some parameters still need to be added (Issue #46). 

Currently this PR allows fot use of the old constructor, which was part of the datastax driver. Although no one should construct this object themselves, it's left to allows running of the old unit tests for the result set, which are still applicable to the new driver. 

Currently I don't see a way of creating unit tests fot the napi layer as the rust driver doesn't allow creation of the QueryResult object. 

Although this changes may be affected by the new rust driver API, this PR focuses mostly on the JS side of the things and is required for one of the examples to run correctly. 

TODO introduced in this PR:
```
```


lib/types/result-set.js
```diff
@@ -140,8 +46,89 @@ function ResultSet(result) {
...
+   * </p>
+   * <p>
+   *   Note that the schema agreement check is only performed for schema-altering queries For other
+   *   query types, this method will always return <code>true</code>. If this method returns <code>false</code>,
+   *   clients can call [Metadata.checkSchemaAgreement()]{@link module:metadata~Metadata#checkSchemaAgreement} later to
+   *   perform the check manually.
+   * </p>
+   */
+  this.info = {
+    queriedHost: undefined, // Is supported by rust driver?
```
wprzytula, line 107: Not yet; there's an issue for that: https://github.com/scylladb/scylla-rust-driver/issues/1030

lib/types/result-set.js
```diff
@@ -140,8 +46,89 @@ function ResultSet(result) {
...
+    triedHosts: undefined, // Is supported by rust driver?
+    speculativeExecutions: undefined, // Is supported by rust driver?
```
wprzytula, line 109: These are not collected by default, but could be acquired by use of a custom [`HistoryListener`](https://docs.rs/scylla/latest/scylla/history/trait.HistoryListener.html).

lib/types/result-set.js
```diff
@@ -140,8 +46,89 @@ function ResultSet(result) {
...
+    achievedConsistency: undefined, // Is supported by rust driver?
+    traceId: result.getTraceId(),
+    warnings: result.getWarnings(),
+    customPayload: null, // supported by rust driver?
```
wprzytula, line 113: Rust driver does recognize custom payload, but it does not expose it to its public API. We would rather not expose this, because we believe that custom payload is intended for cluster-driver private communication purposes.

lib/types/result-set.js
```diff
@@ -140,8 +46,89 @@ function ResultSet(result) {
...
+   *   query types, this method will always return <code>true</code>. If this method returns <code>false</code>,
+   *   clients can call [Metadata.checkSchemaAgreement()]{@link module:metadata~Metadata#checkSchemaAgreement} later to
+   *   perform the check manually.
+   * </p>
+   */
+  this.info = {
+    queriedHost: undefined, // Is supported by rust driver?
+    triedHosts: undefined, // Is supported by rust driver?
+    speculativeExecutions: undefined, // Is supported by rust driver?
+    achievedConsistency: undefined, // Is supported by rust driver?
```
wprzytula, line 110: At the moment I don't think it's available anywhere in the Rust driver.

wprzytula: When using any policy different than `DowngradingConsistencyRetryPolicy`, the fact that the `ResultSet` was constructed should imply that the requested consistency was met. I.e., we could manually set here the consistency requested by the user.

When using `DowngradingConsistencyRetryPolicy`, however, the consistency can be lower than requested. For that, I think that `HistoryListener::log_attempt_start` could accept current query attempt's consistency as the second parameter. Then, the final consistency could be retrieved from `HistoryListener`.

wprzytula: To sum up, `DowngradingConsistencyRetryPolicy` complicates things. Let's check if Datastax JS driver has such policy; if not, we can do without this complication.

lib/types/result-set.js
```diff
@@ -140,8 +46,89 @@ function ResultSet(result) {
...
+    traceId: result.getTraceId(),
+    warnings: result.getWarnings(),
+    customPayload: null, // supported by rust driver?
+    isSchemaInAgreement: false, // Is supported by rust driver?
```
wprzytula, line 114: Rust driver by default automatically awaits schema agreement if needed (if the executed statement introduced some schema change), so when the `QueryResult` is returned, it means that the schema already is in agreement. See `Session::handle_auto_await_schema_agreement` in Rust driver's code (it's not `pub`, so it's not visible in the documentation).

Let's investigate how Datastax JS driver handles schema changes. From existence of this boolean field, I suppose that it does not auto-await schema agreement, and then the default auto-awaiting behavior of Rust driver does not suit our use case.

lib/types/result-set.js
```diff
@@ -156,6 +143,7 @@ ResultSet.prototype.first = function () {
 
 ResultSet.prototype.getPageState = function () {
   // backward-compatibility
+  // Do we want to keep backward-compatibility things for js-driver?
   return this.pageState;
 };
```
wprzytula, line 148: No, we can happily delete such legacy stuff :rocket:

src/result.rs
```diff
@@ -81,6 +89,35 @@ impl QueryResultWrapper {
...
+  #[napi]
+  pub fn get_columns_specs(&self) -> Vec<MetaColumnsWrapper> {
+    self
+      .internal
+      .col_specs()
+      .iter()
+      .map(|f| {
+        MetaColumnsWrapper {
+          // ToDo: Check if rust provides such information
+          ksname: "unknown?".to_string(),
```
wprzytula, line 106: `f.table_spec.keyspace_name().to_owned()`

wprzytula: I'm against using `to_string()` for `&str` -> `String` conversions. `to_string()` is too powerful and it suggests some stringification, such as `u32` -> `String`.

src/result.rs
```diff
@@ -81,6 +89,35 @@ impl QueryResultWrapper {
...
+  #[napi]
+  pub fn get_warnings(&self) -> Vec<String> {
+    self.internal.warnings.clone()
+  }
+
+  #[napi]
+  // As we don't have UUID type, we temporary use string
+  pub fn get_trace_id(&self) -> Option<String> {
+    self.internal.tracing_id.map(|f| f.to_string())
+  }
```
wprzytula, line 124: I'm not happy with this...
But it's not very important. Please add `// FIXME: ` here, though.

lib/types/results-wrapper.js
```diff
@@ -71,5 +71,27 @@ function getRowsFromResultsWrapper(result) {
...
+function getColumnsMetadata(result) {
+  let res = [];
+  let columnsWrapper = result.getColumnsSpecs();
+  for (let i = 0; i < columnsWrapper.length; i++) {
+    let e = columnsWrapper[i];
+    res.push({
+      ksname: e.ksname,
+      tablename: e.tablename,
+      name: e.name,
+      // ToDo: setup type and info
```
wprzytula: Nit: `TODO` stands out more than `ToDo`, that's why recommend it more.

src/result.rs
wprzytula, line 106: Keep in mind that once the ongoing deserialization refactor is done in Rust driver (in the next release, possibly next Thursday), `QueryResult` and its API is going to change much.

wprzytula (commented):
I suggest opening an issue for each key of the `info` map, because filling them with proper values is going to be a considerable effort, worth a separate PR each IMO.
Such issue should have my comments copied/quoted as an entrypoint to the discussion.

index.d.ts
```diff
@@ -5,41 +5,44 @@
...
-  Uuid = 24,
-  Varint = 25
+  Duration = 21,
+  List = 32,
+  Map = 33,
+  Set = 34,
+  UserDefinedType = 48,
+  Tuple = 49,
+  Empty = 50
 }
```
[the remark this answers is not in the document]
adespawn: Those values are based on: 
https://github.com/scylladb/scylla-rust-driver/blob/4649acbd9f63c0fcd0fc2e52d66fcbd253a3110f/scylla-cql/src/frame/response/result.rs#L629-L652
Although the current state of this PR doesn't use any particular mappings to the integer in this enum, it may become useful later as the datastax driver uses the same values for corresponding types: https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/6d1c8208910f942982b58d8642079d7deb485e4d/lib/types/index.d.ts#L22-L50

Pleas note, that values in this file you mentioned are auto-generated from https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/1a90383c901f3da8ea9110aef55cb7157febbd86/src/result.rs#L31-L59

index.d.ts
```diff
@@ -5,41 +5,44 @@
...
-  Text = 12,
-  Timestamp = 13,
-  Inet = 14,
-  List = 15,
-  Map = 16,
-  Set = 17,
-  UserDefinedType = 18,
+  Ascii = 1,
+  BigInt = 2,
+  Boolean = 3,
```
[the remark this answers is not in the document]
adespawn: I don't think I understand the question?

src/result.rs
```diff
@@ -100,14 +100,11 @@ impl QueryResultWrapper {
...
-        MetaColumnsWrapper {
-          // ToDo: Check if rust provides such information
-          ksname: "unknown?".to_string(),
-          tablename: f.table_spec.table_name().to_string(),
-          name: f.name.clone(),
-          type_code: map_column_type_to_cql_value(&f.typ),
-        }
+      .map(|f| MetaColumnsWrapper {
+        ksname: f.table_spec.ks_name().to_owned(),
+        tablename: f.table_spec.table_name().to_string(),
```
wprzytula, line 105: `to_string()` -> `to_owned()

src/result.rs
```diff
@@ -120,6 +117,7 @@ impl QueryResultWrapper {
   #[napi]
   // As we don't have UUID type, we temporary use string
   pub fn get_trace_id(&self) -> Option<String> {
+    // TODO, FIXME when UUID type is implemented
```
wprzytula, line 120: The proper format is:
`// FIXME: stop using strings when UUID type is implemented`

wprzytula: Dedicated tools recognize `FIXME:` and `TODO:` parts, detect them and even provide syntax colouring.

Stapox35 (approved):

adespawn:
Added issue #46 to fill the missing resultSet info fields based on @wprzytula suggestions.

src/result.rs
```diff
@@ -90,6 +94,32 @@ impl QueryResultWrapper {
...
     }
+
+    #[napi]
+    pub fn get_columns_specs(&self) -> Vec<MetaColumnsWrapper> {
+        self.internal
+            .col_specs()
+            .iter()
+            .map(|f| MetaColumnsWrapper {
+                ksname: f.table_spec.ks_name().to_owned(),
+                tablename: f.table_spec.table_name().to_string(),
```
wprzytula, line 105: inconsistency: `to_owned` and `to_string`.

src/result.rs
```diff
@@ -248,3 +278,34 @@ impl CqlValueWrapper {
         }
     }
 }
+
+fn map_column_type_to_cql_value(typ: &ColumnType) -> CqlType {
```
wprzytula, line 282: The name is very incorrect. `CqlValue` is an enum that can contain an arbitrary CQL value.

wprzytula:
```suggestion
fn map_column_type_to_cql_type(typ: &ColumnType) -> CqlType {
```

lib/types/result-set.js
```diff
@@ -10,30 +10,76 @@ const asyncIteratorSymbol = Symbol.asyncIterator || "@@asyncIterator";
...
+        triedHosts,
+        speculativeExecutions,
+        consistency,
+        isSchemaInAgreement,
+    ) {
+        if (!(result instanceof rust.QueryResultWrapper)) {
+            console.warn(
+                "Using legacy version of the ResultSet constructor.\n" +
+                    "This is deprecated and may be removed at any moment.",
+            );
```
wprzytula, line 42: Isn't there some customary way of issuing deprecation warnings in JS? In Rust, you would just use `#[deprecated]` attributes.

wprzytula: (I'm OK with this way, just wondering if it's the usual way of doing this)

lib/types/result-set.js
```diff
@@ -10,30 +10,76 @@ const asyncIteratorSymbol = Symbol.asyncIterator || "@@asyncIterator";
...
+            if (speculativeExecutions === -1) {
+                speculativeExecutions = 0;
+            }
```
wprzytula, line 45: Hmm? This is _a bit_ suspicious...
Absolutely explain this in a comment.

lib/types/result-set.js
```diff
@@ -10,30 +10,76 @@ const asyncIteratorSymbol = Symbol.asyncIterator || "@@asyncIterator";
...
-    constructor(result) {
+    constructor(
+        result,
+        host,
+        triedHosts,
+        speculativeExecutions,
+        consistency,
+        isSchemaInAgreement,
+    ) {
+        if (!(result instanceof rust.QueryResultWrapper)) {
```
wprzytula, line 38: So you say it's an old constructor (with old code) simply re-introduced?

adespawn: wrapped in the if statement with removed comments (those comments are in the "new" path of the constructor)

lib/types/result-set.js
adespawn: That's the old logic. It was commented with ``// if no execution was made at all, set to 0.`` so I can add this back.

lib/types/result-set.js
adespawn: I'm not aware of dedicated depreciation warning in JS.

lib/types/result-set.js
```diff
@@ -4,132 +4,83 @@ const utils = require("../utils");
 const errors = require("../errors");
 const rust = require("../../index");
 const Row = require("./row");
-const { getRowsFromResultsWrapper } = require("./results-wrapper");
+const resWrapper = require("./results-wrapper");
```
ZuzaOsa, line 7: Why ``resWrapper`` instead of ``resultsWrapper``?

src/result.rs
```diff
@@ -86,6 +98,32 @@ impl QueryResultWrapper {
...
+
+    #[napi]
+    pub fn get_warnings(&self) -> Vec<String> {
+        self.internal.warnings.clone()
+    }
+
+    #[napi]
+    // As we don't have UUID type, we temporary use string
+    pub fn get_trace_id(&self) -> Option<String> {
+        // TODO, FIXME when UUID type is implemented
```
ZuzaOsa, line 124: UUID is now implemented so this TODO can be fixed.

ZuzaOsa (changes requested):

ZuzaOsa (approved):

PR #38, opened 2024-10-24
Fix prettier for JS and Rust. 

Stapox35:


.prettierrc
```diff
@@ -1 +1,3 @@
-{}
+{
+    "tabWidth": 4
+}
```
adespawn, line 3: No new line

adespawn (changes requested):

wprzytula (approved):

PR #39, opened 2024-10-24
Inet type

PiotrJunior:


wprzytula:
Woho, this seems to need a rebase on main.

PiotrJunior:
Now InetAddress consists only from InetAddressWrapper which is a class in Rust. InetAddress have getters which calls getters in InetAddressWrapper which acquires information from IpAddr (same type as CqlValue::Inet). All other logic is handled currently in JS

wprzytula:
> Now InetAddress consists only from InetAddressWrapper which is a class in Rust

There are no classes in Rust :wink:

adespawn:
Also remember to fix integration tests for inet and move them to correct location

lib/types/inet-address.js
```diff
@@ -10,238 +11,243 @@ const utils = require("../utils");
...
-    );
-};
-
-/**
- * Returns the underlying buffer
- * @returns {Buffer}
- */
-InetAddress.prototype.getBuffer = function () {
-    return this.buffer;
-};
```
wprzytula, line 134: Transformation from functions/prototypes into proper classes should be done in a separate, preliminary commit, to reduce reviewers' confusion.

wprzytula (changes requested):

src/types/inet.rs
```diff
@@ -0,0 +1,70 @@
+use napi::{bindgen_prelude::Buffer, Error, Status};
```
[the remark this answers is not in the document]
wprzytula: This is potentially important for all of you. @adespawn @PiotrJunior @Stapox35 @ZuzaOsa

src/types/inet.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub fn from_cql_value(value: CqlValue) -> Result<Self, Error> {
+        match value {
+            CqlValue::Inet(inet) => Ok(Self { inet }),
+            _ => Err(Error::new(
+                Status::GenericFailure,
+                "Invalid CqlValue type for InetAddress",
+            )),
+        }
+    }
+}
```
adespawn, line 70: Again, consistency: lack of getter 
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/03317eeda2165e200cb006cd772ce8178ffb0b5d/src/types/duration.rs#L52

src/types/inet.rs
```diff
@@ -0,0 +1,70 @@
...
+        let buffer = match self.inet {
+            IpAddr::V4(ip) => ip.octets().to_vec(),
+            IpAddr::V6(ip) => ip.octets().to_vec(),
+        };
+        Ok(Buffer::from(buffer))
+    }
+}
+
+impl InetAddressWrapper {
+    pub fn from_cql_value(value: CqlValue) -> Result<Self, Error> {
```
adespawn, line 61: In my implementation of Duration wrapper I have function crating wrapper directly form intended type, not CqlValue enum. Make it consistent, I would recommend it doing my way. But if you think your way, try to convince me why it’s better.
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/03317eeda2165e200cb006cd772ce8178ffb0b5d/src/types/duration.rs#L45

adespawn (changes requested):
Remember to also add appropriate getter here: https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/03317eeda2165e200cb006cd772ce8178ffb0b5d/src/result.rs#L110

lib/types/inet-address.js
```diff
@@ -10,238 +11,243 @@ const utils = require("../utils");
...
+     * @param {InetAddress} other
+     * @returns {Boolean}
+     */
+    equals(other) {
+        if (!(other instanceof InetAddress)) {
+            return false;
+        }
+        return (
+            this.buffer.length === other.buffer.length &&
+            this.buffer.toString("hex") === other.buffer.toString("hex")
```
ZuzaOsa, line 123: You are comparing attributes that do not exist anymore.

ZuzaOsa (changes requested):

lib/types/inet-address.js
wprzytula: IIUC, this is exactly what we want! We emulate reading attributes by use of `get`ters. Thanks to that, the API is unchanged wrt attribute reads; only writes are no longer valid. BTW, what happens if an emulated attribute (i.e., a name with a `get`ter but without a `set`ter) is written to?

lib/types/inet-address.js
```diff
@@ -10,238 +11,243 @@ const utils = require("../utils");
...
     }
 
-    /**
-     * Immutable buffer that represents the IP address
-     * @type Array
-     */
-    this.buffer = buffer;
-
     /**
      * Returns the length of the underlying buffer
```
adespawn, line 20: Maybe explicitly say it's read only?
From what I understand there is nothing stoping in the old driver from editing internal fields (though it may generate UB?).

lib/types/inet-address.js
```diff
@@ -10,238 +11,243 @@ const utils = require("../utils");
...
-            this.buffer[2] +
-            "." +
-            this.buffer[3]
-        );
+    /**
+     * Provide the name of the constructor and the string representation
+     * @returns {string}
+     */
+    inspect() {
+        return this.constructor.name + ": " + this.toString();
```
adespawn, line 140:
```js
`${this.constructor.name}: ${this.toString()}`
```
instead of 
```js
this.constructor.name + ": " + this.toString()
```
It may be personal preference, but maybe consider switching to this way of creating strings?
@wprzytula @ZuzaOsa @Stapox35 @PiotrJunior

lib/types/inet-address.js
```diff
@@ -10,238 +11,243 @@ const utils = require("../utils");
...
-
-    for (let i = 0; i < buffer.length - 6; i++) {
-        if (buffer[i] !== 0) {
-            return false;
+    /**
+     * Validates for a IPv4-Mapped IPv6 according to https://tools.ietf.org/html/rfc4291#section-2.5.5
+     * @private
+     * @param {Buffer} buffer
+     */
+    static isValidIPv4Mapped(buffer) {
```
adespawn, line 238: This function wasn't exposed in the old api. Is this a mistake or deliberate change to add it to inet class?

adespawn (changes requested):

lib/types/inet-address.js
wprzytula: Definitely, let's use formatting syntax instead of concatenation.

lib/types/inet-address.js
```diff
@@ -10,238 +11,243 @@ const utils = require("../utils");
...
-    this.buffer = buffer;
-
     /**
      * Returns the length of the underlying buffer
      * @type Number
      */
-    this.length = buffer.length;
+    get length() {
+        return this._core.getLength();
+    }
```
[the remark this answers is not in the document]
PiotrJunior: This code won't throw any errors, but also it won't change value of length. I think length shouldn't be changed by user, so it's intended behavior.

lib/types/inet-address.js
PiotrJunior: Nothing happens, no error thrown, no value changed

src/types/inet.rs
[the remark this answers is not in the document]
PiotrJunior: What if methods from Inet will be called inside async funtions. Does this count as synchronous context for Inet method? @wprzytula @Chlor87

lib/types/inet-address.js
PiotrJunior: From Datastax API:
> buffer: Immutable buffer that represents the IP address
So by exposing only getter I assure immutability
But I will add @readonly to readonly properties

lib/types/inet-address.js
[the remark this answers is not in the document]
wprzytula: > This code won't throw any errors, but also it won't change value of length.

Wow, really?! Silent errors in JS???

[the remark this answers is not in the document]
wprzytula: Any sources of this?

src/types/inet.rs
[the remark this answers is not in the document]
wprzytula: Oh, I assumed async/non-async context in Rust; forgot that JS has them, too. Then @Chlor87's advice is indeed needed here.

src/types/inet.rs
```diff
@@ -0,0 +1,70 @@
...
+            ip.copy_from_slice(&buffer);
+            Ok(Self {
+                inet: IpAddr::from(ip),
+            })
+        } else {
+            return Err(Error::new(
+                Status::GenericFailure,
+                "Invalid IP address length",
+            ));
+        }
```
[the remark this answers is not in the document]
wprzytula: Right, I feared this, too :smile:
So another suggestion:
instead of `if 4 == len then copy_from_slice else if 16 == len then ...`
let's use `TryFrom<[u8]> for [u8; n]` impl:
```rust
        if let Ok(arr) = [u8; 4]::try_from(buffer) {
            Ok(Self {
                inet: IpAddr::from(ip),
            })
        } else if let Ok(arr) = [u8; 16]::try_from(buffer) {
            Ok(Self {
                inet: IpAddr::from(ip),
            })
        } else {
            Err(Error::new(
                Status::GenericFailure,
                "Invalid IP address length",
            ));
        }

File: lib/types/inet-address.js (hunk 9 above)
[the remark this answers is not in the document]
Comment (Chlor87) [reply lib/types/inet-address.js:25]: You can use private field for core I think:

```js
class InetAddress {
  #core/* = default value */

  constructor(buffer) {
      this.#core = rust.InetAddressWrapper.new(buffer)
  }

  get length() {
    return this.#core.getLength()
  }
}
```

And you can throw an error in the `length()` setter just to be explicit.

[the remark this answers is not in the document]
PiotrJunior: > Any sources of this?

I tested that

lib/types/inet-address.js
PiotrJunior: I didn't saw that function when I was rewriting DataStax code into JS class. I will remove that as it is never used and there isn't any mention about this in docs.

test/unit/inet-address-tests.js
```diff
@@ -168,28 +168,28 @@ describe("InetAddress", function () {
...
-            }, TypeError);
+            }, Error);
             assert.throws(function () {
                 InetAddress.fromString("::Z:11:2233:4455:aa:bb");
-            }, TypeError);
+            }, Error);
             assert.throws(function () {
                 InetAddress.fromString(" ::a:11");
-            }, TypeError);
+            }, Error);
```
Stapox35, line 192: Maybe put this value to array and use `forEach`?

lib/types/inet-address.js
```diff
@@ -10,238 +11,260 @@ const utils = require("../utils");
...
+
+    set length(value) {
+        throw new Error("InetAddress length is read-only");
+    }
 
     /**
      * Returns the Ip version (4 or 6)
+     * @readonly
      * @type Number
```
Stapox35, line 37: Maybe `@returns {Number}`? 
Docs: https://jsdoc.app/tags-returns

lib/types/inet-address.js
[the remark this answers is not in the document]
wprzytula: > And you can throw an error in the length() setter just to be explicit.

I'm in favour of explicit errors. cc @PiotrJunior

lib/types/inet-address.js
PiotrJunior: According to https://stackoverflow.com/a/22276207 @type {number} should be used

test/unit/inet-address-tests.js
PiotrJunior: Do u mean something like this:
```rust
const testCases = [
                { buffer: utils.allocBufferUnsafe(10), shouldThrow: true },
                { buffer: null, shouldThrow: true },
                { buffer: undefined, shouldThrow: true },
                { buffer: utils.allocBufferUnsafe(16), shouldThrow: false },
                { buffer: utils.allocBufferUnsafe(4), shouldThrow: false }
            ];

            testCases.forEach(({ buffer, shouldThrow }) => {
                if (shouldThrow) {
                    assert.throws(() => new InetAddress(buffer), Error);
                } else {
                    assert.doesNotThrow(() => new InetAddress(buffer), Error);
                }
            });
```

src/types/inet.rs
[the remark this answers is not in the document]
Chlor87: Sorry, I've been pretty swamped with work.
Gotta tell you, that's a tricky problem, not one I've ever dealt with before, so unfortunately I don't have a clear answer. Generally speaking calling a sync function in an async context shouldn't make the function itself async, but considering the fact that it's NAPI - can't say.
What I would advise is to stick to the sync variant AND test what will happen in an async JS context.
If all else fails, is it feasible to opt out of using buffers in this scenario?

test/unit/inet-address-tests.js
Stapox35: Yes :>

lib/types/inet-address.js
```diff
@@ -10,238 +11,260 @@ const utils = require("../utils");
...
-        address += ((this.buffer[h] << 8) | this.buffer[h + 1]).toString(16);
+        return address;
     }
-    if (address.charAt(address.length - 1) === ":") {
-        address += ":";
+
+    /**
+     * Returns the string representation.
+     * Method used by the native JSON.stringify() to serialize this instance.
+     */
```
adespawn, line 245: missing ``@returns``?

adespawn (changes requested):
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/pull/40/commits/2f6934e7fd02287384c65ba9a21facfe320d7e0c
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/pull/40/commits/fa7412b39749b96050ac9583c7f439ddea2dd23d
Please add changes as in those commits, related to inet field

lib/types/inet-address.js
```diff
@@ -10,238 +11,260 @@ const utils = require("../utils");
...
+        if (this.buffer.length === 4) {
+            return (
+                this.buffer[0] +
+                "." +
+                this.buffer[1] +
+                "." +
+                this.buffer[2] +
+                "." +
+                this.buffer[3]
+            );
```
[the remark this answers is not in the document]
Chlor87: Or `this.buffer.join('.')`

lib/types/inet-address.js
```diff
@@ -10,238 +11,260 @@ const utils = require("../utils");
...
+     * @returns {Boolean}
+     */
+    equals(other) {
+        if (!(other instanceof InetAddress)) {
+            return false;
+        }
+        return (
+            this.buffer.length === other.buffer.length &&
+            this.buffer.toString("hex") === other.buffer.toString("hex")
         );
```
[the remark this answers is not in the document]
Chlor87: In JS `return` only considers what's in the same line, so without the parens that would yield `undefined`.

lib/types/inet-address.js
```diff
@@ -10,238 +11,260 @@ const utils = require("../utils");
...
- * Compares 2 addresses and returns true if the underlying bytes are the same
- * @param {InetAddress} other
- * @returns {Boolean}
- */
-InetAddress.prototype.equals = function (other) {
-    if (!(other instanceof InetAddress)) {
-        return false;
+    /**
+     * Returns the underlying buffer
+     * @returns {Buffer}
```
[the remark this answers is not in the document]
PiotrJunior: `@return` should be used for methods and `@type` should be used for variables and properties. From the user point of view using setters and getters looks like using properties of class, so that case `@type` should be used but for this case `@return` should be used

lib/types/inet-address.js
```diff
@@ -10,238 +11,260 @@ const utils = require("../utils");
...
  */
-function InetAddress(buffer) {
-    if (
-        !(buffer instanceof Buffer) ||
-        (buffer.length !== 4 && buffer.length !== 16)
-    ) {
-        throw new TypeError("The ip address must contain 4 or 16 bytes");
-    }
+class InetAddress {
+    #internal;
```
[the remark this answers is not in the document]
PiotrJunior: `SyntaxError: Private field '#internal' must be declared in an enclosing class`

lib/types/inet-address.js
```diff
@@ -10,238 +11,269 @@ const utils = require("../utils");
...
+    /**
+     * Returns the string representation of the IP address.
+     * <p>For v4 IP addresses, a string in the form of d.d.d.d is returned.</p>
+     * <p>
+     *   For v6 IP addresses, a string in the form of x:x:x:x:x:x:x:x is returned, where the 'x's are the hexadecimal
+     *   values of the eight 16-bit pieces of the address, according to rfc5952.
+     *   In cases where there is more than one field of only zeros, it can be shortened. For example, 2001:0db8:0:0:0:1:0:1
+     *   will be expressed as 2001:0db8::1:0:1.
+     * </p>
+     * @param {String} [encoding]
```
adespawn, line 177: maybe remove html tags from ducumentation

lib/types/inet-address.js
```diff
@@ -10,238 +11,269 @@ const utils = require("../utils");
...
+            if (buffer[i] !== 0) {
+                return false;
+            }
         }
+        return !(buffer[10] !== 255 || buffer[11] !== 255);
+    }
+
+    /**
+     * @param {rust.InetAddressWrapper} InetAddress
+     * @returns {InetAddress}
```
adespawn, line 272: add ``@package`` tag

src/result.rs
```diff
@@ -205,6 +206,14 @@ impl CqlValueWrapper {
         }
     }
 
+    #[napi]
+    pub fn get_inet(&self) -> napi::Result<InetAddressWrapper> {
+        match self.inner.as_inet() {
+            Some(r) => Ok(InetAddressWrapper::from_ip_addr(r)),
+            None => Err(Error::new(Status::GenericFailure, "Error")),
```
adespawn, line 213: use ``generic_error``

adespawn (changes requested):

test/unit/inet-address-tests.js
```diff
@@ -168,28 +168,44 @@ describe("InetAddress", function () {
...
+            );
+            assert.throws(function () {
+                val.buffer = utils.allocBufferFromArray([127, 0, 0, 1]);
+            }, Error);
+            assert.throws(function () {
+                val.version = 6;
+            }, Error);
+            assert.throws(function () {
+                val.length = 4;
+            }, Error);
```
Stapox35, line 208: Please check is the correct errors -> `name` and `message`.

lib/types/inet-address.js
```diff
@@ -10,238 +11,269 @@ const utils = require("../utils");
...
+    #internal;
 
-    /**
-     * Immutable buffer that represents the IP address
-     * @type Array
-     */
-    this.buffer = buffer;
+    constructor(buffer) {
+        this.#internal = rust.InetAddressWrapper.new(buffer);
+    }
```
Stapox35, line 19: Where JSDoc for constructor?

lib/types/inet-address.js
```diff
@@ -10,238 +11,269 @@ const utils = require("../utils");
...
-    const embeddedIp4 = ipv4Pattern.test(parts[parts.length - 1]);
-    if (embeddedIp4) {
-        // Its IPv6 address with an embedded IPv4 address:
-        // subtract 1 from the potential empty filling as ip4 contains 4 bytes instead of 2 of a ipv6 section
-        filling -= 1;
+
+    /**
+     * Immutable buffer that represents the IP address
+     * @readonly
+     * @type Array
```
Stapox35, line 50: Why do field types not have brackets?

Stapox35 (changes requested):

lib/types/inet-address.js
```diff
@@ -10,238 +11,269 @@ const utils = require("../utils");
...
+     * @param {string} value - The string representation of the IP address.
+     * @returns {InetAddress} - An instance of InetAddress representing the given IP address.
+     * @throws {TypeError} - If the input string is not a valid IPv4 or IPv6 address.
+     */
+    static fromString(value) {
+        if (!value) {
+            return new InetAddress(utils.allocBufferFromArray([0, 0, 0, 0]));
+        }
+        const ipv4Pattern = /^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}$/;
+        const ipv6Pattern = /^[\da-f:.]+$/i;
```
ZuzaOsa, line 77: Provide more description of this regexes. Maybe you can use a more complcated one but skip some of the checks?

test/unit/inet-address-tests.js
PiotrJunior: I don't think it is necessary as in first test written by DataStax  error names are not checked. Also in case of changing error message, tests also would need to be changed, slowing down development time.

lib/types/inet-address.js
PiotrJunior: I'll add reference to the source of this regex

adespawn (changes requested):

lib/types/inet-address.js
```diff
@@ -10,238 +11,276 @@ const utils = require("../utils");
...
+    #internal;
 
     /**
-     * Immutable buffer that represents the IP address
-     * @type Array
+     * Creates an instance of InetAddress.
+     *
+     * @constructor
+     * @param {Buffer} buffer - The buffer containing the IPv4 or IPv6 address.
+     * @throws {Error} - If the buffer is not a valid IPv4 or IPv6 address.
```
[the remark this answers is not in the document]
PiotrJunior: yes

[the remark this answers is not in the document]
adespawn: Why?

lib/types/inet-address.js
```diff
@@ -10,238 +11,276 @@ const utils = require("../utils");
...
-            if (length > longest.length) {
-                longest.length = length;
-                longest.start = start;
-                start = -1;
-            }
+
+    /**
+     * Returns the string representation of the IP address.
+     * For v4 IP addresses, a string in the form of d.d.d.d is returned.
+     * For v6 IP addresses, a string in the form of x:x:x:x:x:x:x:x is returned, where the 'x's are the hexadecimal
```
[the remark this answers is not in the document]
adespawn: Still there is one paragraph instead of two for Ipv4 and Ipv6

src/types/inet.rs
```diff
@@ -0,0 +1,66 @@
...
+            })
+        } else if let Ok(arr) = <[u8; 16]>::try_from(buffer) {
+            Ok(Self {
+                inet: IpAddr::from(arr),
+            })
+        } else {
+            Err(Error::new(
+                Status::GenericFailure,
+                "Invalid IP address length",
+            ))
```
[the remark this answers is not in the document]
adespawn: ...

lib/types/inet-address.js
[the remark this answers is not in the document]
PiotrJunior: In https://jsdoc.app/howto-es2015-classes `-` are used. Also In VSCode hints are more readable with `-`. But if you want I can remove them.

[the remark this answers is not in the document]
adespawn: Currently I believe no other place uses ``-``, so let's be consistent.

lib/types/inet-address.js
```diff
@@ -1,247 +1,282 @@
...
  */
-function InetAddress(buffer) {
-    if (
-        !(buffer instanceof Buffer) ||
-        (buffer.length !== 4 && buffer.length !== 16)
-    ) {
-        throw new TypeError("The ip address must contain 4 or 16 bytes");
-    }
+class InetAddress {
+    #internal;
```
Stapox35, line 11:
```suggestion
    /**
    * @type {rust.InetAddressWrapper}
    * @private
    */
    #internal;
```

PiotrJunior: Why do you want to add`@private` tag, when it is already marked private by `#`

adespawn: Then at least add ``@type`` tag

test/unit/inet-address-tests.js
```diff
@@ -9,19 +9,19 @@ describe("InetAddress", function () {
...
-            }, TypeError);
+            }, Error);
             assert.doesNotThrow(function () {
                 return new InetAddress(utils.allocBufferUnsafe(16));
-            }, TypeError);
+            }, Error);
             assert.doesNotThrow(function () {
                 return new InetAddress(utils.allocBufferUnsafe(4));
-            }, TypeError);
+            }, Error);
```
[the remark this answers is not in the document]
adespawn: Mention it here in the comment also

ZuzaOsa (approved):

lib/types/inet-address.js
ZuzaOsa, line 1: Remember to run prettier before merging.

lib/types/inet-address.js
```diff
@@ -1,247 +1,285 @@
...
-function InetAddress(buffer) {
-    if (
-        !(buffer instanceof Buffer) ||
-        (buffer.length !== 4 && buffer.length !== 16)
-    ) {
-        throw new TypeError("The ip address must contain 4 or 16 bytes");
-    }
+class InetAddress {
+    /**
+    * @type {rust.InetAddressWrapper}
```
Stapox35, line 12: In other files we add here `@private`

lib/types/inet-address.js
```diff
@@ -1,247 +1,285 @@
...
+            if (buffer[i] !== 0) {
+                return false;
+            }
         }
+        return !(buffer[10] !== 255 || buffer[11] !== 255);
+    }
+
+    /**
+     * @package
+     * @param {rust.InetAddressWrapper} InetAddress
```
Stapox35, line 277: Here should be the name of the argument `arg`, not `InetAddress`.

test/unit/inet-address-tests.js
```diff
@@ -166,28 +166,62 @@ describe("InetAddress", function () {
...
                     InetAddress.fromString(address);
-                }, TypeError);
+                }, Error);
             });
         });
         it("should throw TypeError when the string is not a valid address", function () {
             assert.throws(function () {
                 InetAddress.fromString("127.0.0.1.10");
-            }, TypeError);
+            }, Error);
```
Stapox35, line 175: Same TypeError in description.

Stapox35 (changes requested):

test/unit/inet-address-tests.js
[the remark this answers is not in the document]
PiotrJunior: Napi can't throw TypeError, so i need to change checked Error type. #88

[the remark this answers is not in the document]
adespawn: Yes. But that's changing of the driver behavior. So mention it also here in a comment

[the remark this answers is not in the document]
PiotrJunior: Sorry, I thought you mean in GitHub comment

PR #40, opened 2024-10-24
Execute parameters

adespawn:
Allow to execute queries to database while passing parameters. Adds support for adding and receiving empty values (partially solves #79).

Updates rust driver version to 0.15.0.

Add simple implementation of batch queries through rust driver. Batch query object is kept in a new wrapper, exposed to JS driver.


src/utils.rs
```diff
@@ -0,0 +1,7 @@
+use std::fmt::Display;
+
+use napi::{Error, Status};
+
+pub fn js_generic_err(msg: impl Display) -> Error<Status> {
+  Error::new(Status::GenericFailure, msg)
+}
```
wprzytula, line 7: I'm happy to see an instant use of new knowledge from today's lab - `impl Display` :wink:

src/result.rs
```diff
@@ -51,6 +54,7 @@ pub enum CqlType {
     Tuple,
     Uuid,
     Varint,
+    Custom,
```
PiotrJunior, line 57: What is Custom type? It is not specified in rust driver.

lib/client.js
```diff
@@ -792,62 +793,46 @@ class Client extends events.EventEmitter {
...
-            // Only Cassandra 2.1 and above supports named parameters
-            throw new errors.ArgumentError(
-                "Named parameters for simple statements are not supported, use prepare flag",
-            );
+            throw new Error(`TODO: Implement any support for named parameters`);
+            // // Only Cassandra 2.1 and above supports named parameters
+            // throw new errors.ArgumentError(
+            //   "Named parameters for simple statements are not supported, use prepare flag",
+            // );
         }
```
PiotrJunior, line 815: Maybe add issue for every temporally disabled feature to not forget about fixing those.

lib/client.js
```diff
@@ -792,62 +793,46 @@ class Client extends events.EventEmitter {
...
 
-        let request;
-
         if (!this.connected) {
-            // Micro optimization to avoid an async execution for a simple check
+            // Check this logic and decide if it's needed. Probably toDo while implementing (better) connection
+            // // Micro optimization to avoid an async execution for a simple check
```
PiotrJunior, line 819: double `// //`

lib/client.js
```diff
@@ -792,62 +793,46 @@ class Client extends events.EventEmitter {
...
-                (types.protocolVersion.supportsKeyspaceInRequest(version) &&
-                    execOptions.getKeyspace()) ||
-                this.keyspace;
-    
-        const { queryId, meta } = await PrepareHandler.getPrepared(this, lbp, query, queryKeyspace);
-        request = await this.#createExecuteRequest(query, queryId, execOptions, params, meta);
-        */
-        }
-
-        return await RequestHandler.send(request, execOptions, this);
```
PiotrJunior, line 850: Also add this as an issue

lib/client.js
adespawn: That's intentional

lib/client.js
```diff
@@ -792,62 +793,46 @@ class Client extends events.EventEmitter {
...
+        return new Promise(async (resolve, reject) => {
+            try {
+                let statement = await this.rust_client.prepareStatement(query);
+                let parsed_params = parseParams(statement, params);
```
ZuzaOsa, line 826: Use camelcase in JS.

ZuzaOsa (approved):
One variable name comment, otherwise lgtm.

lib/client.js
```diff
@@ -792,62 +793,46 @@ class Client extends events.EventEmitter {
     }
 
     /**
-     * Connects and handles the execution of prepared and simple statements.
+     * Wrapper for executing queries by rust driver
```
Stapox35, line 796: Please check the rest of the private methods. They do not have arguments listed in JSDoc.

Stapox35 (changes requested):
At the end of the `client.js` file is a mysterious description of a function that is not there....

wprzytula:
> Changed the behavior of the driver, so it executes all queries as prepared.

Can we mimic the Rust driver's behaviour that all queries with nonempty parameter list are prepared, but those without parameters can be sent as unprepared? This should not cause any problems around serialization, because there is nothing to be serialized in such case.

src/request.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    #[napi]
+    pub fn from_text(val: String) -> QueryParameterWrapper {
+        QueryParameterWrapper {
+            parameter: CqlValue::Ascii(val),
+        }
+    }
+
+    #[napi]
+    pub fn from_set(val: Vec<&QueryParameterWrapper>) -> QueryParameterWrapper {
```
wprzytula, line 102: This seems a very weird type for a set to me. Why have a `Vec` of references at all?
Perhaps this function should accept `impl Iterator<Item = &QueryParameterWrapper>`?

wprzytula: Also, we don't seem to have list nor map supported here, only set. Why so?

src/request.rs
```diff
@@ -1,17 +1,26 @@
...
 #[napi]
 pub struct QueryParameterWrapper {
-    parameter: CqlValue,
+    pub(crate) parameter: CqlValue,
+}
+
+#[napi]
+pub struct PreparedStatementWrapper {
+    pub(crate) query: PreparedStatement,
 }
```
wprzytula: When introducing public structs like these, let's try to write at least minimal docstrings, so that reading the code is easier. Just by hovering the type's name we will get a short info what it is for.

src/session.rs
```diff
@@ -9,7 +13,7 @@ pub struct SessionOptions {
 
 #[napi]
 pub struct SessionWrapper {
-    internal_session: Session,
+    internal: Session,
 }
```
wprzytula, line 17: Such a rename should be done in a separate commit, not to clutter the commit that contains meaningful new logic.

src/request.rs
```diff
@@ -1,17 +1,26 @@
...
 
 #[napi]
 pub struct QueryParameterWrapper {
-    parameter: CqlValue,
+    pub(crate) parameter: CqlValue,
+}
+
+#[napi]
+pub struct PreparedStatementWrapper {
+    pub(crate) query: PreparedStatement,
```
wprzytula, line 23: Let's have this field named `prepared`, because `query` is unfortunately a usual name for the unprepared statement in the Rust driver.

src/session.rs
```diff
@@ -31,18 +35,45 @@ impl SessionWrapper {
             .build()
             .await
             .map_err(err_to_napi)?;
-        Ok(SessionWrapper {
-            internal_session: sb,
-        })
+        Ok(SessionWrapper { internal: sb })
```
wprzytula, line 38: The variable name should be `s`, not `sb`, because it's an already built Session, not SessionBuilder.

src/session.rs
```diff
@@ -31,18 +35,45 @@ impl SessionWrapper {
...
     }
 
     #[napi]
     pub async fn query_unpaged_no_values(&self, query: String) -> napi::Result<QueryResultWrapper> {
         let query = self
-            .internal_session
+            .internal
```
wprzytula, line 44: Naming nit: `query` -> `query_result`

src/session.rs
```diff
@@ -31,18 +35,45 @@ impl SessionWrapper {
...
+            query: self
+                .internal
+                .prepare(statement)
+                .await
+                .map_err(err_to_napi)?,
+        })
+    }
+
+    #[napi]
+    pub async fn query_simple(
```
wprzytula, line 66: Please explain the name.

wprzytula, line 66: Please provide docstrings to these functions, as they are crucial for the new driver.

src/request.rs
adespawn: 1. This functions is exposed to napi, so I don't think iterators will work here.

2. Those types are implemented here: https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/pull/95

src/request.rs
```diff
@@ -63,6 +63,13 @@ impl QueryParameterWrapper {
         }
     }
 
+    #[napi]
+    pub fn from_duration(val: &DurationWrapper) -> QueryParameterWrapper {
+        QueryParameterWrapper {
+            parameter: CqlValue::Duration(val.get_cql_duration()),
+        }
+    }
+
```
wprzytula, line 72: This should belong to the previous commit.

lib/types/cql-utils.js
```diff
@@ -0,0 +1,79 @@
...
+            return rust.QueryParameterWrapper.fromBlob(value);
+        case rust.CqlType.Boolean:
+            return rust.QueryParameterWrapper.fromBoolean(value);
+        case rust.CqlType.Counter:
+            return rust.QueryParameterWrapper.fromCounter(value);
+        case rust.CqlType.Double:
+            return rust.QueryParameterWrapper.fromDouble(value);
+        case rust.CqlType.Duration:
+            if (!(value instanceof Duration)) throw new TypeError("");
+            return rust.QueryParameterWrapper.fromDuration(value.getInternal());
```
wprzytula, line 33: Why is this check present only for Duration?

wprzytula: Well, and for Uuid and TimeUuid.

lib/types/cql-utils.js
```diff
@@ -0,0 +1,79 @@
...
+            if (!(value instanceof Uuid)) throw new TypeError("");
+            return rust.QueryParameterWrapper.fromUuid(value.getInternal());
+        case rust.CqlType.Timeuuid:
+            if (!(value instanceof TimeUuid)) throw new TypeError("");
+            return rust.QueryParameterWrapper.fromTimeUuid(value.getInternal());
+        default:
+            // Or not yet implemented type
+            throw new ReferenceError(
+                `[INTERNAL DRIVER ERROR] Unknown type: ${type}`,
+            );
```
wprzytula, line 56: Should it really be a `ReferenceError`?

lib/client.js
```diff
@@ -792,62 +793,46 @@ class Client extends events.EventEmitter {
...
+                let rusty = await this.rustClient.querySimple(
+                    statement,
+                    parsedParams,
+                );
+                resolve(new ResultSet(rusty));
+            } catch (e) {
+                reject(e);
+            }
+        });
     }
```
wprzytula, line 836: I don't see the previous code use `reject` at all, and you do. Please explain it in a comment.

test/integration/supported/client-execute-tests.js
```diff
@@ -217,8 +221,9 @@ describe("Client @SERVER_API", function () {
             ];
             // no hint
             insertSelectTest(client, table, columns, values, null, done);
-        });
-        vit(
+        }); */
+        // No support for parameter hints
+        /* vit(
```
wprzytula, line 226: I strongly reject such practice of commenting out test cases without having them documented clearly and marked with a tag for easy lookup.
Tests are here to reliably measure our progress; we mustn't decrease their accuracy.

lib/types/cql-utils.js
```diff
@@ -14,6 +14,9 @@ const Duration = require("./duration");
  * @returns {rust.QueryParameterWrapper}
  */
 function getWrapped(type, value) {
+    if (value === null || value === types.unset) {
+        return rust.QueryParameterWrapper.fromEmpty();
+    }
```
wprzytula, line 19: `CqlValue::Empty` is something totally different than the DB concept of `null`. Please don't mix them up.

src/result.rs
```diff
@@ -129,11 +129,16 @@ impl QueryResultWrapper {
...
+    pub fn get_columns(&self) -> napi::Result<Vec<Option<CqlValueWrapper>>> {
+        let s: Vec<Option<CqlValueWrapper>> = self
             .internal
             .iter()
-            .filter_map(|f| f.as_ref().map(|r| CqlValueWrapper { inner: r.clone() }))
+            .map(|f| {
+                if f.is_none() {
+                    return None;
+                }
+                f.as_ref().map(|r| CqlValueWrapper { inner: r.clone() })
```
wprzytula, line 140: FP: let's use `Option::map` instead of imperative style.

wprzytula (changes requested):

lib/types/cql-utils.js
adespawn: For those types we call internal function on the passed object. For other types, the program will throw an error with meaningful description, while it's not the case for those types.

(I've updated the error messages for those TypeErrors)

test/integration/supported/client-execute-tests.js
adespawn: How would you suggest to mark those tests that are currently failing due to lack of nesesery features in the driver then?

wprzytula: At least with a `TODO:`. But it's best to come up with a dedicated keyword, e.g. `TODO: TESTING:` or something like this.

lib/client.js
wprzytula: It's simply me not knowing JS. Nvm.

lib/types/cql-utils.js
wprzytula: Could you put the above explanation in a comment?

src/request.rs
```diff
@@ -123,3 +134,15 @@ impl QueryParameterWrapper {
         }
     }
 }
+
+#[napi]
+impl PreparedStatementWrapper {
+    #[napi]
+    pub fn get_expected_types(&self) -> Vec<CqlType> {
```
wprzytula, line 141: Missing docstring.

src/session.rs
```diff
@@ -52,6 +53,46 @@ impl SessionWrapper {
...
+    /// Creating Prepared statement may help to determine required types
+    ///
+    /// Currently `execute_unpaged` from rust driver is used, so no pagin is done
+    /// and there is no support for any query parameters
+    #[napi]
+    pub async fn execute_prepared(
+        &self,
+        query: &PreparedStatementWrapper,
+        params: Vec<&QueryParameterWrapper>,
+    ) -> napi::Result<QueryResultWrapper> {
```
wprzytula: > there is no support for any query parameters

Really? Isn't that what this PR is for?

lib/types/cql-utils.js
```diff
@@ -14,6 +14,9 @@ const Duration = require("./duration");
  * @returns {rust.QueryParameterWrapper}
  */
 function getWrapped(type, value) {
+    if (value === null || value === types.unset) {
+        return null;
+    }
```
wprzytula, line 19: This commit still mentions "empty" values, whereas it's about null values.

test/integration/supported/client-execute-tests.js
```diff
@@ -600,8 +600,7 @@ describe("Client @SERVER_API", function () {
                 done,
             );
         }); */
-        // issues #79
-        /* vit("2.2", "should accept unset as a valid value", function (done) {
+        vit("2.2", "should accept unset as a valid value", function (done) {
```
wprzytula, line 603: This test mentions `unset`, not `null`s. Isn't it contrary to the commit's description?

src/session.rs
adespawn: My mistake. I meant query option

test/integration/supported/client-execute-tests.js
adespawn: The commit should enable both things

src/session.rs
```diff
@@ -52,6 +61,73 @@ impl SessionWrapper {
...
+    }
+
+    /// Query a database with a given prepared query and provided parameters.
+    ///
+    /// Returns a wrapper of the value provided by the rust driver
+    ///
+    /// All parameters need to be wrapped into QueryParameterWrapper keeping CqlValue of correct type
+    /// Creating Prepared statement may help to determine required types
+    ///
+    /// Currently `execute_unpaged` from rust driver is used, so no pagin is done
```
wprzytula, line 88: :recycle: typo: pagin -> paging

src/session.rs
```diff
@@ -52,6 +61,73 @@ impl SessionWrapper {
...
+        Ok(PreparedStatementWrapper {
+            prepared: self
+                .internal
+                .prepare(statement)
+                .await
+                .map_err(err_to_napi)?,
+        })
+    }
+
+    /// Query a database with a given prepared query and provided parameters.
```
wprzytula, line 81: :recycle: naming: `prepared query` -> `prepared statement`

src/request.rs
```diff
@@ -0,0 +1,160 @@
...
+            parameter: CqlValue::Duration(val.get_cql_duration()),
+        }
+    }
+
+    #[napi]
+    pub fn from_float(val: f64) -> QueryParameterWrapper {
+        QueryParameterWrapper {
+            parameter: CqlValue::Float(val as f32),
+        }
+    }
```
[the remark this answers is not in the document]
adespawn: I think this is the best thing we can do here. We can't assume that value can be converted without loses even for small values like ``0.1``. By casting we just loses precision and we can't reasonably require JS part to provide value in a way that will be lossless.
``f32::try_from<f64>`` is not implemented.

src/result.rs
```diff
@@ -70,71 +78,91 @@ pub enum CqlType {
...
         };
 
-        Some(
-            rows.iter()
-                .map(|f| RowWrapper {
-                    internal: f.columns.clone(),
-                })
-                .collect(),
-        )
+        let rows = r2.rows::<Row>().map_err(err_to_napi)?.flatten();
```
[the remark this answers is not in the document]
adespawn: 2. I would prefer to do it later, as this would require investigation how old driver behaves when it encounters that kind of problems.

src/result.rs
```diff
@@ -70,71 +78,91 @@ pub enum CqlType {
...
+        Ok(Some(
+            rows.map(|f| RowWrapper {
+                internal: f.columns,
+            })
+            .collect(),
+        ))
```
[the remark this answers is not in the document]
adespawn: I don't understand the difference. Additionally ``collect`` after this change requires type annotation

lib/client.js
```diff
@@ -608,45 +609,32 @@ class Client extends events.EventEmitter {
...
+                );
+                parametersRows.push([]);
+            } else {
+                let query = queries[i].query;
+                let params = queries[i].params;
+                let prepared = await this.rustClient.prepareStatement(query);
+                let parameters = parseParams(prepared, params);
+                preparedQueries.push(prepared);
+                parametersRows.push(parameters);
             }
```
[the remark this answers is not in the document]
adespawn: This logic is taken from Datastax code. I'm not aware of any sensible way of doing what you asked for. Either way if some invalid value is provided, Error will be thrown.

src/result.rs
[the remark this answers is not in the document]
wprzytula: > I would prefer to do it later, as this would require investigation how old driver behaves when it encounters that kind of problems.

I insist on not allowing any errors to be silently ignored in our merged code.

lib/client.js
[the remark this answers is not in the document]
wprzytula: OK, I'm not going to wrestle with JS.

src/result.rs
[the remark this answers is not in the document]
adespawn: In this case I updated the code to panic in case on Deserialization error. The correct error handling I would still prefer to do at a later time.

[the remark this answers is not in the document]
wprzytula: OK, I can accept this as a temporary measure.

wprzytula:
:recycle: I suggest that bumping up the Rust driver's version be done in a separate PR, to reduce size of this PR (and future PRs, too).

lib/types/cql-utils.js
```diff
@@ -0,0 +1,93 @@
...
+ * @param {CqlType} type
+ * @param {*} value
+ * @returns {rust.QueryParameterWrapper}
+ */
+function getWrapped(type, value) {
+    if (value === null || value === types.unset) {
+        return null;
+    }
+    // To increase clarity of the error messages, if we call some function on value,
+    // type is checked explicitly. In other cases default Error will be thrown.
```
wprzytula, line 21: :question: :wrench: It's not clear to me what scenario this describes.

wprzytula: :question:  What function? What value?

src/result.rs
```diff
@@ -70,71 +75,93 @@ pub enum CqlType {
 #[napi]
 impl QueryResultWrapper {
     pub fn from_query(internal: QueryResult) -> QueryResultWrapper {
-        QueryResultWrapper { internal }
+        let value = match internal.clone().into_rows_result() {
+            Ok(v) => QueryResultVariant::RowsResult(v),
+            Err(_) => QueryResultVariant::EmptyResult(internal),
+        };
```
wprzytula, line 81: :wrench: IIRC, `into_rows_result` may fail in two cases:
1. if the underlying Result response is not of Rows kind (correctly handled here)
2. if deserialization of Result metadata fails. In such case, the error should be propagated, not converted into `QueryResultVariant::EmptyResult`, because the current handling is basically silent ignoring of errors.

src/result.rs
```diff
@@ -70,71 +75,93 @@ pub enum CqlType {
 #[napi]
 impl QueryResultWrapper {
     pub fn from_query(internal: QueryResult) -> QueryResultWrapper {
-        QueryResultWrapper { internal }
+        let value = match internal.clone().into_rows_result() {
```
wprzytula, line 78: :wrench: Needless `clone()` - `internal` is an owned `QueryResult` that we can consume.

wprzytula (changes requested):

adespawn:
> ♻️ I suggest that bumping up the Rust driver's version be done in a separate PR, to reduce size of this PR (and future PRs, too).

It was part of  other PR that got merged into this one: https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/pull/111

lib/types/cql-utils.js
adespawn: I hope the comment is more meaningful now.

src/result.rs
```diff
@@ -69,72 +77,97 @@ pub enum CqlType {
...
-        QueryResultWrapper { internal }
+    pub fn from_query(internal: QueryResult) -> napi::Result<QueryResultWrapper> {
+        let value = match internal.result_not_rows() {
+            Ok(_) => QueryResultVariant::EmptyResult(internal),
+            // We can unwrap, as error will be returned only if result_not_rows is OK
+            Err(_) => {
+                QueryResultVariant::RowsResult(internal.into_rows_result().map_err(err_to_napi)?)
+            }
+        };
+        Ok(QueryResultWrapper { internal: value })
```
wprzytula, line 88: :wrench: This is logically correct, but not rusty (no clear exhaustive pattern matching). What should be done here is pattern matching on `internal.into_rows_result()`.

wprzytula (changes requested):

src/result.rs
adespawn: Then we can't avoid cloning, as matching on it consumes it, and we can't create empty result from it

wprzytula: [Have you looked at the error variants that can arise there?](https://docs.rs/scylla/latest/scylla/transport/query_result/enum.IntoRowsResultError.html)
`IntoRowsResultError::ResultNotRows(QueryResult)` allows for getting the `QueryResult` back.

wprzytula (approved):

PR #53, opened 2024-10-30
Split unit into three categories 

adespawn:
One category for test that are currently passing because of old datastax code
Second category for tests that are being broken by current changes (this category is not tested on CI)
Third category for unit tests that should work with the new driver code

(Note, CI examples fails due changes in ``simple-query``)

PR #54, opened 2024-10-31
Refactor examples & CI

adespawn:
Split examples and unit tests into two CI tasks
Refactor examples: Update runner, and allow for running not only Datastax example
To ensure some testing on different theoretically supported node version updated the examples CI so they run on all major supported versions up to node20.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,5 @@
+# Examples
+
+Currently there is no examples showing usage of the driver. For examples from the DataStax driver, see ``./DataStax`` directory.
+
+The ``basic.js`` example is a proof of concept connection to the database. It call the Rust layer directly, which will not be possible in the final package, as the rust layer will not be exposed in the API.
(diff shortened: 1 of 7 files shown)


adespawn:
Unit tests are failing as they lack #53.
Examples are failing as the branch already contains some changes to the driver code

examples/README.md
```diff
@@ -0,0 +1,5 @@
+# Examples
+
+Currently there is no examples showing usage of the driver. For examples from the DataStax driver, see ``./DataStax`` directory.
```
wprzytula, line 3: Is that true? I think the examples from DataStax driver should be valid for our driver too, so there actually **are** examples showing usage of the driver.

wprzytula (approved):

examples/README.md
adespawn: It will change very soon and what I wanted to say that there is no new examples

PR #55, opened 2024-10-31
Add type conversion helper function

adespawn:
Add utility functions for converting between nodeJS bigint, type for arbitrary length integers (used also in NAPI when passing certain types between rust and node layers) and Long type used internally and in exposed API by datastax driver.

### DIFF
--- a/test/unit/new-utils.js.js
+++ b/test/unit/new-utils.js.js
@@ -1,0 +1,41 @@
+"use strict";
+const { assert } = require("chai");
+const { bigintToLong, longToBigint } = require("../../lib/new-utils");
+const Long = require("long");
+
+const values = [
+    "0",
+    "1",
+    "-1",
+    "12",
+    "-12",
+    "20040400020",
+    "120000000000",
+    "9223372036854775807",
+    "9223372036854775806",
+    "-9223372036854775808",
+    "-9223372036854775805",
+    "-20040400020",
+];
+
+describe("Numeric types conversions", function () {
+    // We assume BigInt and Long are fully correct
+    describe("BigInt to Long", function () {
+        it("should convert correctly", function () {
+            values.forEach((value) => {
+                let bigInt = BigInt(value);
+                let long = bigintToLong(bigInt);
+                assert.strictEqual(value, long.toString());
+            });
+        });
+    });
+    describe("Long to BigInt", function () {
+        it("should convert correctly", function () {
+            values.forEach((value) => {
+                let long = Long.fromString(value);
+                let bigInt = longToBigint(long);
+                assert.strictEqual(value, bigInt.toString());
+            });
+        });
+    });
+});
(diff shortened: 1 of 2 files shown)


test/unit/utils.js
```diff
@@ -0,0 +1,41 @@
...
+        it("should covert correctly", function () {
+            values.forEach((value) => {
+                let bigInt = BigInt(value);
+                let long = bigintToLong(bigInt);
+                assert.deepEqual(value, long.toString());
+            });
+        });
+    });
+    describe("Long to BigInt", function () {
+        it("should covert correctly", function () {
```
wprzytula, line 33: covert -> convert

test/unit/utils.js
```diff
@@ -0,0 +1,41 @@
...
+            values.forEach((value) => {
+                let long = Long.fromString(value);
+                let bigInt = longToBigint(long);
+                assert.deepEqual(value, bigInt.toString());
```
wprzytula, line 37: Is `deepEqual` a content-wise comparison between strings (as opposed to pointer-wise) ?

adespawn: I missread the docuemntation, and what I wanted was to compare the objects with ===, instead of == in normal equal.
When it comes to deep equal the [docuentation](https://www.chaijs.com/api/assert/) says: ``Asserts that actual is deeply equal to expected.`` and to be fair I don't fully understand what it means...

test/unit/utils.js
```diff
@@ -0,0 +1,41 @@
...
+    describe("Long to BigInt", function () {
+        it("should convert correctly", function () {
+            values.forEach((value) => {
+                let long = Long.fromString(value);
+                let bigInt = longToBigint(long);
+                assert.strictEqual(value, bigInt.toString());
+            });
+        });
+    });
+});
```
wprzytula, line 41: Should this be in `utils.js` or `new-utils.js`?

adespawn: ``new-utils.js`` would make more sense here.

test/unit/new-utils.js.js
```diff
@@ -0,0 +1,41 @@
+"use strict";
+const { assert } = require("chai");
+const { bigintToLong, longToBigint } = require("../../lib/new-utils");
+const Long = require("long");
+
```
wprzytula, line 5: I assume double `.js` in `new-utils.js.js` is on purpose, as it's a `.js` file that tests `new-utils.js`.

wprzytula (approved):

wprzytula:
This is a very good example of a first-order PR, as we discussed in PWN. It's small and clear, so it's for reviewing small details. When we open a second-order PR (that is, `types-support` -> `main`), there will be space for Big-Picture phase of the review.

Stapox35 (approved):

ZuzaOsa (approved):

Issue #46, opened 2024-10-29
Add missing fields in result set


adespawn:
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/ec019583c846ca0dabff41d1e0c61326b89d8d21/lib/types/result-set.js#L108-L117

wprzytula:
FYI @adespawn : https://github.com/scylladb/scylla-rust-driver/issues/1030 is done, so you should be able to support `queriedHosts` field.

Issue #20, opened 2024-10-20
Add support for Integer type


adespawn:

adespawn:
Note. Test the support for this type, once #275 is merged, at least for the encoding side

Issue #19, opened 2024-10-20
Add support for Long type


adespawn:

PiotrJunior:
Shoudn't this issue be closed?

wprzytula:
> Shoudn't this issue be closed?

@adespawn

PR #56, opened 2024-10-31
First iteration of new driver 

adespawn:
***REMEMBER TO EDIT THIS MESSAGE, WHEN MERGING TO THIS BRANCH***

Things already merged into this branch:

- Cleanup of client, to use class syntax
- Add utility functions for converting between nodeJS bigint, type for arbitrary length integers (used also in NAPI when passing certain types between rust and node layers) and Long type used internally and in exposed API by datastax driver.
- Adds unit tests for CqlValueWrapper (tests types passing between rust and js part). Those tests check if objects are passed correctly between Rust and NodeJs driver and have the expected structure. 
- Refactor creating ``CqlObject`` in JS part
- Refactor rust part of ``CqlValueWrapper`` (mostly error throwing and variable names)
- Added some metadata to ResultSet constructor (info field). Some parameters still need to be added (Issue #46). 
- Refactored ResultSet to allow construction from rust results and keep old constructor for unit test compatibility
- Updates rust driver version to 0.15.0.
- Allow to execute queries to database while passing parameters. Adds support for adding and receiving empty values (partially solves #79).
- Add simple implementation of batch queries through rust driver. Batch query object is kept in a new wrapper, exposed to JS driver.

Support for following CQL Types:
 - BigInt
 - List
 - Inet
 - Map
 - Timestamp
 - Add Uuid type (and TimeUuid which extends Uuid)
 - Add Duration type
 - LocalTime

Things required to finish this branch:
- [ ] #19
- [ ] #20 (Not planned in first iteration)
- [ ] #21 (Not planned in first iteration)
- [x] #22
- [x] #23 
- [ ] #24 
- [ ] #25 
- [x] #26 
- [x] #27 
- [x] #40

Issues related to this PR:

TODO introduced in this PR:
```
```


lib/client.js
```diff
@@ -354,868 +166,675 @@ const warmupLimit = 32;
...
-            "Named parameters for simple statements are not supported, use prepare flag",
-        );
+    /**
+     * Closes all connections to all hosts.
+     *
+     * It returns a ``Promise`` when a ``callback`` is not provided.
+     *
+     * @param {Function} [callback] Optional callback to be invoked when finished closing all connections.
+     */
+    shutdown(callback) {
```
PiotrJunior, line 649: Mark as deprecated

lib/client.js
```diff
@@ -354,868 +166,675 @@ const warmupLimit = 32;
...
-        );
+    /**
+     * Gets a snapshot containing information on the connections pools held by this Client at the current time.
+     *
+     * The information provided in the returned object only represents the state at the moment this method was called and
+     * it's not maintained in sync with the driver metadata.
+     *
+     * @returns {ClientState} A [ClientState]{@linkcode module:metadata~ClientState} instance.
+     */
+    getState() {
```
PiotrJunior, line 636: Look further into this

lib/client.js
```diff
@@ -354,868 +166,675 @@ const warmupLimit = 32;
...
+
+    /**
+     * Waits for schema agreements and schedules schema metadata refresh.
+     * @param {Connection} connection
+     * @param event
+     * @returns {Promise<boolean>}
+     * @ignore
+     * @internal
+     */
+    async handleSchemaAgreementAndRefresh(connection, event) {
```
PiotrJunior, line 725: Does this work in our driver and do we need this to be public?

src/types/local_time.rs
```diff
@@ -0,0 +1,125 @@
...
+            },
+            if self.nanosecond > 0 {
+                let zeros = 9 - self.nanosecond.to_string().chars().count();
+                let mut nanos = self.nanosecond;
+                while nanos % 10 == 0 {
+                    nanos /= 10;
+                }
+                format!(".{}{}", "0".repeat(zeros), nanos)
+            } else {
+                "".to_string()
```
PiotrJunior, line 121: Fix this

src/lib.rs
```diff
@@ -20,3 +20,11 @@ pub async fn test_connection(uri: String) -> String {
```
PiotrJunior, line 20: Remove test_connection above

test/unit/inet-address-tests.js
```diff
@@ -6,22 +6,23 @@ const InetAddress = require("../../lib/types").InetAddress;
 
 describe("InetAddress", function () {
     describe("constructor", function () {
+        // TODO: DataStax throws TypeErrors, but due to napi-rs limitations, we temporarily throw Errors.
```
PiotrJunior, line 9: DO

eslint.config.mjs
```diff
@@ -35,6 +36,9 @@ export default [
       // Disabled in some of the tests because they are inserting values into the database
       "camelcase": "error",
 
+      // TODO: FIX
```
PiotrJunior, line 39: DO: FIX

PiotrJunior (changes requested):
Move Rust test files to test directory.
Fix filenames in JS tests (*-tests and .js.js).
Enable typescript tests.

PiotrJunior (changes requested):
- Move test files from rust directory to test directory.
- Make sure typescript tests works.
- Something more but I dont remember what

eslint.config.mjs
adespawn: Otwarty PR

lib/client.js
adespawn: Otwarty PR

lib/client.js
adespawn: Otwarty PR

wprzytula (approved):

Stapox35 (approved):

PiotrJunior (approved):

PR #57, opened 2024-10-31
Add support for CQL types

adespawn:
***REMEMBER TO EDIT THIS MESSAGE, WHEN MERGING TO THIS BRANCH***

Things already merged into this branch:

- Add utility functions for converting between nodeJS bigint, type for arbitrary length integers (used also in NAPI when passing certain types between rust and node layers) and Long type used internally and in exposed API by datastax driver.
- Adds unit tests for CqlValueWrapper (tests types passing between rust and js part). Those tests check if objects are passed correctly between Rust and NodeJs driver and have the expected structure. 
- Refactor creating ``CqlObject`` in JS part
- Refactor rust part of ``CqlValueWrapper`` (mostly error throwing and variable names)
- Added some metadata to ResultSet constructor (info field). Some parameters still need to be added (Issue #46). 
- Refactored ResultSet to allow construction from rust results and keep old constructor for unit test compatibility
- Updates rust driver version to 0.15.0.
- Allow to execute queries to database while passing parameters. Adds support for adding and receiving empty values (partially solves #79).
- Add simple implementation of batch queries through rust driver. Batch query object is kept in a new wrapper, exposed to JS driver.

Support for following CQL Types:
 - BigInt
 - List
 - Map
 - Timestamp
 - Add Uuid type (and TimeUuid which extends Uuid)
 - Add Duration type

Things required to finish this branch:
- [ ] #19
- [ ] #20 (Not planed in first iteration)
- [ ] #21 (Not planed in first iteration)
- [x] #22
- [x] #23 
- [ ] #24 
- [ ] #25 
- [x] #26 
- [x] #27 
- [x] #40

Issues related to this PR:

TODO introduced in this PR:
```
```


lib/client.js
```diff
@@ -713,9 +677,6 @@ class Client extends events.EventEmitter {
         if (this.options.requestTracker) {
             this.options.requestTracker.shutdown();
         }
-
-        // go through all the host and shut down their pools
-        await Promise.all(hosts.map((h) => h.shutdown(false)));
     }
```
PiotrJunior, line 681: Mark as deprecated

lib/client.js
```diff
@@ -681,11 +647,12 @@ class Client extends events.EventEmitter {
      * @param {Function} [callback] Optional callback to be invoked when finished closing all connections.
      */
     shutdown(callback) {
```
PiotrJunior, line 649: Mark as deprecated

PiotrJunior (changes requested):
Move rust test files to test directory.

PR #43, opened 2024-10-27
Add integration tests

adespawn:
Adds integration tests to CI.

Tools required to run the integration tests (not automatically installed):
 - [scylla-ccm python script](https://github.com/scylladb/scylla-ccm) used by tests to manage database instances 
 - [simulacron.jar](https://github.com/datastax/simulacron/) used by some tests. Currently tests assume that this file is in ``SIMULACRON_PATH`` directory if such env variable is provided or ``~/ `` otherwise. 

Some of the tests were removed as they test features that are not planed. Other were disabled due to lack of consistency or just because they don't work. Currently there should be
 - 491 passing tests
 
 All integration testes are in one category, compared to being split ino short and long in DataStax driver. Instead of that, there are split depending on the state of the supported features in the driver.

Additionally I wasn't able to do other things related to integration tests. For this reason I created following issues: #45 #44.
Probably there should be also other issue addressing currently removed tests. 

Thanks to the @wprzytula  suggestion I changed from datastax ccm to scylla ccm, which helped to solve problem mentioned in #45. Thanks to that I also updated README with instructions on running the tests.

PR #58, opened 2024-10-31
Duration type (second attempt)

adespawn:
Second attempt of merging #5. Now in correct way and with just duration type. (Other will be in separate PR)

This is an implementation of one of the datatype used by the database. The current approch keeps the internal data of the type in object created in rust, while all the processing of this object is done in JS part (it usses mostly unchanged datastax code). Appart from diffrent behaviour in case of providing unexpected type in the constructor, everything should work the same way as in old implementation. If not, treat it as a bug.

This is a example implementation. The goal is to have some kind of example how to implement datastax custom types.
Current approach keeps the data in the rust object, but leaves all the processing that is done in the js part unchanged.
As this type is analog to CqlDuration used by the rust driver, there should be a way to create Duration object from CqlDuration, which is implemented.

### DIFF
--- a/lib/types/duration.js
+++ b/lib/types/duration.js
@@ -1,16 +1,9 @@
 "use strict";
 const Long = require("long");
 const util = require("util");
-const utils = require("../utils");
+const rust = require("../../index");
 
 /** @module types */
-
-// Reuse the same buffers that should perform slightly better than built-in buffer pool
-const reusableBuffers = {
-    months: utils.allocBuffer(9),
-    days: utils.allocBuffer(9),
-    nanoseconds: utils.allocBuffer(9),
-};
 
 const maxInt32 = 0x7fffffff;
 const longOneThousand = Long.fromInt(1000);
@@ -39,35 +32,15 @@ const iso8601AlternateRegex =
  * @constructor
  */
 function Duration(months, days, nanoseconds) {
-    /**
-     * Gets the number of months.
-     * @type {Number}
-     */
-    this.months = months;
-    /**
-     * Gets the number of days.
-     * @type {Number}
-     */
-    this.days = days;
-    /**
-     * Gets the number of nanoseconds represented as a <code>int64</code>.
-     * @type {Long}
-     */
-    this.nanoseconds =
-        typeof nanoseconds === "number"
-            ? Long.fromNumber(nanoseconds)
-            : nanoseconds;
+    // ToDo: Check types
+    this.internal = rust.Duration.new(months, days, nanoseconds);
 }
 
 Duration.prototype.equals = function (other) {
     if (!(other instanceof Duration)) {
         return false;
     }
-    return (
-        this.months === other.months &&
-        this.days === other.days &&
-        this.nanoseconds.equals(other.nanoseconds)
-    );
+    return this == other;
 };
 
 /**
@@ -75,27 +48,7 @@ Duration.prototype.equals = function (other) {
  * @returns {Buffer}
  */
 Duration.prototype.toBuffer = function () {
-    const lengthMonths = VIntCoding.writeVInt(
-        Long.fromNumber(this.months),
-        reusableBuffers.months,
-    );
-    const lengthDays = VIntCoding.writeVInt(
-        Long.fromNumber(this.days),
-        reusableBuffers.days,
-    );
-    const lengthNanoseconds = VIntCoding.writeVInt(
-        this.nanoseconds,
-        reusableBuffers.nanoseconds,
-    );
-    const buffer = utils.allocBufferUnsafe(
-        lengthMonths + lengthDays + lengthNanoseconds,
-    );
-    reusableBuffers.months.copy(buffer, 0, 0, lengthMonths);
-    let offset = lengthMonths;
-    reusableBuffers.days.copy(buffer, offset, 0, lengthDays);
-    offset += lengthDays;
-    reusableBuffers.nanoseconds.copy(buffer, offset, 0, lengthNanoseconds);
-    return buffer;
+    return this.internal.toBuffer();
 };
 
 /**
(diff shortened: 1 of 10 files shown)


wprzytula:
Let's rebase this on main once #43 is merged.

test/unit/duration-type-tests.js
```diff
@@ -124,6 +128,31 @@ describe("Duration", function () {
             });
         });
     });
+    describe(" constructor type validation", function () {
```
Stapox35, line 131: Other descriptions do not have spaces at the beginning.

test/unit/duration-type-tests.js
```diff
@@ -124,6 +128,31 @@ describe("Duration", function () {
...
+        it("should create duration type correctly", function () {
+            let keys = [0, Long.ZERO, 123];
+            keys.forEach(function (item) {
+                new Duration(0, 0, item);
+            });
+            new Duration(0, 0, 0);
+        });
+    });
+    describe(" constructor type validation (invalid types)", function () {
```
Stapox35, line 140: Same.

Stapox35 (changes requested):
Where are getters for class fields?

lib/types/duration.js
```diff
@@ -29,172 +31,243 @@ const iso8601AlternateRegex =
...
+     * @return {string}
+     */
+    toString() {
+        let nanoseconds = bigintToLong(this.#internal.getNanoseconds());
+        let value = "";
+        function append(dividend, divisor, unit) {
+            if (dividend === 0 || dividend < divisor) {
+                return dividend;
+            }
+            // string concatenation is supposed to be fasted than join()
```
ZuzaOsa, line 175: ``faster`` instead of ``fasted``

lib/types/duration.js
```diff
@@ -29,172 +31,243 @@ const iso8601AlternateRegex =
...
+     *   </ul>
+     * </li>
+     * <li>ISO 8601 format:  <code>P[n]Y[n]M[n]DT[n]H[n]M[n]S or P[n]W</code></li>
+     * <li>ISO 8601 alternative format: <code>P[YYYY]-[MM]-[DD]T[hh]:[mm]:[ss]</code></li>
+     * </ul>
+     * @param {String} input
+     * @returns {Duration}
+     */
+    static fromString(input) {
+        // This does a lot of stuff. Is it nesesery to implement it in rust, or should we just keep it as is?
```
ZuzaOsa, line 256: ``necessary`` instead of ``nesesery``

test/unit/duration-type-tests.js
```diff
@@ -124,6 +128,37 @@ describe("Duration", function () {
...
+                "0",
+                0xfffffffffffffffff,
+                { _: 1 },
+                [0],
+                BigInt("213769213769213769213769"),
+            ];
+            ns_keys.forEach(function (item) {
+                assert.throws(() => {
+                    new Duration(0, 0, item);
+                }, Error);
```
ZuzaOsa, line 152: Maybe also check if the correct error is thrown?

lib/types/duration.js
```diff
@@ -29,172 +31,243 @@ const iso8601AlternateRegex =
...
      * @type {Number}
      */
-    this.months = months;
+    get months() {
+        return this.#internal.months;
+    }
+
+    set months(_) {
+        throw new SyntaxError("Duration months is read-only");
+    }
```
ZuzaOsa, line 108: Add unit test to check if the error is correctly thrown.

ZuzaOsa (changes requested):

ZuzaOsa:
Why aren't integration tests working?

adespawn:
> Why aren't integration tests working?

They are working 😝. ~~But let's not talk about stability of the unit tests~~

test/unit/duration-type-tests.js
adespawn: no.

src/result_tests.rs
```diff
@@ -37,6 +40,17 @@ pub fn tests_get_cql_wrapper_double() -> CqlValueWrapper {
     CqlValueWrapper { inner: element }
 }
 
+#[napi]
+/// Test function returning sample CqlValueWrapper with Double type
```
ZuzaOsa, line 44: Are you sure this is a Double type?

ZuzaOsa (changes requested):

adespawn:
Old look of the ``fromString`` documentation:
image


New look:
image

ZuzaOsa (approved):
Code quality checks are not passing, otherwise lgtm.

src/types/duration.rs
```diff
@@ -0,0 +1,59 @@
...
+impl DurationWrapper {
+    #[napi]
+    pub fn new(months: i32, days: i32, ns_bigint: BigInt) -> napi::Result<Self> {
+        let (ns_value, is_lossless) = ns_bigint.get_i64();
+        if !is_lossless {
+            return Err(Error::new(
+                Status::GenericFailure,
+                "Nanoseconds must not overflow i64",
+            ));
+        }
```
wprzytula, line 21: Now I'm scared when seeing this `BigInt::get_i64()`...

src/types/duration.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    #[napi]
+    pub fn get_nanoseconds(&self) -> BigInt {
+        let tmp: i128 = self.nanoseconds.into();
+        let mut res: BigInt = BigInt::from(tmp.abs());
+        res.sign_bit = self.nanoseconds < 0;
+        res
+    }
+}
```
wprzytula, line 42: I hope Datastax's integration tests are strong enough to detect potential bugs in this code...

wprzytula: (and napi-rs code, more importantly :P)

test/unit/cql-value-wrapper.js
```diff
@@ -58,6 +59,22 @@ describe("Cql value wrapper", function () {
...
+        let type = element.getType();
+        assert.strictEqual(type, rust.CqlType.Duration);
+        let value = getCqlObject(element);
+        assert.instanceOf(value, Duration);
+        /* Corresponding value: 
+        let element = CqlValue::Duration(CqlDuration {
+        months: 1,
+        days: 2,
+        nanoseconds: 3,
+        }); */
```
wprzytula, line 73: Missing indentation in the comment, hard to read.

wprzytula (approved):

src/types/duration.rs
adespawn: there are decently strong unit tests when it comes to duration type nanoseconds ;)

wprzytula (approved):

src/types/duration.rs
wprzytula: Are we going to replace this with the new util function from #85?

adespawn: I plan to merge #85, update this PR to use those functions and then merge this PR

PR #62, opened 2024-11-04
Change to classes syntax

ZuzaOsa:
This PR will change TimeUuid type logic to use rust object underneath.

### DIFF
--- a/lib/types/time-uuid.js
+++ b/lib/types/time-uuid.js
@@ -44,217 +44,217 @@ let _ticksForCurrentTime = 0;
  */
 let _lastTimestamp = 0;
 
-/**
- * Creates a new instance of Uuid based on the parameters provided according to rfc4122.
- * If any of the arguments is not provided, it will be randomly generated, except for the date that will use the current
- * date.
- * <p>
- *   Note that when nodeId and/or clockId portions are not provided, the constructor will generate them using
- *   <code>crypto.randomBytes()</code>. As it's possible that <code>crypto.randomBytes()</code> might block, it's
- *   recommended that you use the callback-based version of the static methods <code>fromDate()</code> or
- *   <code>now()</code> in that case.
- * </p>
- * @class
- * @classdesc Represents an immutable version 1 universally unique identifier (UUID). A UUID represents a 128-bit value.
- * <p>Usage: <code>TimeUuid.now()</code></p>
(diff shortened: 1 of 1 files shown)


lib/types/time-uuid.js
```diff
@@ -44,217 +44,218 @@ let _ticksForCurrentTime = 0;
...
+     * If any of the arguments is not provided, it will be randomly generated, except for the date that will use the current
+     * date.
+     * <p>
+     *   Note that when nodeId and/or clockId portions are not provided, the constructor will generate them using
+     *   <code>crypto.randomBytes()</code>. As it's possible that <code>crypto.randomBytes()</code> might block, it's
+     *   recommended that you use the callback-based version of the static methods <code>fromDate()</code> or
+     *   <code>now()</code> in that case.
+     * </p>
+     * @class
+     * @classdesc Represents an immutable version 1 universally unique identifier (UUID). A UUID represents a 128-bit value.
```
adespawn, line 59: Move documentation about class directly before class declaration

adespawn (changes requested):

lib/types/time-uuid.js
```diff
@@ -206,13 +206,14 @@ class TimeUuid extends Uuid {
...
 
         let timeHigh = 0;
-        timeHigh |= (this.buffer[4] & 0xff) << 8;
-        timeHigh |= this.buffer[5] & 0xff;
-        timeHigh |= (this.buffer[6] & 0x0f) << 24;
-        timeHigh |= (this.buffer[7] & 0xff) << 16;
+        timeHigh |= (buffer[4] & 0xff) << 8;
+        timeHigh |= buffer[5] & 0xff;
+        timeHigh |= (buffer[6] & 0x0f) << 24;
+        timeHigh |= (buffer[7] & 0xff) << 16;
```
wprzytula, line 216: Why do we need that? Why not implement a `get`ter for `buffer`, so that all reads from the field succeed?

ZuzaOsa: Okay, I forgot about this option. This means TimeUuid only needs to be changed to classes syntax to work. I applied this changes in [PR#60](https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/pull/60) so I will close this one.

adespawn:
🤨

PR #77, opened 2024-11-09
Purge graph

adespawn:
Purge most of the remaining traces of graph from the driver


lib/datastax/graph/index.d.ts
ZuzaOsa, line 1: Wasn't this file supposed to be left empty so that TS does not compile when using graph module? (I don't know if removing the exports does the same thing).

adespawn: It should do the same thing: we don't have the declarations of the functions

lib/execution-profile.js
```diff
@@ -202,18 +173,11 @@ class ProfileManager {
...
      */
-    getOrCreateGraphOptions(profile, createHandler) {
-        let graphOptions = this._graphOptionsCache[profile.name];
-        if (!graphOptions) {
-            graphOptions = this._graphOptionsCache[profile.name] =
-                createHandler();
-        }
-        return graphOptions;
+    getOrCreateGraphOptions() {
+        throwNotSupported("ProfileManager.getOrCreateGraphOptions");
```
wprzytula, line 180: Regarding the question in the comment: let's not leave such questions in comments in merged code. We should answer them before merging.

test/unit/typescript/api-generation-test.ts
```diff
@@ -3,17 +3,15 @@ import {
...
   mapping,
   metadata,
   metrics,
   policies,
   tracker,
   types,
-} from "../../../index";
-import * as root from "../../../index";
+} from "../../../main";
+import * as root from "../../../main";
```
wprzytula, line 13: Did this test pass before this change, even though our `index` only contains Rust exports??

test/unit/typescript/client-tests.ts
```diff
@@ -4,7 +4,7 @@ import {
   ExecutionProfile,
   policies,
   types,
-} from "../../../index";
+} from "../../../main";
```
wprzytula, line 7: ditto

test/unit/typescript/api-generation-test.ts
adespawn: Probably not, as TS was throwing errors from VScode. It's likely they are not being run at all.

wprzytula (approved):

PR #78, opened 2024-11-09
Client to class

adespawn:
Update Client to use class syntax

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -30,215 +30,16 @@ const ResultSet = require("./types/result-set.js");
 const warmupLimit = 32;
 
 /**
- * Client options.
- * <p>While the driver provides lots of extensibility points and configurability, few client options are required.</p>
- * <p>Default values for all settings are designed to be suitable for the majority of use cases, you should avoid
- * fine tuning it when not needed.</p>
- * <p>See [Client constructor]{@link Client} documentation for recommended options.</p>
- * @typedef {Object} ClientOptions
- * @property {Array.<string>} contactPoints
- * Array of addresses or host names of the nodes to add as contact points.
- * <p>
- *  Contact points are addresses of Cassandra nodes that the driver uses to discover the cluster topology.
- * </p>
- * <p>
- *  Only one contact point is required (the driver will retrieve the address of the other nodes automatically),
- *  but it is usually a good idea to provide more than one contact point, because if that single contact point is
- *  unavailable, the driver will not be able to initialize correctly.
- * </p>
- * @property {String} [localDataCenter] The local data center to use.
- * <p>
- *   If using DCAwareRoundRobinPolicy (default), this option is required and only hosts from this data center are
- *   connected to and used in query plans.
- * </p>
- * @property {String} [keyspace] The logged keyspace for all the connections created within the {@link Client} instance.
- * @property {Object} [credentials] An object containing the username and password for plain-text authentication.
- * It configures the authentication provider to be used against Apache Cassandra's PasswordAuthenticator or DSE's
- * DseAuthenticator, when default auth scheme is plain-text.
- * <p>
- *   Note that you should configure either <code>credentials</code> or <code>authProvider</code> to connect to an
- *   auth-enabled cluster, but not both.
- * </p>
- * @property {String} [credentials.username] The username to use for plain-text authentication.
- * @property {String} [credentials.password] The password to use for plain-text authentication.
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -51,6 +51,7 @@ const warmupLimit = 32;
  *   If using DCAwareRoundRobinPolicy (default), this option is required and only hosts from this data center are
  *   connected to and used in query plans.
  * </p>
+ * [TODO:]
  * @property {String} [keyspace] The logged keyspace for all the connections created within the {@link Client} instance.
  * @property {Object} [credentials] An object containing the username and password for plain-text authentication.
```
ZuzaOsa, line 56: Missing TODO?

lib/client.js
```diff
@@ -311,65 +380,76 @@ const warmupLimit = 32;
...
  * <p>
  *   Use <code>null</code> to reset the value and use the default on <code>socketOptions.readTimeout</code> .
  * </p>
+ * [TODO: Delete?]
  * @property {String} [graphOptions.source] The graph traversal source name to use in graph queries. Default:
  * <code>'g'</code>.
+ * [TODO: Delete?]
  * @property {Number} [graphOptions.writeConsistency] Overrides the [consistency
  * level]{@link module:types~consistencies} defined in the query options for graph write queries.
+ * [TODO: Delete?]
```
ZuzaOsa, line 442: What does the last TODO refrence?

lib/client.js
```diff
@@ -311,65 +380,76 @@ const warmupLimit = 32;
...
  */
 
 /**
- * Creates a new instance of {@link Client}.
- * @classdesc
  * Represents a database client that maintains multiple connections to the cluster nodes, providing methods to
  * execute CQL statements.
- * <p>
```
ZuzaOsa, line 366: How removing the ``<p>`` symbol changes how the documentation looks?

lib/client.js
```diff
@@ -382,872 +462,865 @@ const warmupLimit = 32;
...
+        return promiseUtils.optionalCallback(this._connect(), callback);
     }
 
-    return promiseUtils.optionalCallback(this._connect(), callback);
-};
+    /**
+     * Async-only version of {@link Client#connect()}.
+     * @private
+     */
+    async _connect() {
```
ZuzaOsa, line 573: If the _ character is used to denote a private method (I think this was a convention in older JS, but I am not sure) we can check if switching to a [newer syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties) would work.

lib/client.js
adespawn: It's below property, not above. I can change it if would be more readable.

lib/client.js
adespawn: In VSC it show the remaining part of the doc as a new paragraph (compared to not showing it at all).
When generating with markdown plugin (``js-dock-export`` branch), it remains the same

lib/client.js
ZuzaOsa: Oh, I assumed it was above due to the semicolon at the end. I think it is more readable to have it above (``TODO: property`` is more intuitive than ``property TODO``:).

ZuzaOsa:
I think we also want to rebase ``types-support`` and ``simple-query`` on main to add the commit of changing ``new_utils`` to ``new-utils`` (look line 5 of ``client.js``).

adespawn:
> I think we also want to rebase `types-support` and `simple-query` on main to add the commit of changing `new_utils` to `new-utils` (look line 5 of `client.js`).

I forgot that those branches are still not rebased on main.

lib/client.js
```diff
@@ -311,65 +381,75 @@ const warmupLimit = 32;
...
  * </p>
+ * [TODO: Delete?]
  * @property {String} [graphOptions.source] The graph traversal source name to use in graph queries. Default:
  * <code>'g'</code>.
+ * [TODO: Delete?]
  * @property {Number} [graphOptions.writeConsistency] Overrides the [consistency
  * level]{@link module:types~consistencies} defined in the query options for graph write queries.
  */
 
 /**
```
ZuzaOsa, line 445: When I generated the docs from in-code docs I think this whole upper section describing the parameters is lost (at least for me).

ZuzaOsa:
ClientOptions documentation was moved to a different file. QueryOptions still remains, related issue #81.

ZuzaOsa (approved):

PR #80, opened 2024-11-09
Batch statements

adespawn:
(This PR depends on #40)

Add simple implementation of batch queries through rust driver. Batch query object is kept in a new wrapper, exposed to JS driver.

Currently only prepared queries are allowed due to the same limitation, as in execute parameters PR. It should be update alongside that part.

Enables working tests from ``client-batch-tests.js`` and marks not working tests with requirements to enable them. Extended time limit for those test tests, due to occasional timeout on Github actions. It's open to discussion if new timeout is the right value.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -20,7 +20,7 @@ const PrepareHandler = require("./prepare-handler");
 const InsightsClient = require("./insights-client");
 const cloud = require("./datastax/cloud");
 const promiseUtils = require("./promise-utils");
-const { SessionOptions, SessionWrapper } = require("../index.js");
+const { SessionOptions, SessionWrapper, createBatch } = require("../index.js");
 const ResultSet = require("./types/result-set.js");
 const { parseParams } = require("./types/cql-utils.js");
 
@@ -585,7 +585,7 @@ class Client extends events.EventEmitter {
         }
 
         return promiseUtils.optionalCallback(
-            this.#batch(queries, options),
+            this.#rustyBatch(queries, options),
             callback,
         );
     }
@@ -597,7 +597,7 @@ class Client extends events.EventEmitter {
      * @returns {Promise<ResultSet>}
      * @private
      */
-    async #batch(queries, options) {
+    async #rustyBatch(queries, options) {
         if (!Array.isArray(queries)) {
             throw new errors.ArgumentError("Queries should be an Array");
         }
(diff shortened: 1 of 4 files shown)


lib/client.js
```diff
@@ -184,7 +185,7 @@ class Client extends events.EventEmitter {
      */
     constructor(options) {
         super();
-        let rust_options = SessionOptions.empty();
+        let rust_options = rust.SessionOptions.empty();
```
ZuzaOsa, line 188: Camel case

lib/client.js
```diff
@@ -307,7 +308,7 @@ class Client extends events.EventEmitter {
         );
 
         try {
-            this.rust_client = await SessionWrapper.createSession(
+            this.rust_client = await rust.SessionWrapper.createSession(
```
ZuzaOsa, line 311: (same) Camel case

lib/client.js
```diff
@@ -607,45 +607,29 @@ class Client extends events.EventEmitter {
...
+                prepared_queries.push(prepared);
+                parameters_rows.push(parameters);
             }
         }
 
-        const request = await this.#createBatchRequest(queryItems, execOptions);
-        return await RequestHandler.send(request, execOptions, this);
+        let batch = createBatch(prepared_queries);
+        let rusty = await this.rust_client.queryBatch(batch, parameters_rows);
+        return new ResultSet(rusty);
```
wprzytula: The `rusty` name says nothing to me about what this variable contains.

src/request.rs
```diff
@@ -140,6 +140,12 @@ impl QueryParameterWrapper {
     }
 }
 
+impl QueryParameterWrapper {
+    pub(crate) fn map_to_row(row: Vec<&QueryParameterWrapper>) -> Vec<CqlValue> {
+        row.iter().map(|e| e.parameter.clone()).collect()
+    }
+}
```
wprzytula, line 147: `map_to_row` - what does it mean? IIUC, it just clones the Vec, unwrapping and cloning the `CqlValue` on the way.

src/session.rs
```diff
@@ -68,12 +73,39 @@ impl SessionWrapper {
         query: &PreparedStatementWrapper,
         params: Vec<&QueryParameterWrapper>,
```
wprzytula, line 74: I don't feel I understand how we can pass the JS `Array<T` as `Vec<&T>`. Please investigate how it works.

If this works **and behaves correctly**, then I conclude that we can pass the `Array<T>` as `&[T]`, too. In such case let's do that to save allocations.

wprzytula (changes requested):

src/session.rs
adespawn: The trait ``FromNapiValue`` required for the value to be used as a napi function argument is implemented only for ``Vec``. For this reason using slices doesn't work.

adespawn: Here is the implementation of this trait for Vec and I see no reason this will work differently then expected:

https://github.com/napi-rs/napi-rs/blob/7815851b1ebc1e55c8af6e5b4d428fcfb2e11e02/crates/napi/src/bindgen_runtime/js_values/array.rs#L248-L270

lib/client.js
```diff
@@ -607,45 +607,29 @@ class Client extends events.EventEmitter {
...
-                }
 
-                const query = typeof item === "string" ? item : item.query;
-                if (!query) {
-                    throw new errors.ArgumentError(
-                        `Invalid query at index ${i}`,
-                    );
-                }
-
-                queryItems[i] = { query, params: item.params };
```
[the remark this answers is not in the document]
adespawn: It's undefined ;)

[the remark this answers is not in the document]
adespawn: So it behaves (almost / the same) as if no parameters were provided.

src/session.rs
wprzytula: Now that I think of it, it makes perfect sense. After all, JS `Array<T>` is most probably represented in the memory differently than `&[T]`; at least, objects are held there by reference, not by value.

wprzytula: But then, if `Array<T>` is physically an array of pointers, then we could convert it to `&[&T]`. If it's true, then I believe adding conversion of `Array<T>` -> `&[&T]` to napi-rs would be an amazing achievement of your project/thesis.

lib/client.js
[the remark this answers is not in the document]
wprzytula: I'll trust you that you understand JS.

src/session.rs
adespawn: I've added issue for that.

src/request.rs
adespawn: I've updated this function name to something more (hopefully) meaningfull

wprzytula (approved):

PR #60, opened 2024-11-03
UUID type

ZuzaOsa:
This is implementation of the ``Uuid`` and ``Timeuuid`` types used by the database. It introduces wrappers for the types in rust - ``TimeUuidWrapper`` and ``UuidWrapper``. Current implementation holds data in a Rust object and most processing is done in JS. While creating an uuid it is now checked using a regex instead of checking only if the length of the given buffer is correct. There is a ``get``ter and a ``set``ter introduced for the buffer attribute. As it is read-only using a ``set``ter throws a ``SyntaxError`` exception.

PR #82, opened 2024-11-12
Disable broken examples

adespawn:
It appears that making all examples will take more effort than I expected, so for this reason I disabled currently broken examples, to see on CI those examples that currently should work.


ZuzaOsa:
What needs to be implemented for the examples to work?

adespawn:
User defined types, tuple,  Resultset info filed (#37) and mapper and changes in (#40 and #60)

ZuzaOsa (approved):

PR #85, opened 2024-11-13
Util functions for error throwing and bigint to int conversion

adespawn:
Regarding functions for error conversion:
If at some point in the future we decide that converting errors would work differently
having just one function would require updating existing code. For this reason
I created a separate function, although they currently have the same logic


src/utils.rs
```diff
@@ -0,0 +1,25 @@
+use std::{error::Error, fmt::Display};
+
+use napi::{bindgen_prelude::BigInt, Status};
+
+/// Convert rust error to napi::Error
+pub(crate) fn err_to_napi<T: Error>(e: T) -> napi::Error {
+    js_error(e)
+}
+
+/// Crete napi::Error error from a message
```
ZuzaOsa, line 10: ``Create`` instead of ``Crete``

ZuzaOsa: I think the specification that ``napi::Error`` is an error is not needed.

src/utils.rs
```diff
@@ -0,0 +1,25 @@
...
+pub(crate) fn js_error<T: Display>(e: T) -> napi::Error {
+    napi::Error::new(Status::GenericFailure, e.to_string())
+}
+
+// This is util function that will be used in #58 and #49 PR
```
ZuzaOsa, line 15: ``a util function``

src/utils.rs
```diff
@@ -0,0 +1,25 @@
...
+#[allow(dead_code)]
+pub(crate) fn bigint_to_i64<T: Display>(value: BigInt, error_msg: T) -> napi::Result<i64> {
+    let (i64_value, is_lossless) = value.get_i64();
+    // If i64 is negative, this means we would loose first bit of information,
+    // as value is converted from u64 without checking sign bit
```
ZuzaOsa, line 20: Imo ``this means`` is not needed here and ``checking the sign bit`` would sound better.

Cargo.toml
```diff
@@ -8,7 +8,7 @@ crate-type = ["cdylib"]
 
 [dependencies]
 # Default enable napi4 feature, see https://nodejs.org/api/n-api.html#node-api-version-matrix
```
ZuzaOsa, line 10: This comment is outdated now? I do not quite understand it (and it's purpose).

ZuzaOsa (changes requested):

Cargo.toml
adespawn: some leftovers from napi auto-generated template

wprzytula (changes requested):

Cargo.toml
```diff
@@ -7,8 +7,7 @@ version = "0.0.0"
 crate-type = ["cdylib"]
 
 [dependencies]
-# Default enable napi4 feature, see https://nodejs.org/api/n-api.html#node-api-version-matrix
-napi = { version = "2.12.2", default-features = false, features = ["napi4", "async"] }
+napi = { version = "2.12.2", default-features = false, features = ["napi4", "napi6", "async"] }
```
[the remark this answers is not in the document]
adespawn: Bigint

src/utils.rs
```diff
@@ -0,0 +1,25 @@
...
+    js_error(e)
+}
+
+/// Create napi::Error from a message
+pub(crate) fn js_error<T: Display>(e: T) -> napi::Error {
+    napi::Error::new(Status::GenericFailure, e.to_string())
+}
+
+// This is a util function that will be used in #58 and #49 PR
+#[allow(dead_code)]
```
ZuzaOsa, line 16: Remember to change this to ``expect(dead_code)``

src/utils.rs
```diff
@@ -1,13 +1,25 @@
...
+pub(crate) fn bigint_to_i64<T: Display>(value: BigInt, error_msg: T) -> napi::Result<i64> {
+    let (i64_value, is_lossless) = value.get_i64();
+    // If i64 is negative, we would loose first bit of information,
+    // as value is converted from u64 without checking the sign bit
+    if !is_lossless || i64_value < 0 {
+        return Err(js_error(error_msg));
+    }
```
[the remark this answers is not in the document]
adespawn: Here I try to go around the bug and this currently works for all values except i64::MIN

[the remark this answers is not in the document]
wprzytula: Can you work around the bug by completely ignoring `get_i64` and using your own correct conversion?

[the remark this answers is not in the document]
adespawn: Sure.

src/utils.rs
```diff
@@ -11,3 +11,17 @@ pub(crate) fn err_to_napi<T: Error>(e: T) -> napi::Error {
...
+pub(crate) fn bigint_to_i64(value: BigInt, error_msg: impl Display) -> napi::Result<i64> {
+    // Currently BigInt.get_i64() doesn't work as intended, so for now convert it manually
+    if value.words.len() != 1 || value.words[0] > i64::MAX as u64 {
+        if value.sign_bit && value.words.first().unwrap_or(&0) == &i64::MIN.unsigned_abs() {
+            return Ok(i64::MIN);
+        }
+        return Err(js_error(error_msg));
+    }
+    Ok(value.words[0] as i64 * if value.sign_bit { -1 } else { 1 })
+}
```
wprzytula, line 27: Let's have a suite of unit tests for this. Just in case.

ZuzaOsa (approved):
One small comment, otherwise lgtm.

PR #87, opened 2024-11-14
Reformat client docs

ZuzaOsa:
Remove html tags from documentation, replace with markdown. ``Client``, ``ClientOptions`` and ``QueryOptions`` were updated.

### DIFF
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -10,35 +10,39 @@ const auth = require("./auth");
 
 /**
  * Client options.
- * <p>While the driver provides lots of extensibility points and configurability, few client options are required.</p>
- * <p>Default values for all settings are designed to be suitable for the majority of use cases, you should avoid
- * fine tuning it when not needed.</p>
- * <p>See [Client constructor]{@link Client} documentation for recommended options.</p>
+ *
+ * While the driver provides lots of extensibility points and configurability, few client options are required.
+ *
+ * Default values for all settings are designed to be suitable for the majority of use cases, you should avoid
+ * fine tuning it when not needed.
+ *
+ * See [Client constructor]{@link Client} documentation for recommended options.
+ *
  * @typedef {Object} ClientOptions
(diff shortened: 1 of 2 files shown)


adespawn:
We need some automatic way of building and sharing documentation, but that's a separate issue...

adespawn:
In the meantime can you share before and after (ex specific HTML files) to speed up a review?

ZuzaOsa:
> In the meantime can you share before and after (ex specific HTML files) to speed up a review?

Unfortunately the HTMLs do not display properly if they are opened separately from the other files. Probably the easiest solution is to generate docs from ``simple-query`` and from this PR to compare them locally.

wprzytula:
> We need some automatic way of building and sharing documentation, but that's a separate issue...

Take a look at docs auto deployment in Rust driver:
- https://github.com/scylladb/scylla-rust-driver/blob/main/.github/workflows/docs-pages.yaml
- https://github.com/scylladb/scylla-rust-driver/blob/main/.github/workflows/docs-pr.yaml

lib/client-options.js
```diff
@@ -10,35 +10,38 @@ const auth = require("./auth");
...
+ *
  * @property {String} [localDataCenter] The local data center to use.
- * <p>
- *   If using DCAwareRoundRobinPolicy (default), this option is required and only hosts from this data center are
- *   connected to and used in query plans.
- * </p>
+ *
+ * If using DCAwareRoundRobinPolicy (default), this option is required and only hosts from this data center are
+ * connected to and used in query plans.
+ *
```
adespawn, line 35: It's my mistake, but this field is missing TODO. Only ``contactPoints`` is somehow supported

adespawn (approved):

lib/client-options.js
ZuzaOsa: Ok, I fixed that.

PR #92, opened 2024-11-16
Basic query options

adespawn:
Add wrapper for query options. Passes those options from JS code to rust. Make use of some basic options: 

- consistency
- is_idempotent
- fetch_size
- timestamp (would require #93 to function properly)
- trace_query (doesn't take use of the trace)

Reorganize rust request structure, by moving code related to request logic into separate folder.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -18,126 +18,7 @@ const promiseUtils = require("./promise-utils");
 const rust = require("../index");
 const ResultSet = require("./types/result-set.js");
 const { parseParams } = require("./types/cql-utils.js");
-
-/**
- * Query options
- * @typedef {Object} QueryOptions
- * [TODO: Add support for this field]
- * @property {boolean} [autoPage] Determines if the driver must retrieve the following result pages automatically.
- *
- * This setting is only considered by the [Client#eachRow()]{@link Client#eachRow} method. For more information,
- * check the
- * [paging results documentation]{@link https://docs.datastax.com/en/developer/nodejs-driver/latest/features/paging/}.
- *
- * [TODO: Add support for this field]
- * @property {boolean} [captureStackTrace] Determines if the stack trace before the query execution should be
- * maintained.
- *
- * Useful for debugging purposes, it should be set to ``false`` under production environment as it adds an
- * unnecessary overhead to each execution.
- *
- * Default: false.
- * [TODO: Add support for this field]
- * @property {number} [consistency] [Consistency level]{@link module:types~consistencies}.
- *
- * Defaults to ``localOne`` for Apache Cassandra and DSE deployments.
- * For DataStax Astra, it defaults to ``localQuorum``.
- *
- * [TODO: Add support for this field]
- * @property {Object} [customPayload] Key-value payload to be passed to the server. On the Cassandra side,
- * implementations of QueryHandler can use this data.
- * [TODO: Add support for this field]
- * @property {string|ExecutionProfile} [executionProfile] Name or instance of the [profile]{@link ExecutionProfile} to
- * be used for this execution. If not set, it will the use "default" execution profile.
- * [TODO: Add support for this field]
- * @property {number} [fetchSize] Amount of rows to retrieve per page.
- * [TODO: Add support for this field]
- * @property {Array|Array<Array>} [hints] Type hints for parameters given in the query, ordered as for the parameters.
- *
- * For batch queries, an array of such arrays, ordered as with the queries in the batch.
- * [TODO: Add support for this field]
- * @property {Host} [host] The host that should handle the query.
- *
- * Use of this option is **heavily discouraged** and should only be used in the following cases:
- *
- * 1. Querying node-local tables, such as tables in the ``system`` and ``system_views`` keyspaces.
- * 2. Applying a series of schema changes, where it may be advantageous to execute schema changes in sequence on the
- *    same node.
- *
- * Configuring a specific host causes the configured
- * [LoadBalancingPolicy]{@link module:policies/loadBalancing~LoadBalancingPolicy} to be completely bypassed.
- * However, if the load balancing policy dictates that the host is at a
- * [distance of ignored]{@link module:types~distance} or there is no active connectivity to the host, the request will
- * fail with a [NoHostAvailableError]{@link module:errors~NoHostAvailableError}.
- *
- * [TODO: Add support for this field]
- * @property {boolean} [isIdempotent] Defines whether the query can be applied multiple times without changing the result
- * beyond the initial application.
- *
- * The query execution idempotence can be used at [RetryPolicy]{@link module:policies/retry~RetryPolicy} level to
- * determine if an statement can be retried in case of request error or write timeout.
- *
- * Default: ``false``.
- *
- * [TODO: Add support for this field]
- * @property {string} [keyspace] Specifies the keyspace for the query. It is used for the following:
- *
- * 1. To indicate what keyspace the statement is applicable to (protocol V5+ only).  This is useful when the
- * query does not provide an explicit keyspace and you want to override the current {@link Client#keyspace}.
- * 2. For query routing when the query operates on a different keyspace than the current {@link Client#keyspace}.
- *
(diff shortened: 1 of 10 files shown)


src/session.rs
```diff
@@ -83,21 +88,54 @@ impl SessionWrapper {
         &self,
         query: &PreparedStatementWrapper,
         params: Vec<Option<&QueryParameterWrapper>>,
-        _options: &QueryParameterWrapper,
+        options: &QueryOptionsWrapper,
     ) -> napi::Result<QueryResultWrapper> {
```
wprzytula, line 92: I believe the type was wrong in the previous commit. Let's fix it.

src/session.rs
```diff
@@ -83,21 +88,54 @@ impl SessionWrapper {
...
+    }
+    if let Some(o) = options.is_idempotent {
+        query.set_is_idempotent(o);
+    }
+    if let Some(o) = options.fetch_size {
+        if o.is_negative() {
+            return Err(js_error("fetch size cannot be negative"));
+        }
+        query.set_page_size(o);
+    }
```
wprzytula, line 125: One important note about `fetch_size`:

- In Rust driver, we used to have `page_size` as `Option<i32>`. `None` meant no paging, and `Some(n)` would use pages of size `n`. This approach showed to be error-prone, so we changed it in a way that the page size is always some positive number, and it's the execution method that chooses between paged and unpaged queries (`query_unpaged` vs `query_single_page` and `query_iter`).
- in JS driver, what was true for Rust driver in the past may still be true, i.e. positive page size might imply paging and `0` page size - no paging.

Please make sure to support those semantics properly. Not to forget about it, I suggest opening an issue for that.

test/integration/supported/client-execute-tests.js
```diff
@@ -795,7 +794,7 @@ describe("Client @SERVER_API", function () {
                 done,
             );
         }); */
-        // No support for consistency levels
+        // No support for ExecutionProfile levels
```
wprzytula, line 797: `levels` was part of `consistency levels`, should now be removed.

wprzytula:
Please re-request review from me once PRs this PR depends on are merged to `execute-parameters` and this gets rebased.

src/session.rs
```diff
@@ -131,6 +135,43 @@ pub fn create_batch(queries: Vec<&PreparedStatementWrapper>) -> BatchWrapper {
     BatchWrapper { inner: batch }
 }
 
+fn apply_options(
+    mut query: PreparedStatement,
+    options: &QueryOptionsWrapper,
+) -> napi::Result<PreparedStatement> {
```
wprzytula, line 141: :wrench: naming nit again: please use `prepared` as the name for `PreparedStatement` variables; `query` should be reserved for unprepared statements.

src/session.rs
```diff
@@ -131,6 +135,43 @@ pub fn create_batch(queries: Vec<&PreparedStatementWrapper>) -> BatchWrapper {
...
+        query.set_timestamp(Some(bigint_to_i64(
+            o.clone(),
+            "Timestamp cannot overflow i64",
+        )?));
+    }
+    // TODO: Update it to allow collection of information from traced query
+    // Currently it's just passing the value, but not able to access any tracing information
+    if let Some(o) = options.trace_query {
+        query.set_tracing(o);
+    }
```
wprzytula, line 170: :question: IIUC, we will be able to turn on tracing, but we won't be able to read the returned tracing data?

src/session.rs
```diff
@@ -131,6 +135,43 @@ pub fn create_batch(queries: Vec<&PreparedStatementWrapper>) -> BatchWrapper {
...
+fn apply_options(
+    mut query: PreparedStatement,
+    options: &QueryOptionsWrapper,
+) -> napi::Result<PreparedStatement> {
+    if let Some(o) = options.consistency {
+        query.set_consistency(
+            Consistency::try_from(o)
+                .map_err(|_| js_error(format!("Unknown consistency value: {o}")))?,
+        );
+    }
```
wprzytula, line 147: Let's add support for `serial_consistency`, too, as it's going to be fully analogous.

lib/query-options.js
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * Useful for debugging purposes, it should be set to ``false`` under production environment as it adds an
+ * unnecessary overhead to each execution.
+ *
+ * Default: false.
+ * [TODO: Add support for this field]
+ * @property {Number} [consistency] [Consistency level]{@link module:types~consistencies}.
+ *
+ * Defaults to ``localOne`` for Apache Cassandra and DSE deployments.
+ * For DataStax Astra, it defaults to ``localQuorum``.
```
wprzytula, line 25: :wrench: Is this TODO justified? This PR already makes the statement set the chosen consistency level.
Also, the defaults are no longer true, because Rust driver has different defaults and you override them with no default value on the JS side.
Let's remove mentions of DSE and Astra; Rust driver has one default `Consistency` for all cases.

wprzytula (changes requested):

wprzytula (changes requested):

src/requests/request.rs
```diff
@@ -7,6 +8,32 @@ pub struct PreparedStatementWrapper {
     pub(crate) prepared: PreparedStatement,
 }
 
+#[napi]
+pub struct QueryOptionsWrapper {
+    pub auto_page: Option<bool>,
+    pub capture_stack_trace: Option<bool>,
+    pub consistency: Option<u16>,
+    pub counter: Option<bool>,
```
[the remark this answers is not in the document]
adespawn: All fields here are one to one as in query options object in js. All of the fields are described there. I didn't see the need to duplicate this information in rust part of the code, but if you believe it's nesesery I can always change that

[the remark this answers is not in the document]
wprzytula: OK, documentation in JS is enough. For some reason my brain concluded that the first thing to do in order to find out the meaning is to ask here, instead of looking up in the JS docs. Nvm.

src/session.rs
```diff
@@ -143,6 +147,51 @@ pub fn create_batch(queries: Vec<&PreparedStatementWrapper>) -> BatchWrapper {
     BatchWrapper { inner: batch }
 }
 
+fn apply_options(
+    mut prepared: PreparedStatement,
+    options: &QueryOptionsWrapper,
+) -> napi::Result<PreparedStatement> {
```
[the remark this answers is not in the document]
adespawn: With major changes to unprepared queries in #140 I would prefer to add it immediately after merging both of those PR to avoid any conflicts

src/requests/request.rs
```diff
@@ -19,3 +46,27 @@ impl PreparedStatementWrapper {
...
+            logged: None,
+            prepare: None,
+            read_timeout: None,
+            routing_indexes: None,
+            routing_names: None,
+            serial_consistency: None,
+            timestamp: None,
+            trace_query: None,
+        }
+    }
```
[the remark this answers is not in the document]
adespawn: Yes. There is something called ExecutionProfiles, which is default settings for session. It's likely this object at a later stage will keep also queryOptions before applying the ExecutionProfiles, and in this case keeping all the fields as options is nesesery.

lib/client.js
```diff
@@ -47,14 +47,7 @@ const { parseParams } = require("./types/cql-utils.js");
...
  * [TODO: Add support for this field]
- * @property {string} [executeAs] The user or role name to act as when executing this statement.
- *
- * When set, it executes as a different user/role than the one currently authenticated (a.k.a. proxy execution).
- *
- * This feature is only available in DSE 5.1+.
- *
- * [TODO: Add support for this field Delete?]
- * @property {string|ExecutionProfile} [executionProfile] Name or instance of the [profile]{@link ExecutionProfile} to
+ * @property {String|ExecutionProfile} [executionProfile] Name or instance of the [profile]{@link ExecutionProfile} to
```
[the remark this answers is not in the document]
adespawn: No. We agreed some time ago that all simple typenames should be lowercase, so this change was a mistake

PR #94, opened 2024-11-16
Client cleanup

adespawn:
A number of minor changes related to code style, mostly focused on client class.


lib/client.js
```diff
@@ -587,7 +586,7 @@ class Client extends events.EventEmitter {
 
         await this.#connect();
 
-        const execOptions = DefaultExecutionOptions.create(options, this);
+        const _execOptions = DefaultExecutionOptions.create(options, this);
```
wprzytula, line 589: Why do we rename this constant?

I see we renamed it, but I can see no changes in code that uses it. Actually, is any code using it? If not, maybe it's better to delete this instead of renaming?

lib/client.js
```diff
@@ -784,21 +783,24 @@ class Client extends events.EventEmitter {
...
+            let parsedParams = parseParams(statement, params);
+            let rusty = await this.rustClient.executePrepared(
+                statement,
+                parsedParams,
+            );
+            resolve(new ResultSet(rusty));
+        } catch (e) {
+            reject(e);
+        }
+    }
```
wprzytula, line 803: Justify the refactor here.

lib/client.js
adespawn: We will have to use this variable in near future, so I just left it here. Rename was due to eslint rules, which I updated to allow unused variables to start with ``_``.

lib/client.js
adespawn: ESLint doesn't allow previous version of the code:
https://eslint.org/docs/latest/rules/no-async-promise-executor

In the future we can consider doing it the opposite way (using async instead of promise).

src/session.rs
```diff
@@ -52,6 +54,14 @@ impl SessionWrapper {
         Ok(SessionWrapper { internal: s })
     }
 
+    #[napi]
+    pub fn get_keyspace(&self) -> Option<String> {
+        self.internal.get_keyspace().map(|v| {
+            let r: &String = v.borrow();
+            r.to_owned()
+        })
+    }
```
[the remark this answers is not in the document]
adespawn: It's ``Arc<String>``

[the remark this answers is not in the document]
wprzytula: OK, then this should work:
```rust
self.internal
  .get_keyspace()
  .as_deref()
  .map(ToOwned::to_owned)
```

lib/client.js
```diff
@@ -641,11 +641,13 @@ class Client extends events.EventEmitter {
...
     shutdown(callback) {
-        return promiseUtils.optionalCallback(this._shutdown(), callback);
+        return promiseUtils.optionalCallback(this.#shutdown(), callback);
     }
 
     /** @private */
-    async _shutdown() {
+    async #shutdown() {
+        // @wprzytula: Is this true: I wasn't able any way to do it on rust client?
+        this.log("warning", "Rust driver doesn't support explicite shutdown?");
```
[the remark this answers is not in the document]
adespawn: In this case, warning should be the best thing we can do for now

ZuzaOsa (approved):
lgtm

wprzytula (approved):

PR #95, opened 2024-11-17
Miscellaneous types

adespawn:
Adds support for complex data types (types that hold values of different types).

Add supports for following cql types:

 - BigInt
 - List
 - Map
 - Timestamp

### DIFF
--- a/lib/types/cql-utils.js
+++ b/lib/types/cql-utils.js
@@ -197,3 +197,6 @@ function parseParams(preparedStatement, params) {
 }
 
 module.exports.parseParams = parseParams;
+
+// For unit test usage
+module.exports.getWrapped = getWrapped;
--- a/lib/types/cql-utils.js
+++ b/lib/types/cql-utils.js
@@ -7,6 +7,7 @@ const { ResponseError } = require("../errors");
 const TimeUuid = require("./time-uuid");
 const Uuid = require("./uuid");
 const Duration = require("./duration");
+const LocalTime = require("./local-time");
 
 /**
  * Ensures the value isn't one of many ways to express null value
@@ -136,6 +137,14 @@ function getWrapped(type, value) {
             return rust.QueryParameterWrapper.fromMap(encodedMap);
         case rust.CqlType.SmallInt:
             return rust.QueryParameterWrapper.fromSmallInt(value);
+        case rust.CqlType.Time:
+            if (!(value instanceof LocalTime))
+                throw new TypeError(
+                    "Expected LocalTime type to parse into Cql Time",
+                );
+            return rust.QueryParameterWrapper.fromLocalTime(
+                value.getInternal(),
+            );
         case rust.CqlType.TinyInt:
             return rust.QueryParameterWrapper.fromTinyInt(value);
         case rust.CqlType.Uuid:
--- a/lib/types/cql-utils.js
+++ b/lib/types/cql-utils.js
@@ -28,7 +28,7 @@ function getWrapped(type, value) {
         case rust.CqlType.Boolean:
             return rust.QueryParameterWrapper.fromBoolean(value);
         case rust.CqlType.Counter:
-            return rust.QueryParameterWrapper.fromCounter(value);
+            return rust.QueryParameterWrapper.fromCounter(BigInt(value));
         case rust.CqlType.Double:
             return rust.QueryParameterWrapper.fromDouble(value);
         case rust.CqlType.Duration:
--- a/lib/types/cql-utils.js
+++ b/lib/types/cql-utils.js
@@ -50,6 +50,9 @@ function getWrapped(type, value) {
         case rust.CqlType.List:
             // This will be added later in this PR
             throw new TypeError("Parsing lists is not yet supported");
+        case rust.CqlType.Map:
+            // This will be added later in this PR
+            throw new TypeError("Parsing maps is not yet supported");
         case rust.CqlType.SmallInt:
             return rust.QueryParameterWrapper.fromSmallInt(value);
         case rust.CqlType.TinyInt:
--- a/lib/types/cql-utils.js
+++ b/lib/types/cql-utils.js
@@ -45,6 +45,8 @@ function getWrapped(type, value) {
             return rust.QueryParameterWrapper.fromSet(value);
         case rust.CqlType.Text:
             return rust.QueryParameterWrapper.fromText(value);
+        case rust.CqlType.Timestamp:
+            return rust.QueryParameterWrapper.fromTimestamp(BigInt(value));
         case rust.CqlType.List:
             // This will be added later in this PR
             throw new TypeError("Parsing lists is not yet supported");
--- a/lib/types/cql-utils.js
+++ b/lib/types/cql-utils.js
@@ -45,6 +45,9 @@ function getWrapped(type, value) {
             return rust.QueryParameterWrapper.fromSet(value);
         case rust.CqlType.Text:
             return rust.QueryParameterWrapper.fromText(value);
+        case rust.CqlType.List:
+            // This will be added later in this PR
+            throw new TypeError("Parsing lists is not yet supported");
         case rust.CqlType.SmallInt:
             return rust.QueryParameterWrapper.fromSmallInt(value);
         case rust.CqlType.TinyInt:
--- a/lib/types/cql-utils.js
+++ b/lib/types/cql-utils.js
@@ -8,6 +8,7 @@ const TimeUuid = require("./time-uuid");
 const Uuid = require("./uuid");
 const Duration = require("./duration");
 const LocalTime = require("./local-time");
+const InetAddress = require("./inet-address");
 
 /**
  * Ensures the value isn't one of many ways to express null value
@@ -119,6 +120,12 @@ function getWrapped(type, value) {
             return rust.QueryParameterWrapper.fromText(value);
         case rust.CqlType.Timestamp:
             return rust.QueryParameterWrapper.fromTimestamp(BigInt(value));
+        case rust.CqlType.Inet:
+            if (!(value instanceof InetAddress))
+                throw new TypeError(
+                    "Expected InetAddress type to parse into Cql Inet",
+                );
+            return rust.QueryParameterWrapper.fromInet(value.getInternal());
         case rust.CqlType.List:
             let encodedList = encodeListLike(value, type.getFirstSupportType());
             return rust.QueryParameterWrapper.fromList(encodedList);
--- a/lib/types/cql-utils.js
+++ b/lib/types/cql-utils.js
@@ -82,7 +82,7 @@ function getWrapped(type, value) {
 /**
  * Parses array of params into rust objects according to preparedStatement expected types
  * Throws ResponseError when received different amount of parameters than expected
- * @param {PreparedStatementWrapper} preparedStatement
+ * @param {rust.PreparedStatementWrapper} preparedStatement
  * @param {Array<any>} params
  * @returns {Array<rust.QueryParameterWrapper>}
  */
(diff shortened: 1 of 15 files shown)


src/result.rs
```diff
@@ -128,11 +128,16 @@ impl QueryResultWrapper {
...
-        let s: Vec<CqlValueWrapper> = self
+    pub fn get_columns(&self) -> napi::Result<Vec<Option<CqlValueWrapper>>> {
+        let s: Vec<Option<CqlValueWrapper>> = self
             .internal
             .iter()
-            .filter_map(|f| f.as_ref().map(|r| CqlValueWrapper { inner: r.clone() }))
+            .map(|f| {
+                if f.is_none() {
+                    return None;
+                }
```
PiotrJunior, line 138: Is this necessary?
.map(f) on Option<T> returns None if f  is None and f(T) in other case

lib/types/cql-utils.js
```diff
@@ -0,0 +1,90 @@
...
+        case rust.CqlType.Set:
+            return rust.QueryParameterWrapper.fromSet(value);
+        case rust.CqlType.Text:
+            return rust.QueryParameterWrapper.fromText(value);
+        case rust.CqlType.Timestamp:
+            return rust.QueryParameterWrapper.fromTimestamp(BigInt(value));
+        case rust.CqlType.List:
+            throw new TypeError("Parsing lists is not yet supported");
+        case rust.CqlType.Map:
+            throw new TypeError("Parsing maps is not yet supported");
```
PiotrJunior, line 49: maybe add issue for that to not forget about it

lib/types/results-wrapper.js
```diff
@@ -37,6 +42,21 @@ function getCqlObject(field) {
...
+            let listFields = field.getList();
+            for (let i = 0; i < listFields.length; i++)
+                list.push(getCqlObject(listFields[i]));
+            return list;
+        case rust.CqlType.Map:
+            let mapKeys = field.getMap(true);
+            let mapValues = field.getMap(false);
+            let map = {};
+            for (let i = 0; i < mapKeys.length; i++)
+                map[getCqlObject(mapKeys[i])] = getCqlObject(mapValues[i]);
```
PiotrJunior, line 58: I think it's better to use .map on `list` and maybe on `map`

PiotrJunior (changes requested):

lib/types/cql-utils.js
```diff
@@ -7,19 +7,76 @@ const TimeUuid = require("./time-uuid");
 const Uuid = require("./uuid");
 const Duration = require("./duration");
 
+/**
+ * Ensures the value isn't one of may ways to express null value
+ * @param {string} message Message of error, in case value is "null"
```
ZuzaOsa, line 12: Why not ``Error message``? ``Message of error`` sounds like Yoda. ``if`` instead of ``in case``.

ZuzaOsa (approved):
One comment, otherwise lgtm

lib/types/cql-utils.js
```diff
@@ -7,17 +7,72 @@ const TimeUuid = require("./time-uuid");
...
+ * Wraps each of the elements into given subtype
+ * @param {Array} values
+ * @param {rust.ComplexType} subtype
+ * @returns {Array<rust.QueryParameterWrapper>}
+ */
+function encodeList(values, subtype) {
+    if (!Array.isArray(values))
+        throw new TypeError(
+            `Not a valid list value, expected Array obtained ${util.inspect(values)}`,
+        );
```
wprzytula, line 30: :question: I don't understand the syntax of `expected Array obtained [values]`.

test/integration/supported/client-execute-tests.js
```diff
@@ -211,7 +211,7 @@ describe("Client @SERVER_API", function () {
             );
         }); */
 
-        // No support for lists
+        // No support for custom Row (#117)
         /* vit("2.0", "should guess known types", function (done) {
             const client = setupInfo.client;
             const columns =
```
wprzytula, line 217: :wrench: Missing TODO.

lib/types/cql-utils.js
```diff
@@ -7,17 +7,72 @@ const TimeUuid = require("./time-uuid");
...
+        throw new TypeError(message);
+}
+
+/**
+ * Wraps each of the elements into given subtype
+ * @param {Array} values
+ * @param {rust.ComplexType} subtype
+ * @returns {Array<rust.QueryParameterWrapper>}
+ */
+function encodeList(values, subtype) {
```
wprzytula, line 26: :pick: In ScyllaDB code, List and Set both belong to the group called `Listlike` - you can use the name here, too.

lib/types/cql-utils.js
adespawn: This was taken directly form datastax code: https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/2a750ff0e11955b04161ab4a05feeda1855a827e/lib/encoder.js#L816-L819
I can update this message to something more meaningful.

lib/types/cql-utils.js
```diff
@@ -7,9 +7,64 @@ const TimeUuid = require("./time-uuid");
...
+            values[i],
+            "A collection can't contain null or unset values",
+        );
+        res.push(getWrapped(subtype, values[i]));
+    }
+    return res;
+}
+
+/**
+ *
```
ZuzaOsa, line 45: :question: Why the additional empty line, does it  change anything when generating docs?

src/result.rs
```diff
@@ -314,6 +314,20 @@ impl CqlValueWrapper {
...
+        match self.inner.as_map() {
+            Some(r) => Ok(r
+                .iter()
+                .map(|f| CqlValueWrapper {
+                    inner: if keys { f.0.clone() } else { f.1.clone() },
+                })
+                .collect()),
+            None => Err(Self::generic_error("map")),
+        }
+    }
```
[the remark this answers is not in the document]
adespawn: > Or, maybe we can always return a Vec of pairs (key, value)
returning ``Vec<(CqlValueWrapper, CqlValueWrapper)>`` is not supported by napi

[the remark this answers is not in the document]
adespawn: > Or, maybe we can always return a Vec of pairs (key, value)
returning ``Vec<(CqlValueWrapper, CqlValueWrapper)>`` is not supported by napi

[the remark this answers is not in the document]
adespawn: > Or, maybe we can always return a Vec of pairs (key, value)
returning ``Vec<(CqlValueWrapper, CqlValueWrapper)>`` is not supported by napi

src/types/type_wrappers.rs
```diff
@@ -0,0 +1,86 @@
...
+        support1: Option<ComplexType>,
+        support2: Option<ComplexType>,
+    ) -> Self {
+        ComplexType {
+            base_type,
+            support_type_1: support1.map(Box::new),
+            support_type_2: support2.map(Box::new),
+        }
+    }
+}
```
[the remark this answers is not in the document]
adespawn: To simplify creating complex type, based on the number of support types

src/result.rs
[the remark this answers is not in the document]
ZuzaOsa: Then i agree that ``get_map_keys`` and ``get_map_values`` would be more readable than passing an additional parameter.

src/types/type_wrappers.rs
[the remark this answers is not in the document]
ZuzaOsa: :question: Will we use types that use more than two supporting types?

ZuzaOsa (changes requested):

lib/types/results-wrapper.js
```diff
@@ -42,6 +45,18 @@ function getCqlObject(field) {
             return res;
         case rust.CqlType.Text:
             return field.getText();
+        case rust.CqlType.Timestamp:
+            return new Date(Number(field.getTimestamp()));
```
Stapox35, line 49: Have you test for this? 
`Cql::Timestamp` hold `i64` milliseconds (max `i64` in Rust is: `9_223_372_036_854_775_807`).
But ... A JS `Date` object can represent a maximum of  `+- 8,640,000,000,000,000` milliseconds. This is less than the `Cql::Timestamp`.

src/result_tests.rs
```diff
@@ -78,6 +85,27 @@ pub fn tests_get_cql_wrapper_text() -> CqlValueWrapper {
     CqlValueWrapper { inner: element }
 }
 
+#[napi]
+/// Test function returning sample CqlValueWrapper with Timestamp type
+pub fn tests_get_cql_wrapper_timestamp() -> CqlValueWrapper {
+    let element = CqlValue::Timestamp(CqlTimestamp(1_000_000_i64));
+    CqlValueWrapper { inner: element }
+}
```
Stapox35, line 93: Please add tests for max and min value of `CqlTimestamp`.

lib/types/results-wrapper.js
adespawn: This is how does Timestamp behave in old driver:
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/2a750ff0e11955b04161ab4a05feeda1855a827e/lib/encoder.js#L260-L262
If you believe this may be a problem, create an issue for that, and we can come back to it in second iteration

src/types/type_wrappers.rs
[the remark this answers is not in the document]
adespawn: From the types that have constant amount of supporting types: yes.
There is also tuple and UDT, but this should be updated to allow that be a person implementing encoding of those types

lib/types/cql-utils.js
```diff
@@ -42,9 +98,41 @@ function getWrapped(type, value) {
         case rust.CqlType.Int:
             return rust.QueryParameterWrapper.fromInt(value);
         case rust.CqlType.Set:
-            return rust.QueryParameterWrapper.fromSet(value);
+            // TODO:
```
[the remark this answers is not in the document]
adespawn: #127

lib/types/cql-utils.js
wprzytula: Well, if it comes from the Datastax driver, then users should probably be already used to it...

src/types/type_wrappers.rs
```diff
@@ -0,0 +1,81 @@
...
+    Custom,
+}
+
+#[napi]
+#[derive(Clone)]
+pub struct ComplexType {
+    pub base_type: CqlType,
+    pub(crate) support_type_1: Option<Box<ComplexType>>,
+    pub(crate) support_type_2: Option<Box<ComplexType>>,
+}
```
wprzytula, line 40: :wrench: This does allow for erroneous variants:
1. `ComplexType { base_type: Uuid, support_type_1: None, support_type_2: Some(...) }`
2. `ComplexType { base_type: Tuple, support_type_1: None, support_type_2: None }`

`ComplexType` would be a good representation when and only when it would permit creation of correct instantiations and would prevent creation of incorrect ones.

src/types/type_wrappers.rs
```diff
@@ -0,0 +1,81 @@
...
+            _ => {
+                return err();
+            }
+        };
+        match v {
+            Some(s) => Ok(*s.clone()),
+            None => err(),
+        }
+    }
+}
```
wprzytula, line 59: :wrench: If the valid options are only `1` and `2`, then I'm strongly against using one function for it, accepting an arbitrary integer. Let's have distinct functions for both supported types.

wprzytula: :question: Still, I don't understand the naming `support` in general. What does it mean that a `ComplexType` "has support" for some another type? It just doesn't make sense for me.

wprzytula (changes requested):

src/types/type_wrappers.rs
adespawn: Things like list or map have some other type to determine what does it have as keys / values. I called it support types for those types

PR #97, opened 2024-11-17
Remove datastax auth providers from driver

adespawn:


lib/datastax/depricated-auth/dse-gssapi-auth-provider.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * @deprecated Not supported by the driver. Usage will throw an error.
+ */
+class DseGssapiAuthProvider extends AuthProvider {
+    /**
+     * @deprecated Not supported by the driver. Usage will throw an error.
+     */
+    constructor() {
+        throwNotSupported("DseGssapiAuthProvider.constructor");
```
ZuzaOsa, line 12: You did not import throwNotSupported here.

test/unit/depricated/dse-auth.js
```diff
@@ -0,0 +1,20 @@
+"use strict";
+const assert = require("assert");
+const DsePlainTextAuthProvider = require("../../../lib/datastax/depricated-auth/dse-plain-text-auth-provider");
+const DseGssapiAuthProvider = require("../../../lib/datastax/depricated-auth/dse-gssapi-auth-provider");
+
+describe("DsePlainTextAuthProvider", function () {
+    it("tests not supported error", function () {
+        assert.throws(function () {
+            new DsePlainTextAuthProvider();
+        }, ReferenceError);
```
ZuzaOsa, line 10: Those tests work because a reference error is thrown (``ReferenceError: throwNotSupported is not defined``) but not the one that should be here. Check the error message in at least one place.

ZuzaOsa (changes requested):
Folder name should be ``deprecated`` not ``depricated``.

PiotrJunior (commented):
Why `base-dse-authenticator.js` is removed, but `base-dse-authenticator.js` is moved to `datastax/deprecated`?

adespawn:
> Why `base-dse-authenticator.js` is removed, but `base-dse-authenticator.js` is moved to `datastax/deprecated`?

You asking for the same file.
But if I understand that question correctly, those files moved to ``./datastax`` directory are exposed in the API, while the deleted file is not.

ZuzaOsa (approved):

Stapox35 (approved):

PR #105, opened 2024-11-22
UUID fixes

ZuzaOsa:


src/result.rs
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    #[napi]
+    pub fn get_time_uuid(&self) -> napi::Result<TimeUuidWrapper> {
+        match self.inner.as_timeuuid() {
+            Some(r) => Ok(TimeUuidWrapper::from_cql_time_uuid(r)),
+            None => Err(Self::generic_error("time_uuid")),
+        }
+    }
+}
```
Stapox35, line 302: Different order than in an enum.

lib/types/time-uuid.js
```diff
@@ -378,8 +378,8 @@ function getOrGenerateRandom(id, length, callback) {
  * @private
  * @param {Date} date
  * @param {Number} ticks
- * @param {String|Buffer} nodeId
- * @param {String|Buffer} clockId
+ * @param {string|Buffer} nodeId
```
PiotrJunior, line 381: if we are writing `string` in lower case then we should start writning `number` in lower case for consistency

PiotrJunior (changes requested):

lib/types/uuid.js
```diff
@@ -18,6 +18,10 @@ const uuidRegex =
  * Represents an immutable universally unique identifier (UUID). A UUID represents a 128-bit value.
  */
 class Uuid {
+    /**
+     * @type {rust.UuidWrapper}
+     * @private
+     */
```
PiotrJunior, line 24: Add proper description for people developing this code in the future

PiotrJunior (changes requested):

lib/types/uuid.js
ZuzaOsa: I do not think a comment would add a lot more decription that what can be extracted from JSDocs for this functon.

Stapox35 (approved):

PR #110, opened 2024-11-24
Add support for client options

ZuzaOsa:
Add support for basic client options: ``contact points``, ``keyspace``, ``application_name``, ``application_version``, ``credentials``.

### DIFF
--- a/lib/auth/plain-text-auth-provider.js
+++ b/lib/auth/plain-text-auth-provider.js
@@ -1,67 +1,63 @@
 "use strict";
-const util = require("util");
-
 const provider = require("./provider.js");
 const utils = require("../utils");
 const AuthProvider = provider.AuthProvider;
 const Authenticator = provider.Authenticator;
+
 /**
- * Creates a new instance of the Authenticator provider
- * @classdesc Provides plain text [Authenticator]{@link module:auth~Authenticator} instances to be used when
+ * Provides plain text [Authenticator]{@link module:auth~Authenticator} instances to be used when
  * connecting to a host.
  * @extends module:auth~AuthProvider
  * @example
  * var authProvider = new cassandra.auth.PlainTextAuthProvider('my_user', 'p@ssword1!');
  * //Set the auth provider in the clientOptions when creating the Client instance
  * const client = new Client({ contactPoints: contactPoints, authProvider: authProvider });
- * @param {String} username User name in plain text
- * @param {String} password Password in plain text
  * @alias module:auth~PlainTextAuthProvider
- * @constructor
  */
-function PlainTextAuthProvider(username, password) {
-    this.username = username;
-    this.password = password;
+class PlainTextAuthProvider extends AuthProvider {
+    /**
+     * Creates a new instance of the Authenticator provider
+     * @param {String} username User name in plain text
+     * @param {String} password Password in plain text
+     */
+    constructor(username, password) {
+        super();
+        this.username = username;
+        this.password = password;
+    }
+    /**
+     * Returns a new [Authenticator]{@link module:auth~Authenticator} instance to be used for plain text authentication.
+     * @override
+     * @returns {Authenticator}
+     */
+    newAuthenticator() {
+        return new PlainTextAuthenticator(this.username, this.password);
+    }
 }
-
-util.inherits(PlainTextAuthProvider, AuthProvider);
-
-/**
- * Returns a new [Authenticator]{@link module:auth~Authenticator} instance to be used for plain text authentication.
- * @override
- * @returns {Authenticator}
- */
-PlainTextAuthProvider.prototype.newAuthenticator = function () {
-    return new PlainTextAuthenticator(this.username, this.password);
-};
 
 /**
  * @ignore
  */
-function PlainTextAuthenticator(username, password) {
-    this.username = username;
-    this.password = password;
+class PlainTextAuthenticator extends Authenticator {
+    constructor(username, password) {
+        super();
+        this.username = username;
+        this.password = password;
+    }
+    initialResponse(callback) {
+        const initialToken = Buffer.concat([
+            utils.allocBufferFromArray([0]),
+            utils.allocBufferFromString(this.username, "utf8"),
+            utils.allocBufferFromArray([0]),
+            utils.allocBufferFromString(this.password, "utf8"),
+        ]);
+        callback(null, initialToken);
+    }
+    evaluateChallenge(challenge, callback) {
+        // noop
+        callback();
+    }
 }
-
-util.inherits(PlainTextAuthenticator, Authenticator);
-
-PlainTextAuthenticator.prototype.initialResponse = function (callback) {
-    const initialToken = Buffer.concat([
-        utils.allocBufferFromArray([0]),
-        utils.allocBufferFromString(this.username, "utf8"),
-        utils.allocBufferFromArray([0]),
-        utils.allocBufferFromString(this.password, "utf8"),
-    ]);
-    callback(null, initialToken);
-};
-
-PlainTextAuthenticator.prototype.evaluateChallenge = function (
-    challenge,
-    callback,
-) {
-    // noop
-    callback();
-};
 
 module.exports = {
     PlainTextAuthenticator,
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -45,11 +45,8 @@ const errors = require("./errors.js");
  * Note that you should configure either ``credentials`` or ``authProvider`` to connect to an
  * auth-enabled cluster, but not both.
  *
- * [TODO: Add support for this field]
  * @property {String} [credentials.username] The username to use for plain-text authentication.
- * [TODO: Add support for this field]
  * @property {String} [credentials.password] The password to use for plain-text authentication.
- * [TODO: Add support for this field]
  * @property {Uuid} [id] A unique identifier assigned to a {@link Client} object, that will be communicated to the
  * server (DSE 6.0+) to identify the client instance created with this options. When not defined, the driver will
  * generate a random identifier.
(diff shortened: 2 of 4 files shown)


lib/client.js
```diff
@@ -176,14 +176,20 @@ class Client extends events.EventEmitter {
...
         let rustOptions = rust.SessionOptions.empty();
         rustOptions.connectPoints = options.contactPoints;
+        rustOptions.keyspace = options.keyspace;
         rustOptions.applicationName = options.applicationName;
         rustOptions.applicationVersion = options.applicationVersion;
+        if (options.credentials) {
+            rustOptions.credentialsUsername = options.credentials.username;
+            rustOptions.credentialsPassword = options.credentials.password;
+        }
         this.rustOptions = rustOptions;
```
adespawn, line 186: If you moved code related to client options to separate function, maybe move it also in JS?

adespawn (changes requested):
If I understand correctly, the authentication part will work correctly if username and password will be provided, but not if the PlainTextAuthProvider as authProvider option

adespawn:
Did you checked if any integration tests can be enabled with those changes?

lib/client-options.js
```diff
@@ -36,7 +36,6 @@ const { throwNotSupported } = require("./new-utils");
  *
  * [TODO: Add support for this field]
  * @property {String} [keyspace] The logged keyspace for all the connections created within the {@link Client} instance.
```
wprzytula, line 38: :wrench: This TODO is obsolete in this commit.

src/session.rs
```diff
@@ -47,13 +49,15 @@ impl SessionOptions {
...
-        Ok(SessionWrapper { inner: s })
+        let builder = configure_session_builder(options);
+        let session = builder.build().await.map_err(err_to_napi)?;
+        if let Some(keyspace) = &options.keyspace {
+            session
+                .use_keyspace(keyspace, false)
+                .await
+                .map_err(err_to_napi)?;
+        }
+        Ok(SessionWrapper { inner: session })
```
wprzytula, line 64: :wrench: There is a `use_keyspace()` method on `SessionBuilder` that should be used instead of `use`ing the keyspace on an already set up `Session`.

src/session.rs
```diff
@@ -162,6 +166,13 @@ pub fn create_prepared_batch(queries: Vec<&PreparedStatementWrapper>) -> BatchWr
     BatchWrapper { inner: batch }
 }
 
+fn configure_session_builder(options: &SessionOptions) -> SessionBuilder {
+    let mut builder = SessionBuilder::new();
+    builder = builder.custom_identity(get_self_identity(options));
```
wprzytula, line 175: :pick: It really pains me when I see naming of `get_*` when not really referring to just retrieving a field. We should rather use `create_*`, `build_*`, `configure_*` etc., whichever you find best.

wprzytula: The same goes for `getWrappedType`. IMO, it should be `wrapType`, because the wrapped type has not existed anywhere yet -- we are creating it on-fly.

lib/client-options.js
```diff
@@ -43,11 +43,8 @@ const { throwNotSupported } = require("./new-utils");
  * Note that you should configure either ``credentials`` or ``authProvider`` to connect to an
  * auth-enabled cluster, but not both.
```
wprzytula, line 44: :pick: typo in commit message: `suport` -> `support`

src/session.rs
```diff
@@ -162,6 +170,19 @@ pub fn create_prepared_batch(queries: Vec<&PreparedStatementWrapper>) -> BatchWr
...
+    builder = builder.known_nodes(&options.connect_points);
+    if let (Some(username), Some(password)) = (
+        options.credentials_username.as_ref(),
+        options.credentials_password.as_ref(),
+    ) {
+        builder = builder.user(username, password);
+    }
```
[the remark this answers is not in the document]
wprzytula: Here we skip setting credentials **silently** if some credential is missing - this is unacceptable.

lib/client-options.js
```diff
@@ -43,11 +43,8 @@ const { throwNotSupported } = require("./new-utils");
...
  *
- * [TODO: Add support for this field]
  * @property {String} [credentials.username] The username to use for plain-text authentication.
- * [TODO: Add support for this field]
  * @property {String} [credentials.password] The password to use for plain-text authentication.
- * [TODO: Add support for this field]
  * @property {Uuid} [id] A unique identifier assigned to a {@link Client} object, that will be communicated to the
  * server (DSE 6.0+) to identify the client instance created with this options. When not defined, the driver will
  * generate a random identifier.
```
wprzytula, line 50: :wrench: You removed TODO for `id`, but I don't see where it's supported.

lib/auth/plain-text-auth-provider.js
```diff
@@ -19,50 +19,48 @@ const Authenticator = provider.Authenticator;
...
+     * Returns a new [Authenticator]{@link module:auth~Authenticator} instance to be used for plain text authentication.
+     * @override
+     * @returns {Authenticator}
+     */
+    newAuthenticator() {
+        return new PlainTextAuthenticator(this.username, this.password);
+    }
 }
 
 util.inherits(PlainTextAuthProvider, AuthProvider);
```
wprzytula, line 37: :question: Shouldn't this also be converted to the classes syntax, i.e., `extends`?

lib/auth/plain-text-auth-provider.js
```diff
@@ -19,50 +19,48 @@ const Authenticator = provider.Authenticator;
...
+        ]);
+        callback(null, initialToken);
+    }
+    evaluateChallenge(challenge, callback) {
+        // noop
+        callback();
+    }
 }
 
 util.inherits(PlainTextAuthenticator, Authenticator);
```
wprzytula, line 62: ditto

wprzytula (changes requested):

lib/client-options.js
ZuzaOsa: The TODOs are after the description not before. This was discussed with @adespawn when reviewing #78 and is done this way because it displays better in the compiled docs.

wprzytula: Oh, OK.

src/session.rs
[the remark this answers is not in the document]
ZuzaOsa: DataStax driver throws an error if partial credentials are provided, the check is performed in the `extend` function in `client-options.js` (using `validateAuthenticationOptions` function).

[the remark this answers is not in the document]
wprzytula: > DataStax driver throws an error if partial credentials are provided, the check is performed in the `extend` function in `client-options.js` (using `validateAuthenticationOptions` function).

Does this imply that with present code:
1. we would have already thrown an error if we were to call this code with partial credentials, or
2. we need to introduce the check?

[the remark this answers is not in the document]
ZuzaOsa: We would throw an error while executing the client constructor.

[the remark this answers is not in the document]
wprzytula: Then, let's throw `unreachable!()` upon error here instead of silent ignore.
```suggestion
    match (
        options.credentials_username.as_ref(),
        options.credentials_password.as_ref(),
    ) {
        (Some(username), Some(password)) => {
            builder = builder.user(username, password);
        }
        (None, None) => (),
        (Some(_), None) | (None, Some(_)) => {
            unreachable!("There is a check in JS Client constructor that should have prevented only one credential passed")
        }
    }
```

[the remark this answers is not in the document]
wprzytula: :thought_balloon: Even better: we could change credentials' representation in `ClientOptions` from (`Option<String>, Option<String>)` into `Option<(String, String)>`. Or couldn't we?

wprzytula (changes requested):

src/session.rs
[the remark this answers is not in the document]
ZuzaOsa: It is a tuple so passing it was not yet introduced in the napi version we are using.

lib/auth/plain-text-auth-provider.js
```diff
@@ -19,50 +18,46 @@ const Authenticator = provider.Authenticator;
  * @alias module:auth~PlainTextAuthProvider
  * @constructor
```
adespawn, line 19: classdesc and constructor tags no longer needed.

Move description of constructor to constructor

lib/client.js
```diff
@@ -90,6 +86,20 @@ class Client extends events.EventEmitter {
...
+     */
+    #setRustOptions(options) {
+        let rustOptions = rust.SessionOptions.empty();
+        rustOptions.connectPoints = options.contactPoints;
+        rustOptions.applicationName = options.applicationName;
+        rustOptions.applicationVersion = options.applicationVersion;
+        rustOptions.keyspace = options.keyspace;
+        return rustOptions;
+    }
+
```
adespawn: If we have ``client-options.js``, then maybe it would be a better place for this function?

adespawn (changes requested):

lib/client.js
ZuzaOsa: Then we would export a private function, I don't know if it is better.

adespawn: From client options file only default options are exported by the driver:
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/b3037b3ab0f6256b27ec1db7aa6a2d83b9befdcc/main.js#L14
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/b3037b3ab0f6256b27ec1db7aa6a2d83b9befdcc/main.js#L42-L44

adespawn (approved):
LGTM

wprzytula (approved):

PR #111, opened 2024-11-26
Upgrade rust driver

adespawn:
Update rust driver to version 0.15.0. 

Move out of legacy API, while still converting data to Row type

### DIFF
--- a/src/result.rs
+++ b/src/result.rs
@@ -5,7 +5,8 @@ use napi::{
 };
 use scylla::{
     frame::response::result::{ColumnType, CqlValue},
-    QueryResult,
+    transport::legacy_query_result::IntoLegacyQueryResultError,
+    LegacyQueryResult, QueryResult,
 };
 
 use crate::types::duration::DurationWrapper;
@@ -15,7 +16,7 @@ use crate::utils::js_error;
 
 #[napi]
 pub struct QueryResultWrapper {
-    internal: QueryResult,
+    internal: LegacyQueryResult,
 }
 
 #[napi]
@@ -69,8 +70,12 @@ pub enum CqlType {
 
 #[napi]
 impl QueryResultWrapper {
-    pub fn from_query(internal: QueryResult) -> QueryResultWrapper {
-        QueryResultWrapper { internal }
+    pub fn from_query(
+        internal: QueryResult,
+    ) -> Result<QueryResultWrapper, IntoLegacyQueryResultError> {
+        Ok(QueryResultWrapper {
+            internal: internal.into_legacy_result()?,
+        })
     }
 
     #[napi]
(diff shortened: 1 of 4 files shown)


src/result.rs
```diff
@@ -69,64 +73,83 @@ pub enum CqlType {
...
-        internal: QueryResult,
-    ) -> Result<QueryResultWrapper, IntoLegacyQueryResultError> {
-        Ok(QueryResultWrapper {
-            internal: internal.into_legacy_result()?,
-        })
+    pub fn from_query(internal: QueryResult) -> QueryResultWrapper {
+        let value = match internal.clone().into_rows_result() {
+            Ok(v) => QueryResultVariant::RowsResult(v),
+            Err(_) => QueryResultVariant::EmptyResult(internal),
+        };
```
wprzytula, line 80: This is not about the number of rows returned, but about the kind of the Result response: Rows or not Rows. For non-Rows Result kinds, this conversion fails.

src/result.rs
```diff
@@ -69,64 +73,83 @@ pub enum CqlType {
...
+            QueryResultVariant::RowsResult(v) => v,
+            QueryResultVariant::EmptyResult(_) => {
                 return None;
             }
         };
 
+        let rows = r2
+            .rows::<Row>()
+            .expect("Expected result to bo converted into Row")
+            .flatten();
```
wprzytula, line 96: This is absolutely unacceptable!!! We mustn't panic on bad data received!

src/result.rs
```diff
@@ -69,64 +73,83 @@ pub enum CqlType {
...
         Some(
-            rows.iter()
-                .map(|f| RowWrapper {
-                    internal: f.columns.clone(),
-                })
-                .collect(),
+            rows.map(|f| RowWrapper {
+                internal: f.columns.clone(),
```
wprzytula, line 99: Why clone the columns instead of moving them out?

wprzytula (changes requested):

src/result.rs
```diff
@@ -70,59 +78,79 @@ pub enum CqlType {
...
+        QueryResultWrapper { internal: value }
     }
 
     #[napi]
-    pub fn get_rows(&self) -> Option<Vec<RowWrapper>> {
-        let rows = match &self.internal.rows {
-            Some(r) => r,
-            None => {
-                return None;
+    pub fn get_rows(&self) -> napi::Result<Option<Vec<RowWrapper>>> {
```
wprzytula, line 89: So you changed this function's signature and didn't update any calling code. Is this function currently unused?

adespawn: It's used from nodeJS

wprzytula (approved):

PR #113, opened 2024-11-26
LocalDate type

Stapox35:


src/types/local_date.rs
```diff
@@ -0,0 +1,268 @@
...
+            total_days += day as i32 - 1;
+
+            for y in 1970..year {
+                total_days += if Self::is_leap_year(y) { 366 } else { 365 };
+            }
+        } else {
+            for m in (month + 1)..=12 {
+                total_days -= Self::days_in_month(m, year) as i32;
+            }
+            total_days -= Self::days_in_month(month, year) as i32 - day as i32 + 1;
```
adespawn, line 55:
```suggestion
            total_days -= (Self::days_in_month(month, year) - day) as i32 + 1;
```

src/types/local_date.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        if year >= 1970 {
+            for m in 1..month {
+                total_days += Self::days_in_month(m, year) as i32;
+            }
+            total_days += day as i32 - 1;
+
+            for y in 1970..year {
+                total_days += if Self::is_leap_year(y) { 366 } else { 365 };
+            }
```
adespawn, line 50: It's not the biggest issue, but this is soo inefficient...

src/types/local_date.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    /// Counts the number of days since 01.01.1970.
+    fn days_since_epoch(day: i8, month: i8, year: i32) -> i32 {
+        let mut total_days = 0;
+
+        if year >= 1970 {
+            for m in 1..month {
+                total_days += Self::days_in_month(m, year) as i32;
+            }
```
adespawn, line 45: Again: performance. Here we can maybe preprocess the values for each month accounting the leap year.

src/types/local_date.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub fn new_day(value: i32) -> napi::Result<Self> {
+        let mut x = LocalDateWrapper::new(1, 1, 1970)?;
+        x.value = value;
+        x.in_date = (MIN_JS_DATE..=MAX_JS_DATE).contains(&value);
+        if x.in_date {
+            x.add_day(value.into());
+        }
+        Ok(x)
+    }
+
```
adespawn, line 37: I would prefer to see this as creating value from days since epoch, rather then modifying value created in other ways

src/types/local_date.rs
```diff
@@ -0,0 +1,268 @@
...
+            for y in (year + 1)..1970 {
+                total_days -= if Self::is_leap_year(y) { 366 } else { 365 };
+            }
+        }
+
+        total_days
+    }
+
+    /// Adds the number of days to self.
+    fn add_day(&mut self, mut n: i64) {
```
adespawn, line 66: so this could create LocalDateWrapper object instead of modifying it

src/types/local_date.rs
```diff
@@ -0,0 +1,268 @@
...
+                n -= days_in_year;
+                self.year += 1;
+            } else if n <= -days_in_year {
+                n += days_in_year;
+                self.year -= 1;
+            } else {
+                break;
+            }
+        }
+
```
adespawn, line 83: My heart cries when I see this part done linearly 🥲

src/types/local_date.rs
```diff
@@ -0,0 +1,268 @@
...
+        while n != 0 {
+            let days_in_current_month = Self::days_in_month(self.month, self.year) as i64;
+
+            if n >= days_in_current_month {
+                n -= days_in_current_month;
+                self.month += 1;
+                if self.month > 12 {
+                    self.month = 1;
+                    self.year += 1;
+                }
```
adespawn, line 93: is this reachable?

src/types/local_date.rs
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        while self.day > Self::days_in_month(self.month, self.year) {
+            self.day -= Self::days_in_month(self.month, self.year);
+            self.month += 1;
+            if self.month > 12 {
+                self.month = 1;
+                self.year += 1;
+            }
+        }
```
adespawn, line 123: Again is this reachable at all?

adespawn: If no, than maybe add some kind of assert(false)

src/types/local_date.rs
```diff
@@ -0,0 +1,268 @@
...
+
+impl fmt::Display for LocalDateWrapper {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.in_date {
+            write!(
+                f,
+                "{}-{}-{}",
+                if self.year >= 1000 || self.year <= -1000 {
+                    format!("{}", self.year)
+                } else if self.year < 0 {
```
adespawn, line 238: Is this first branch nesesery? Other two should work the same here

src/types/local_date.rs
```diff
@@ -0,0 +1,268 @@
...
+                    let year_string = (-(self.year)).to_string();
```
adespawn, line 239:
```suggestion
                    let year_string = (self.year.unsigned_abs()).to_string();
```
With the current logic we shouldn't reach the values where this is a problem, but is better to use safer way of doing it 😛. ~~We don't want to have someone open PR to this repo complaining about invalid handling of negative values~~

adespawn (changes requested):

lib/types/local-date.js
```diff
@@ -1,264 +1,270 @@
...
+        if (isNaN(date.getTime())) {
+            throw new TypeError(`Invalid date: ${date}`);
         }
-        const parts = value.split("-");
         return new LocalDate(
-            multiplier * parseInt(parts[0], 10),
-            parseInt(parts[1], 10),
-            parseInt(parts[2], 10),
+            date.getFullYear(),
+            date.getMonth() + 1,
```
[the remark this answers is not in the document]
Stapox35: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth 
`getMonth()` return int between 0 and 11, but `LocalDate` month is between 1 and 12.

lib/types/local-date.js
```diff
@@ -1,32 +1,42 @@
...
  */
-const millisecondsPerDay = 86400000;
+const dateCenter = Math.pow(2, 31);
+
 /**
  * @private
+ * @const
  */
-const dateCenter = Math.pow(2, 31);
+const millisecondsPerDay = 86400000;
```
PiotrJunior, line 17: If it is private use #varName

lib/types/local-date.js
```diff
@@ -1,32 +1,42 @@
...
+const rust = require("../../index");
 /** @module types */
 
 /**
+ * 2^31 days before unix epoch is -5877641-06-23. This is the first day that can be represented by this class.
  * @private
  * @const
  */
-const millisecondsPerDay = 86400000;
+const dateCenter = Math.pow(2, 31);
```
PiotrJunior, line 11: If it is private use #varName

lib/types/local-date.js
```diff
@@ -40,61 +50,106 @@ class LocalDate {
             typeof month === "number" &&
             typeof day === "number"
         ) {
-            // Use setUTCFullYear as if there is a 2 digit year, Date.UTC() assumes
-            // that is the 20th century.
-            this.date = new Date();
-            this.date.setUTCHours(0, 0, 0, 0);
-            this.date.setUTCFullYear(year, month - 1, day);
-            if (isNaN(this.date.getTime())) {
+            // it will throw an error if the values are wrong. TODO: add tests for this.
```
PiotrJunior, line 53: 1. Sentences start with uppercase
2. TODO - add those tests or remove this TODO

lib/types/local-date.js
```diff
@@ -40,61 +50,106 @@ class LocalDate {
...
+            this.#internal = rust.LocalDateWrapper.new(day, month, year);
+            if (!this.#internal.inDate) {
+                throw new Error("You must provide a valid year, month and day");
+            }
+        } else if (
+            typeof month === "undefined" &&
+            typeof day === "undefined" &&
+            typeof year === "number"
+        ) {
+            // in days since epoch.
```
PiotrJunior, line 63: Upper case

lib/types/local-date.js
```diff
@@ -1,264 +1,270 @@
...
      */
-    this.month = this.date.getUTCMonth() + 1;
+    get _value() {
+        return this.#internal.inDate ? null : this.#internal.value;
+    }
+
+    set _value(_) {
+        throw new SyntaxError("LocalDate _value is read-only");
+    }
+
```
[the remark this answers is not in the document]
PiotrJunior: if it is private and not mentioned in docs it should be marked as `@private` and `#` should be used

lib/types/local-date.js
```diff
@@ -172,20 +210,12 @@ class LocalDate {
      * if the given one is greater.
```
PiotrJunior, line 210: Uppercase

lib/types/local-date.js
```diff
@@ -172,20 +210,12 @@ class LocalDate {
...
-            return -1;
-        }
-        if (diff > 0) {
+        if (this.valueSinceEpoch == other.valueSinceEpoch) {
+            return 0;
+        } else if (this.valueSinceEpoch > other.valueSinceEpoch) {
             return 1;
         }
-        return 0;
+        return -1;
```
PiotrJunior, line 218: Why did you droped using `isNan`. Didn't it change output?

PiotrJunior (changes requested):

lib/types/local-date.js
Stapox35: Why uppercase in the middle of a sentence? 

> 0 if they are the same, 1 if the this is greater, and -1 **if** the given one is greater.

lib/types/local-date.js
Stapox35: This used `isNaN` because `_value` returned `NaN` in some cases. The new member returns value always. This does not change output of function.

src/types/local_date.rs
```diff
@@ -0,0 +1,268 @@
...
+            1 | 3 | 5 | 7 | 8 | 10 | 12 => 31,
+            4 | 6 | 9 | 11 => 30,
+            2 => {
+                if Self::is_leap_year(year) {
+                    29
+                } else {
+                    28
+                }
+            }
+            _ => 0,
```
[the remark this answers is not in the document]
Stapox35: Now this function is only used in the condition:
`if !(day >= 1 && day <= Self::days_in_month(month, year)) {`
that is after the conditional 
`if !(1..=12).contains(&month)`
So it will never be called with a non-existent month.

lib/types/local-date.js
Stapox35: My mistake. A variable that is not in a class cannot be private....

src/types/local_date.rs
```diff
@@ -0,0 +1,271 @@
...
+
+            LocalDateWrapper {
+                value,
+                year: Some(date.2),
+                month: Some(date.1),
+                day: Some(date.0),
+                in_date: true,
+            }
+        }
+    }
```
adespawn, line 232: It looks almost the same as ``new_day``. Maybe one should use the other

src/types/local_date.rs
```diff
@@ -0,0 +1,271 @@
...
+            Ok(LocalDateWrapper {
+                value,
+                year: None,
+                month: None,
+                day: None,
+                in_date: false,
+            })
+        } else {
+            let date = Self::date_from_days(value.into());
+            Ok(LocalDateWrapper::new(date.0, date.1, date.2)?)
```
adespawn, line 43: Why do use ``new`` instead of creating the object directly?

src/types/local_date.rs
```diff
@@ -0,0 +1,271 @@
...
+        let mut total_days = 0;
+
+        let number_day = DAY_IN_MONTH[month as usize - 1] // number of days from 1 January
+            + if Self::is_leap_year(year) && month > 2 {
+                1
+            } else {
+                0
+            }
+            + day as i32
+            - 1;
```
adespawn, line 58: ❓ We offset the value two times by a constant. Maybe there is a cleaner way to write it?

src/types/local_date.rs
```diff
@@ -0,0 +1,271 @@
...
+                    28
+                }
+            }
+            _ => 0,
+        }
+    }
+
+    fn number_leap_years(n: i32) -> i32 {
+        n / 4 - n / 100 + n / 400
+    }
```
adespawn, line 149: In each case we need a number of leap years between some range. Maybe it would be cleaner to have a function that does that rather than calling it twice?

src/types/local_date.rs
```diff
@@ -0,0 +1,271 @@
...
+                n += (k.abs() * 365 + Self::number_leap_years(year - k)
+                    - Self::number_leap_years(year + 1)) as i64;
+            }
+        }
+
+        if n < 0 {
+            // If the remaining number of days is negative, change the year and count the complement.
+            year -= 1;
+            n += if Self::is_leap_year(year) { 366 } else { 365 };
+        }
```
adespawn, line 102: Can you guarantee we will always be at most one year behind?

adespawn: It seems like this may not be the case for dates ~1500 years in the future

adespawn (changes requested):
There is also missing logic for passing LocalDate back to rust

src/types/local_date.rs
Stapox35: If there were `n > 365` then it would not come out of the loop above with the condition: 
`while n.abs() >= 365`

src/types/local_date.rs
```diff
@@ -0,0 +1,271 @@
...
+const LEAP_YEAR_1970: i32 = 477;
+// Number of days to the beginning of each month.
+const DAY_IN_MONTH: [i32; 12] = [0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334];
+#[napi]
+pub struct LocalDateWrapper {
+    pub value: i32, // wrapper for number of days from 01.01.1970
+    pub year: Option<i32>,
+    pub month: Option<i8>,
+    pub day: Option<i8>,
+    pub in_date: bool, // value can be represented as Date class in JS
```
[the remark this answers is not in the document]
Stapox35: I add a comment: 
> LocalDateWrapper holds two data representations - value and day, month, year. The class in JS has getters for both representations. When a new LocalDateWrapper instance is created, the second representation is calculated.`

src/types/local_date.rs
Stapox35: in `days_since_epoch` this function is called once.

lib/types/local-date.js
```diff
@@ -1,264 +1,296 @@
...
+        throw new SyntaxError("LocalDate day is read-only");
+    }
+
     /**
-     * A number between 1 and 12 inclusive representing the month.  May return
-     * NaN if cannot be represented as a Date.
-     * @type Number
+     * If date cannot be represented yet given a valid days since epoch, track it internally.
+     * @readonly
+     * @deprecated The new member value returns always, regardless of the date.
```
adespawn, line 129:
```suggestion
     * @deprecated This member is in Datastax documentation, but it seems to not be exposed in the API. Additionally we added a new class member: ``value`` that always returns ???, regardless of the date.
```

adespawn (dismissed):

wprzytula (changes requested):

lib/types/local-date.js
```diff
@@ -259,6 +306,24 @@ class LocalDate {
     toJSON() {
         return this.toString();
     }
+
+    /**
+     * Get LocalDate from rust object. Not intended to be exposed in the API
```
PiotrJunior, line 279: > Not intended to be exposed in the API

That's what '@package' is for.

PiotrJunior (commented):
Why you didn't use something like `chrono` crate for date type handling in Rust.

src/types/local_date.rs
```diff
@@ -0,0 +1,228 @@
...
+                    + Self::number_leap_years(year.abs() + 1)
+                    + LEAP_YEAR_1970
+                    + 1;
+            } else {
+                total_days -=
+                    (1970 - year - 1) * 365 + LEAP_YEAR_1970 - Self::number_leap_years(year + 1);
+            }
+        }
+        total_days
+    }
```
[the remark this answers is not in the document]
Stapox35: Does it look that bad? This is my own code...

src/types/local_date.rs
```diff
@@ -0,0 +1,228 @@
...
+const DAY_IN_MONTH: [i32; 12] = [0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334];
+// LocalDateWrapper holds two data representations - value and day, month, year.
+// The class in JS has getters for both representations.
+// When a new LocalDateWrapper instance is created, the second representation is calculated.
+#[napi]
+pub struct LocalDateWrapper {
+    pub value: i32, // wrapper for number of days from 01.01.1970
+    pub year: Option<i32>,
+    pub month: Option<i8>,
+    pub day: Option<i8>,
```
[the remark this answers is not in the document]
Stapox35: > 🔧 If these fields are either all Some or all None, then they should be a single `Option`.

Unfortunately NAPI does not support conversion for Option with tuple - `Option<(i32, i8, i8)>`.

Stapox35:
> Why you didn't use something like `chrono` crate for date type handling in Rust.

Because `chrono::NaiveDate` does not cover the entire range supported by the DataStax driver and ScyllaDB. I mentioned this at the meeting.
`NaiveDate`:  [Source](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDate.html)

> from Jan 1, 262145 BCE to Dec 31, 262143 CE

`DataStax`: [Source](https://docs.datastax.com/en/developer/nodejs-driver/4.7/api/module.types/class.LocalDate/index.html)

> range [-5877641-06-23; 5881580-07-17]

`ScyllaDB`: [Source](https://rust-driver.docs.scylladb.com/stable/data-types/date.html)

> number of days since -5877641-06-23

I have added information about this to the PR description.

src/types/local_date.rs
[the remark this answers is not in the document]
wprzytula: Does napi support conversion for `Option<MyTuple>` with `MyTuple` defined as:
```rust
struct MyTuple(Option<T>, Option<T>, Option<T>)
```
or
```rust
struct MyTuple {
  f: Option<T>,
  s: Option<T>,
  t: Option<T>,
}
```
?

If so, then this is the way to do this.

src/types/local_date.rs
```diff
@@ -0,0 +1,290 @@
...
+            _ => Err(js_error("Invalid format of string")),
+        }
+    }
+}
+
+impl fmt::Display for LocalDateWrapper {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.in_date {
+            if self.year.unwrap() < 0 {
+                let year_string = (self.year.unwrap().unsigned_abs()).to_string();
```
wprzytula, line 156: :wrench: These `unwrap`s are unacceptable. There should be one `Option` that you pattern match on, and if it's `Some`, then you immediately get access to its contents without `unwrap`ping.

src/types/local_date.rs
```diff
@@ -0,0 +1,290 @@
...
+            d if d < 2 => match value.parse::<i32>() {
+                Ok(val) => Ok(val),
+                Err(_) => Err(js_error("Invalid format of string")),
+            },
+            2 | 3 => {
+                let lambda = |s: String| -> Result<(i32, i8, i8), ParseIntError> {
+                    let mut cut: usize = 0;
+                    if s.starts_with('-') {
+                        cut = 1;
+                    }
```
wprzytula, line 121: :recycle: Please avoid needless `mut` in variables. This can be easily made a non-`mut` `let`.

src/types/local_date.rs
```diff
@@ -0,0 +1,290 @@
...
+                    let mut parts = s.get(cut..).unwrap().split('-');
+                    let year = parts.next().unwrap().parse::<i32>()?;
+                    Ok((
+                        match cut {
+                            0 => year,
+                            _ => -year,
+                        },
+                        parts.next().unwrap().parse::<i8>()?,
+                        parts.next().unwrap().parse::<i8>()?,
+                    ))
```
wprzytula, line 132: :fearful: No-no. It is possible to panic here if someone passes a string containing a date in an invalid format.

src/types/local_date.rs
```diff
@@ -0,0 +1,290 @@
...
+            if self.year.unwrap() < 0 {
+                let year_string = (self.year.unwrap().unsigned_abs()).to_string();
+                write!(
+                    f,
+                    "-{}{}",
+                    "0".repeat(max(4 - year_string.len() as i8, 0) as usize),
+                    year_string
+                )?;
+            } else {
+                let year_string = self.year.unwrap().to_string();
```
wprzytula, line 164: :wrench: This is an allocation in `Display` - it's unnacceptable again.

wprzytula: Instead, you should pass `year` to `write!` directly, because then its `Display` impl will be employed without allocations.

src/types/local_date.rs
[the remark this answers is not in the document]
wprzytula: Actually, I can see that you introduced `Ymd` struct with exactly these characteristics. Please replace those `Option`s with `Option<Ymd>`.

src/types/local_date.rs
```diff
@@ -0,0 +1,290 @@
...
+                .unwrap();
+
+            Some(Ymd {
+                year,
+                month: q as i8,
+                day: (n - DAY_IN_MONTH[q - 1] as i64) as i8 + 1,
+            })
+        }
+    }
+}
```
wprzytula, line 290: As this is a very nontrivial logic, please add a suite of unit tests for both functions, preferably in this file (in a dedicated `mod tests`).

wprzytula (changes requested):

src/types/local_date.rs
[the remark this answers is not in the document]
wprzytula: It's complicated and thus it deserves an exhaustive suite of unit tests.

lib/types/local-date.js
```diff
@@ -1,264 +1,297 @@
...
- * @private
+ * 2^31 days before unix epoch is -5877641-06-23. This is the first day that can be represented by this class.
  * @const
  */
-const millisecondsPerDay = 86400000;
+const dateCenter = Math.pow(2, 31);
+
 /**
- * @private
+ * Maximum number of days from 01.01.1970 served by the class.
```
adespawn, line 13: served or supported?

Not that it make significant difference

lib/types/local-date.js
```diff
@@ -1,264 +1,297 @@
...
+ * @const
  */
-const dateCenter = Math.pow(2, 31);
+const maxDay = 2147483647;
```
adespawn, line 16: Maybe a note this is i32::MAX?

lib/types/local-date.js
```diff
@@ -1,264 +1,297 @@
...
+
+/**
+ * Minimum number of days from 01.01.1970 served by the class.
+ * @const
+ */
+const minDay = -2147483648;
```
adespawn, line 22: Maybe a note this is i32::MIN?

lib/types/local-date.js
```diff
@@ -1,264 +1,297 @@
...
+        ) {
+            // It will throw an error if the values are wrong.
+            this.#internal = rust.LocalDateWrapper.new(day, month, year);
+            if (!this.#internal.inDate) {
+                throw new Error("You must provide a valid year, month and day");
+            }
+        } else if (
+            typeof month === "undefined" &&
+            typeof day === "undefined" &&
+            typeof year === "number"
```
adespawn, line 71: Nit: different order of argument checking

lib/types/local-date.js
```diff
@@ -1,264 +1,297 @@
...
+        ) {
+            // In days since epoch.
+            if (year < minDay || year > maxDay) {
                 throw new Error(
                     "You must provide a valid value for days since epoch (-2147483648 <= value <= 2147483647).",
```
adespawn, line 76: Use variables, rather than constants written by hand

src/types/local_date.rs
```diff
@@ -0,0 +1,528 @@
...
+
+// Max and min date range of the Date class in JS.
+const MAX_JS_DATE: i32 = 100_000_000;
+const MIN_JS_DATE: i32 = -MAX_JS_DATE;
+
+// Number of leap years up to 1970.
+const LEAP_YEAR_1970: i32 = 477;
+// Number of days to the beginning of each month.
+const DAY_IN_MONTH: [i32; 12] = [0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334];
+// LocalDateWrapper holds two data representations - value and day, month, year.
```
adespawn, line 17: Nit: Add line sepration

src/types/local_date.rs
```diff
@@ -0,0 +1,528 @@
...
+// The class in JS has getters for both representations.
+// When a new LocalDateWrapper instance is created, the second representation is calculated.
+#[napi]
```
adespawn, line 20: Did we decide if we want to have comment or ``#[napi]`` first?

src/types/local_date.rs
```diff
@@ -0,0 +1,528 @@
...
+const MAX_JS_DATE: i32 = 100_000_000;
+const MIN_JS_DATE: i32 = -MAX_JS_DATE;
+
+// Number of leap years up to 1970.
+const LEAP_YEAR_1970: i32 = 477;
+// Number of days to the beginning of each month.
+const DAY_IN_MONTH: [i32; 12] = [0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334];
+// LocalDateWrapper holds two data representations - value and day, month, year.
+// The class in JS has getters for both representations.
+// When a new LocalDateWrapper instance is created, the second representation is calculated.
```
adespawn, line 19: Maybe this would be worth converting to docs format?

src/types/local_date.rs
```diff
@@ -0,0 +1,528 @@
...
+                let pattern = r"^-?\d{1}\d*-(0?[1-9]|1[012])-(0?[1-9]|[12][0-9]|3[01])$";
+                let re = Regex::new(pattern).unwrap();
+                if !re.is_match(&value) {
+                    return Err(DateError::InvalidFormat.into());
+                }
+
+                let lambda = |s: String| -> Result<(i32, i8, i8), DateError> {
+                    let date = s.strip_prefix('-').unwrap_or(&s);
+                    // From checking the regex and from removing the first '-',
+                    // it is clear that the date string has three '-'.
```
adespawn, line 92: Move this comment one line up

adespawn (changes requested):
Ensure all integration tests, that need local date are enabled

src/types/local_date.rs
Stapox35: @ZuzaOsa @wprzytula ?

wprzytula: The usual order in Rust is to have attributes (#[...]) as close to the decorated item as possible; therefore, comments and doc comments come above them.

src/types/local_date.rs
```diff
@@ -0,0 +1,223 @@
...
+    n / 4 - n / 100 + n / 400
+}
+
+#[napi(object)]
+#[derive(Clone, PartialEq, Debug)]
+pub struct Ymd {
+    pub year: i32,
+    pub month: i8,
+    pub day: i8,
+}
```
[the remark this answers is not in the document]
adespawn: Almost yes. With some minor changes, this approach may work. I'll try to add soon.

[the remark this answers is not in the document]
Stapox35: > ❓ What does `#[napi(object)]` do here?

In `LocalDateWrapper` is member `pub date: Option<Ymd>`. Ymd must have `#[napi(object)]`, otherwise NAPI will not return this object as a member in LocalDateWrapper

adespawn (approved):

wprzytula (approved):

PR #118, opened 2024-12-07
Add error types

ZuzaOsa:
napi-rs does not support throwing specific errors, either custom or node.js specific. This PR adds support for it by adding a wrapper function that catches the error and rethrows a correct one based on the message. Tests were added for this functionality. Some errors need special parameters passed when creating, they are not supported for now (#119 ) as I don't think it is a crucial feature for Iteration 1.

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -9,163 +9,158 @@ const util = require("util");
  * Base Error
  * @private
  */
-function DriverError(message) {
-    Error.call(this, message);
-    Error.captureStackTrace(this, this.constructor);
-    this.name = this.constructor.name;
-    this.info = "Cassandra Driver Error";
-    // Explicitly set the message property as the Error.call() doesn't set the property on v8
-    this.message = message;
+class DriverError extends Error {
+    /**
+     * @param {string} [message]
+     */
+    constructor(message) {
+        super(message);
+        Error.captureStackTrace(this, this.constructor);
+        this.name = this.constructor.name;
+        this.info = "Cassandra Driver Error";
+        // Explicitly set the message property as the Error.call() doesn't set the property on v8.
+        this.message = message;
+    }
 }
-
-util.inherits(DriverError, Error);
 
 /**
  * Represents an error when a query cannot be performed because no host is available or could be reached by the driver.
- * @param {Object} innerErrors An object map containing the error per host tried
- * @param {String} [message]
- * @constructor
  */
-function NoHostAvailableError(innerErrors, message) {
-    DriverError.call(this, message);
-    this.innerErrors = innerErrors;
-    this.info =
-        "Represents an error when a query cannot be performed because no host is available or could be reached by the driver.";
-    if (!message) {
-        this.message = "All host(s) tried for query failed.";
-        if (innerErrors) {
-            const hostList = Object.keys(innerErrors);
-            if (hostList.length > 0) {
-                const host = hostList[0];
-                this.message += util.format(
-                    " First host tried, %s: %s. See innerErrors.",
-                    host,
-                    innerErrors[host],
-                );
+class NoHostAvailableError extends DriverError {
+    /**
+     * @param {Object} innerErrors An object map containing the error per host tried.
+     * @param {string} [message]
+     */
+    constructor(innerErrors, message) {
+        super(message);
+        this.innerErrors = innerErrors;
+        this.info =
+            "Represents an error when a query cannot be performed because no host is available or could be reached by the driver.";
+        if (!message) {
+            this.message = "All host(s) tried for query failed.";
+            if (innerErrors) {
+                const hostList = Object.keys(innerErrors);
+                if (hostList.length > 0) {
+                    const host = hostList[0];
+                    this.message += ` First host tried, ${host}: ${innerErrors[host]}. See innerErrors.`;
+                }
             }
         }
     }
 }
 
-util.inherits(NoHostAvailableError, DriverError);
-
 /**
- * Represents an error message from the server
- * @param {Number} code Cassandra exception code
- * @param {String} message
- * @constructor
+ * Represents an error message from the server.
  */
-function ResponseError(code, message) {
-    DriverError.call(this, message);
+class ResponseError extends DriverError {
     /**
-     * The error code as defined in [responseErrorCodes]{@link module:types~responseErrorCodes}.
-     * @type {Number}
+     * @param {number} code Cassandra exception code as defined in [responseErrorCodes]{@link module:types~responseErrorCodes}.
+     * @param {string} message
      */
-    this.code = code;
-    this.info = "Represents an error message from the server";
+    constructor(code, message) {
+        super(message);
+        this.code = code;
+        this.info = "Represents an error message from the server.";
+    }
 }
-
-util.inherits(ResponseError, DriverError);
 
 /**
  * Represents a bug inside the driver or in a Cassandra host.
- * @param {String} message
- * @constructor
  */
-function DriverInternalError(message) {
-    DriverError.call(this, message);
-    this.info = "Represents a bug inside the driver or in a Cassandra host.";
-}
-
-util.inherits(DriverInternalError, DriverError);
-
-/**
- * Represents an error when trying to authenticate with auth-enabled host
- * @param {String} message
- * @constructor
- */
-function AuthenticationError(message) {
-    DriverError.call(this, message);
-    this.info =
-        "Represents an authentication error from the driver or from a Cassandra node.";
-}
-
-util.inherits(AuthenticationError, DriverError);
-
-/**
- * Represents an error that is raised when one of the arguments provided to a method is not valid
- * @param {String} message
- * @constructor
- */
-function ArgumentError(message) {
-    DriverError.call(this, message);
-    this.info =
-        "Represents an error that is raised when one of the arguments provided to a method is not valid.";
-}
-
-util.inherits(ArgumentError, DriverError);
-
-/**
- * Represents a client-side error that is raised when the client didn't hear back from the server within
- * {@link ClientOptions.socketOptions.readTimeout}.
- * @param {String} message The error message.
- * @param {String} [host] Address of the server host that caused the operation to time out.
- * @constructor
- */
-function OperationTimedOutError(message, host) {
-    DriverError.call(this, message, this.constructor);
-    this.info =
-        "Represents a client-side error that is raised when the client did not hear back from the server " +
-        "within socketOptions.readTimeout";
-
+class DriverInternalError extends DriverError {
     /**
-     * When defined, it gets the address of the host that caused the operation to time out.
-     * @type {String|undefined}
+     * @param {string} message
      */
-    this.host = host;
+    constructor(message) {
+        super(message);
+        this.info =
+            "Represents a bug inside the driver or in a Cassandra host.";
+    }
 }
 
-util.inherits(OperationTimedOutError, DriverError);
+/**
+ * Represents an error when trying to authenticate with auth-enabled host.
+ */
+class AuthenticationError extends DriverError {
+    /**
+     * @param {string} message
+     */
+    constructor(message) {
+        super(message);
+        this.info =
+            "Represents an authentication error from the driver or from a Cassandra node.";
+    }
+}
+
+/**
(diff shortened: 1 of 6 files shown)


lib/errors.js
```diff
@@ -9,164 +9,172 @@ const util = require("util");
...
+            this.message = "All host(s) tried for query failed.";
+            if (innerErrors) {
+                const hostList = Object.keys(innerErrors);
+                if (hostList.length > 0) {
+                    const host = hostList[0];
+                    this.message += util.format(
+                        " First host tried, %s: %s. See innerErrors.",
+                        host,
+                        innerErrors[host],
+                    );
```
adespawn, line 49: Why is this kind of string formatting present in datastax code...

lib/errors.js
```diff
@@ -9,164 +9,172 @@ const util = require("util");
...
+     * @param {number} maxRequestsPerConnection
+     * @param {number} connectionLength
+     */
+    constructor(address, maxRequestsPerConnection, connectionLength) {
+        const message = util.format(
+            "All connections to host %s are busy, %d requests are in-flight on %s",
+            address,
+            maxRequestsPerConnection,
+            connectionLength === 1 ? "a single connection" : "each connection",
+        );
```
adespawn, line 170: Again, what is this formatting...

lib/new-utils.js
```diff
@@ -8,4 +9,39 @@ function throwNotSupported(name) {
...
+            return fn.apply(this, args);
+        } catch (error) {
+            // Check if message is of format errorType#errorMessage, if so throw a correct error,
+            // otherwise throw the original error.
+            const [errorType, ...messageParts] = error.message.split("#");
+            const message = messageParts.join("#");
+
+            if (errorTypeMap[errorType]) {
+                const errorClass = errorTypeMap[errorType];
+                throw new errorClass(message);
```
adespawn, line 39: Do we preserve stack trace of the original error?

test/unit/errors.js
```diff
@@ -0,0 +1,204 @@
...
+    });
+}
+describe("napiErrorHandler", function () {
+    const classInstance = new SomeJsClass();
+
+    it("should throw ArgumentError", function () {
+        assert.throws(
+            function () {
+                classInstance.throwError("ArgumentError");
+            },
```
adespawn, line 35: nit: 
```suggestion
            () => classInstance.throwError("ArgumentError"),
```

src/utils.rs
```diff
@@ -12,6 +29,27 @@ pub(crate) fn js_error<T: Display>(e: T) -> napi::Error {
     napi::Error::new(Status::GenericFailure, e.to_string())
 }
```
adespawn, line 30: Maybe let's use ``js_typed_error`` in ``js_error``?

adespawn (changes requested):

lib/errors.js
```diff
@@ -9,164 +9,163 @@ const util = require("util");
...
+        this.message = message;
+    }
 }
 
-util.inherits(DriverError, Error);
-
 /**
  * Represents an error when a query cannot be performed because no host is available or could be reached by the driver.
- * @param {Object} innerErrors An object map containing the error per host tried
- * @param {String} [message]
```
PiotrJunior, line 26: Why those params descriptions were removed? It also regards other constructors

adespawn:
There is separate folder for tests in rust part of the project structure. Fell free to use it :laughing:

src/utils.rs
```diff
@@ -2,14 +2,52 @@ use std::{error::Error, fmt::Display};
...
+    NotSupportedError,
+    OperationTimedOutError, // TODO: Add suport for fields of this error
+    ResponseError,          // TODO: Add suport for fields of this error
+    Error,
+    RangeError,
+    ReferenceError,
+    SyntaxError,
+    TypeError,
+}
+
```
PiotrJunior, line 21: Have you added an issue regarding those TODOs?

ZuzaOsa:
> There is separate folder for tests in rust part of the project structure. Fell free to use it :laughing: 

Uno reverse, result tests on ``types-support`` also aren't in there 🙃

adespawn:
> Uno reverse, result tests on `types-support` also aren't in there 🙃

Well, they should be there 🫣

src/utils.rs
ZuzaOsa: Yes, issue #119

lib/new-utils.js
```diff
@@ -8,4 +9,41 @@ function throwNotSupported(name) {
...
+ *
+ * @param {Function} fn - The original function to be wrapped.
+ * @returns {Function} A wrapped function with error handling logic.
+ */
+function napiErrorHandler(fn) {
+    return function (...args) {
+        try {
+            return fn.apply(this, args);
+        } catch (error) {
+            // Check if message is of format errorType#errorMessage, if so throw a correct error,
```
adespawn, line 32: > correct error

I know what tou want to say, but this sounds a bit off. Maybe something like "Map onto appropriate error"

lib/new-utils.js
```diff
@@ -8,4 +9,41 @@ function throwNotSupported(name) {
...
+            // otherwise throw the original error.
+            const [errorType, ...messageParts] = error.message.split("#");
+            const message = messageParts.join("#");
+
+            if (errorTypeMap[errorType]) {
+                const errorClass = errorTypeMap[errorType];
+                const newError = new errorClass(message);
+                newError.stack = error.stack;
```
adespawn, line 40: Do we need creating ``errorClass`` variable, or can we skip it and just do: ``new (errorTypeMap[errorType])(message)``?

lib/new-utils.js
```diff
@@ -8,4 +9,41 @@ function throwNotSupported(name) {
...
+    Error,
+    RangeError,
+    ReferenceError,
+    SyntaxError,
+    TypeError,
+};
+
+/**
+ * A wrapper function to map napi errors to Node.js errors or custom errors.
+ *
```
adespawn, line 23: Add an expected use case for this function:
Smh like: "This should be used always when calling napi functions that may throw errors"

As this function will be used in a lot of places, it's better to write a bit more about it. (Maybe also things like motivation for this approach?)

adespawn (changes requested):
Few small things, otherwise LGTM

lib/errors.js
```diff
@@ -9,164 +9,159 @@ const util = require("util");
...
+        this.info =
+            "Represents an error that is raised when one of the arguments provided to a method is not valid.";
+    }
 }
 
-util.inherits(ArgumentError, DriverError);
-
 /**
- * Represents a client-side error that is raised when the client didn't hear back from the server within
+ * Represents a client-side error that is raised when the client didn't hear back from the server within.
```
adespawn, line 110: Ah, those pesky dots :laughing:

adespawn (approved):

Stapox35 (approved):
Most test files (in the `test/unit/` folder) have `tests` at the end in their name (e.g. branch types-support). Maybe rename it?

test/unit/errors-tests.js
```diff
@@ -0,0 +1,166 @@
...
+    throwError = napiErrorHandler(function (errorType) {
+        return throwTestError(errorType);
+    });
+    throwMessage = napiErrorHandler(function (message) {
+        return throwTestError("Error", message);
+    });
+    throwSimple = napiErrorHandler(function () {
+        throw new Error("Simple error");
+    });
+}
```
wprzytula, line 27: :question: What's the purpose of each of these?

lib/new-utils.js
```diff
@@ -8,4 +9,44 @@ function throwNotSupported(name) {
...
+ * to catch the original error and throw a new one with the appropriate type.
+ * This should be used to wrap all NAPI-RS functions that may throw errors.
+ *
+ * @param {Function} fn The original function to be wrapped.
+ * @returns {Function} A wrapped function with error handling logic.
+ */
+function napiErrorHandler(fn) {
+    return function (...args) {
+        try {
+            return fn.apply(this, args);
```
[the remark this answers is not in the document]
ZuzaOsa: It applies ``fn`` to arguments and specifies what should be referenced when a function references ``this`` in its body.

src/tests/error_throwing_tests.rs
```diff
@@ -0,0 +1,23 @@
+use crate::utils::{js_typed_error, ErrorType};
+
+#[napi]
+/// Test function that throws specified error with a custom message.
+pub fn throw_test_error(error_type: String, custom_message: Option<String>) -> napi::Result<()> {
```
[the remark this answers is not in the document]
ZuzaOsa: Those values are passed through napi so they cannot be of ``&str`` or generic type.

lib/new-utils.js
```diff
@@ -8,4 +9,44 @@ function throwNotSupported(name) {
...
+        } catch (error) {
+            // Check if message is of format errorType#errorMessage, if so map it to
+            // appropriate error, otherwise throw the original error.
+            const [errorType, ...messageParts] = error.message.split("#");
+            const message = messageParts.join("#");
```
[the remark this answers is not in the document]
ZuzaOsa: ad 2) The errors we catch will be only from Rust driver or other libraries used by the driver so hash should be reliable enough. There is no issue if there are more ``#`` chars in the message. If the part of the message before the first ``#`` (or the whole message if there is no ``#``) does not match any error string the same error is rethrown. Both of those cases are tested in unit tests.

test/unit/errors-tests.js
ZuzaOsa: I added comments to explain each function. They are here to test if the wrapper handles correctly different types of error, errors with a message and errors thrown from JS instead of napi.

src/tests/error_throwing_tests.rs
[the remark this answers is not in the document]
wprzytula: 1. Hmm, but IIRC we can accept `&T` in Rust functions called from JS through napi. At least I saw `Vec<&T>`, so why `&T` would not work?
2. If `&str` is problematic, then maybe `&String`? Just to avoid allocations.

wprzytula (approved):

src/tests/error_throwing_tests.rs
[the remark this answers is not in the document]
adespawn: 1. I don't remember any place in code where we accept generic types (or vector of generic types). From my testing any attempt to add generic to napi exposed functions results in CE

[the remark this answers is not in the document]
wprzytula: > 1. I don't remember any place in code where we accept generic types (or vector of generic types). From my testing any attempt to add generic to napi exposed functions results in CE

That's correct, what I meant by `T` was some specific type `T`, not a generic parameter. Here, I'd just like to pass `&str` or `&String`.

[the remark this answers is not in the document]
ZuzaOsa: Passing a reference instead of a value would mean we have a reference to relevant string in JS and there is not enough guarantees of its lifetime. From what i read there is also an issue that Rust and JS have different ways of encoding strings (UTF-8 in Rust and UTF-16 in JS).

adespawn:
So it means this just doesn’t compile?

src/tests/error_throwing_tests.rs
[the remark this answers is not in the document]
adespawn: So it means this just doesn’t compile?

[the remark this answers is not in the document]
ZuzaOsa: It throws a compilation error because ``&String`` and ``&str`` do not implement FromNapiRef trait.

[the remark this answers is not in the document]
ZuzaOsa: We could do ``Vec<&T>`` in other places because that specific ``T`` implemented this trait.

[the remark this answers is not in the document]
wprzytula: > Passing a reference instead of a value would mean we have a reference to relevant string in JS and there is not enough guarantees of its lifetime.

I don't think that lifetime is an issue here.

> From what I read there is also an issue that Rust and JS have different ways of encoding strings (UTF-8 in Rust and UTF-16 in JS).

This explains everything... Gosh, UTF-16 is so useless... (but still Java, JS, and Windows keep using it for some reason...)

PR #123, opened 2024-12-09
Mark integer as deprecated

adespawn:
As already discussed, I deprecated the Integer class in favor of builtin BigInt type.

Support for CQL Varint which is using this type (and BigDecimal) will be added as a separate PR.


PiotrJunior (approved):
:)

PR #125, opened 2024-12-10
Add support for parsing InetAdress into QueryParameterWrapper

PiotrJunior:


Stapox35:
You have a mistake in the title.

lib/types/inet-address.js
```diff
@@ -283,6 +283,15 @@ class InetAddress {
     static fromRust(rustInetAddress) {
         return new InetAddress(rustInetAddress.getBuffer());
     }
+
+    /**
+     * Get the rust object. Not intended to be exposed in the API
```
adespawn, line 288: > Not intended to be exposed in the API

That what the ``@package`` is for

PR #132, opened 2025-01-08
Deprecate client state

adespawn:
As suggested by @wprzytula during last meeting

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -626,12 +626,10 @@ class Client extends events.EventEmitter {
     }
 
     /**
-     * Gets a snapshot containing information on the connections pools held by this Client at the current time.
+     * @returns {ClientState} A dummy [ClientState]{@linkcode module:metadata~ClientState} instance.
      *
-     * The information provided in the returned object only represents the state at the moment this method was called and
-     * it's not maintained in sync with the driver metadata.
-     *
-     * @returns {ClientState} A [ClientState]{@linkcode module:metadata~ClientState} instance.
+     * @deprecated This is not planned feature for the driver. Currently this remains in place, but returns Client state with
+     * no information. This endpoint may be removed at any point.
      */
     getState() {
         return ClientState.from(this);
--- a/lib/metadata/client-state.js
+++ b/lib/metadata/client-state.js
@@ -1,103 +1,43 @@
 "use strict";
 
-const util = require("util");
-const errors = require("../errors");
+let deprecatedMsg =
+    "Client state is deprecated and currently provides no functionality!";
 
 /**
  * Represents the state of a {@link Client}.
- * <p>
+ *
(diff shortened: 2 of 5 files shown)


lib/client.js
```diff
@@ -626,12 +626,10 @@ class Client extends events.EventEmitter {
...
 
     /**
-     * Gets a snapshot containing information on the connections pools held by this Client at the current time.
+     * @returns {ClientState} A dummy [ClientState]{@linkcode module:metadata~ClientState} instance.
      *
-     * The information provided in the returned object only represents the state at the moment this method was called and
-     * it's not maintained in sync with the driver metadata.
-     *
-     * @returns {ClientState} A [ClientState]{@linkcode module:metadata~ClientState} instance.
+     * @deprecated This is not planed feature for the driver. Currently this remains in place, but returns Client state with
```
wprzytula, line 631: :pick: `planed` -> `planned`

lib/metadata/client-state.js
```diff
@@ -1,103 +1,43 @@
...
 
 /**
  * Represents the state of a {@link Client}.
- * <p>
+ *
  * Exposes information on the connections maintained by a Client at a specific time.
- * </p>
  * @alias module:metadata~ClientState
- * @constructor
+ * @deprecated This is not planed feature for the driver. Currently this remains in place, but contains no information.
```
wprzytula, line 11: :pick: `planed` -> `planned`

test/integration/broken/client-pool-tests.js
```diff
@@ -276,17 +276,18 @@ describe("Client", function () {
...
                         assert.strictEqual(client.hosts.length, 3);
-                        const state = client.getState();
+                        // Cannot test that: getState deprecated
+                        /* const state = client.getState(); */
                         client.hosts.forEach(function (host) {
                             assert.strictEqual(
                                 host.pool.connections.length,
                                 3,
                                 "For host " + host.address,
                             );
```
wprzytula, line 286: :question: Is `host.pool` still valid, even though we are using Rust driver??
I thought we no longer use the Datastax driver's connections and connection pools.

wprzytula (changes requested):

test/integration/broken/client-pool-tests.js
adespawn: This field is currently invalid (as this is a broken integration test), but I don't think we decided what to do with this field in the driver (can we provide maybe read only functionality or if we can't do anything with it).

lib/client.js
wprzytula: I think this has still not been fixed.

adespawn: I don't know how it slipped my attention

wprzytula (approved):

PR #133, opened 2025-01-20
Fix fmt::Display for LocalTime

Stapox35:
One allocation left in Display:
`self.nanosecond.to_string().chars().count()`
I wanted to convert to logarithm, but in Rust `log10()` for `f64` is non-deterministic.  [(source)](https://doc.rust-lang.org/std/primitive.f64.html#method.log10) I think this can cause problems around powers of 10.

### DIFF
--- a/src/types/local_time.rs
+++ b/src/types/local_time.rs
@@ -1,7 +1,7 @@
 use crate::utils::{bigint_to_i64, js_error};
-use core::fmt;
 use napi::bindgen_prelude::BigInt;
 use scylla::frame::value::CqlTime;
+use std::fmt::{self, Write};
 use std::num::ParseIntError;
 
 const NANO_SEC_IN_SEC: i64 = 1000000000;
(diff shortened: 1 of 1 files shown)


src/types/local_time.rs
```diff
@@ -92,34 +92,30 @@ impl LocalTimeWrapper {
...
+        } else {
+            write!(f, "{}:", self.minute)?;
+        }
+        if self.second < 10 {
+            write!(f, "0{}", self.second)?;
+        } else {
+            write!(f, "{}", self.second)?;
+        }
+        if self.nanosecond > 0 {
+            let zeros = 9 - self.nanosecond.to_string().chars().count();
```
wprzytula, line 111: You could write a custom `fmt::Write` implementation that only counts the written characters, feeding them into the void on the fly (generated by ChatGPT):
```rust
use std::fmt;

// A struct that acts as a character counter
pub struct CharCounter {
    pub count: usize,
}

impl CharCounter {
    // Create a new CharCounter instance
    pub fn new() -> Self {
        CharCounter { count: 0 }
    }

    // Reset the counter to zero
    pub fn reset(&mut self) {
        self.count = 0;
    }
}

impl fmt::Write for CharCounter {
    fn write_str(&mut self, s: &str) -> fmt::Result {
        // Increment the count by the length of the input string
        self.count += s.len();
        Ok(())
    }
}

fn main() {
    let mut counter = CharCounter::new();

    // Use the CharCounter with `write!`
    write!(&mut counter, "Hello, {}!", "world").unwrap();
    println!("Formatted output length: {}", counter.count);

    // Reset the counter and use it again
    counter.reset();
    write!(&mut counter, "The number is: {}", 42).unwrap();
    println!("Formatted output length: {}", counter.count);
}

```

src/types/local_time.rs
```diff
@@ -92,34 +92,52 @@ impl LocalTimeWrapper {
...
+        }
+        if self.nanosecond > 0 {
+            let mut zeros = CharCounter::new();
+            write!(&mut zeros, "{}", self.nanosecond)?;
+            let mut nanos = self.nanosecond;
+            while nanos % 10 == 0 {
+                nanos /= 10;
             }
-        )
+            write!(f, ".{}{}", "0".repeat(9 - zeros.count()), nanos)?;
```
wprzytula, line 117: :wrench: `str::repeat` allocates a `String`, so it's still not a good solution.

src/types/local_time.rs
```diff
@@ -92,34 +92,52 @@ impl LocalTimeWrapper {
...
+        }
+
+        Ok(())
+    }
```
wprzytula, line 121: :question: Have you tested this manually?
:wrench: Unit tests are welcome.

Stapox35: We have unit tests for this in `test/unit/local-time-tests.js` for function `toString()`.

adespawn (approved):

wprzytula (approved):
:100:

PR #140, opened 2025-02-23
Unprepared queries

adespawn:
Adds support for type hints and type guessing to allow executing unprepared queries.
Type guessing is taken from the Datastax driver code. 
Parametr hints are converted into js object on the JS side, which is then converted to the Complex Type, used by the driver. 

Enables (all?) integration tests that rely on hint / type guessing.


src/types/type_hints.rs
```diff
@@ -0,0 +1,93 @@
...
+            return Err(js_error(format!(
+                "Number {} doesn't represent CQL type",
+                value
+            )))
+        }
+    })
+}
+
+#[napi]
+pub fn convert_hint(hint: JsObject) -> napi::Result<ComplexType> {
```
wprzytula: :question: Don't we incur a big penalty for calling JS dynamic API (interacting with `JsObject`) from Rust, through the cross-language binding layer?

src/types/type_hints.rs
```diff
@@ -0,0 +1,93 @@
...
+    let base_type: Result<i32, napi::Error> = hint.get_named_property("code");
```
wprzytula, line 54: :warning: When I see dynamic selection of fields by name in Rust, I start to tremble...
:question: :thought_balloon: Could we rely on napi-rs instead to convert the hint to a valid Rust struct, to further take advantage of Rust's strong type system and type safety?

src/types/type_hints.rs
```diff
@@ -0,0 +1,93 @@
...
+            let support_types: Vec<JsObject> = hint.get_named_property("info")?;
+
+            let support_types: Vec<Result<ComplexType, napi::Error>> =
+                support_types.into_iter().map(convert_hint).collect();
+
+            if support_types.iter().any(|e| e.is_err()) {
+                return Err(js_error("Failed to convert one of the subtypes"));
+            }
+            let mut support_types: Vec<ComplexType> =
+                support_types.into_iter().filter_map(|e| e.ok()).collect();
```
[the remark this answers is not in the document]
wprzytula:
```rust
            let support_types: Vec<ComplexType> = support_types.into_iter()
                .map(convert_hint)
                .collect::<Result<_, napi::Error>>()
                .map_err(|err| js_error(format!("Failed to convert one of the subtypes: {}", err))?;
```

src/types/type_hints.rs
```diff
@@ -0,0 +1,93 @@
...
+                value
+            )))
+        }
+    })
+}
+
+#[napi]
+pub fn convert_hint(hint: JsObject) -> napi::Result<ComplexType> {
+    let base_type: Result<i32, napi::Error> = hint.get_named_property("code");
+    let base_type = number_to_types(base_type?)?;
```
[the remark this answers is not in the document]
wprzytula:
```suggestion
    let base_type: CqlType = number_to_types(hint.get_named_property("code")?)?;
```

src/types/type_hints.rs
```diff
@@ -0,0 +1,93 @@
...
+
+                    // With the length check we assume, this values will always be Some,
+                    // but we keep it as Option, as used Complex type constructor assumes Option
+                    let second_support = support_types.pop();
+                    let first_support = support_types.pop();
+                    ComplexType::two_support(base_type, first_support, second_support)
+                }
+                // TODO: update it with tuple implementations
+                CqlType::Tuple => todo!(),
+                _ => panic!("Invalid branch"),
```
wprzytula, line 95: :pick: use `unreachable!()`.

src/types/type_hints.rs
```diff
@@ -0,0 +1,93 @@
...
+                        return Err(js_error(format!(
+                            "Invalid number of support types. Got {}, expected 2",
+                            support_types.len()
+                        )));
+                    }
+
+                    // With the length check we assume, this values will always be Some,
+                    // but we keep it as Option, as used Complex type constructor assumes Option
+                    let second_support = support_types.pop();
+                    let first_support = support_types.pop();
```
wprzytula: :recycle:

wprzytula:
```rust
                    match [CqlType; 2]::try_from(support_types) {
                        Err(support_types) => Err(js_error(format!(
                            "Invalid number of support types. Expected 2, got {}: {:?}",
                            support_types.len()
                            support_types
                        )));
                        Ok([first_support, second_support]) => {
                            ComplexType::two_support(base_type, first_support, second_support)
                        }
                    }

File: test/integration/supported/client-execute-tests.js [hunk 7]
```diff
@@ -71,11 +73,13 @@ describe("Client @SERVER_API", function () {
...
                         client.execute(query, params, (err) => {
-                            helper.assertInstanceOf(err, errors.ResponseError);
+                            // TODO: Ensure correct error is thrown
+                            helper.assertInstanceOf(err, Error);
+                            /* helper.assertInstanceOf(err, errors.ResponseError);
                             assert.strictEqual(
                                 err.code,
                                 types.responseErrorCodes.invalid,
-                            );
+                            ); */
```
wprzytula, line 82: :memo: The reason why this test expects `ResponseError` and Rust driver returns `SerializationError` is that Rust driver quietly prepares statements in order to type-check (and this way avoid silent bugs around type casts). This is not the case in the Datastax JS driver, which puts the responsibility on erroring out the request on the server side.

lib/client.js
```diff
@@ -589,31 +589,49 @@ class Client extends events.EventEmitter {
...
         await this.#connect();
 
-        const _execOptions = executionOptions.DefaultExecutionOptions.create(
+        const execOptions = executionOptions.DefaultExecutionOptions.create(
             options,
             this,
         );
 
-        let preparedQueries = [];
+        let prepared = execOptions.isPrepared();
```
wprzytula, line 597: :pick: naming

wprzytula:
```suggestion
        let should_be_prepared = execOptions.shouldBePrepared();
```

lib/client.js
```diff
@@ -589,31 +589,49 @@ class Client extends events.EventEmitter {
...
         for (let i = 0; i < queries.length; i++) {
             let element = queries[i];
-            if (typeof element === "string") {
-                preparedQueries.push(
-                    await this.rustClient.prepareStatement(element),
-                );
-                parametersRows.push([]);
+            if (!element) {
+                throw new errors.ArgumentError(`Invalid query at index ${i}`);
+            }
```
wprzytula, line 606: :question: What objects evaluate as false here?

lib/client.js
```diff
@@ -589,31 +589,49 @@ class Client extends events.EventEmitter {
...
+            /**
+             * @type {rust.PreparedStatementWrapper | string}
+             */
+            let query = typeof element === "string" ? element : element.query;
```
wprzytula, line 610: :question: Is this anyhow enforced by TypeScript type checker? What's the purpose of this comment? Is it purely informative?

lib/client.js
```diff
@@ -589,31 +589,49 @@ class Client extends events.EventEmitter {
...
+            let items = element.params || [];
```
wprzytula, line 611: :wrench: `items` as a name tells me nothing. Maybe `params`?

lib/client.js
```diff
@@ -589,31 +589,49 @@ class Client extends events.EventEmitter {
...
-                let prepared = await this.rustClient.prepareStatement(query);
-                let parameters = parseParams(prepared, params);
-                preparedQueries.push(prepared);
-                parametersRows.push(parameters);
+                types = convertHints(hints[i] || []);
+            }
+
+            if (items) {
+                items = parseParams(types, items, prepared === false);
             }
```
wprzytula, line 627: :recycle: Ugh. Does `items` change its type? From what type to what type? It's complicated...

test/integration/supported/client-batch-tests.js
```diff
@@ -392,7 +392,8 @@ describe("Client @SERVER_API", function () {
                     done,
                 );
             },
-        ); */
+        );
+
         // Currently no support for timestamps
         // TODO: Fix this test
         /* vit("2.1", "should support protocol level timestamp", function (done) {
```
wprzytula, line 399: :memo: FYI: timestamps are already supported in Rust driver.

test/integration/supported/client-execute-tests.js
```diff
@@ -242,7 +242,7 @@ describe("Client @SERVER_API", function () {
 
         // No support for parameter hints
         // TODO: Fix this test
-        /* vit(
+        vit(
```
wprzytula, line 245: :recycle: leftover TODO comment

lib/types/cql-utils.js
```diff
@@ -21,6 +21,22 @@ function ensureValue(val, message) {
...
+function guessTypeChecked(value) {
+    let type = rust.convertHint(guessType(value));
+    if (!type) {
+        throw new TypeError(
+            "Target data type could not be guessed, you should use prepared statements for accurate type mapping. Value: " +
+                util.inspect(value),
+        );
+    }
+    return type;
+}
```
wprzytula, line 38: :thinking: It bothers me that `guessTypeChecked` is not just a `guessType` with a correctness check, but it also has `convertHint` packed in. Again, I don't fully understand the distinction between `guessType` and `convertHint`.

wprzytula (changes requested):

lib/client.js
```diff
@@ -795,24 +801,34 @@ class Client extends events.EventEmitter {
...
+            let result;
+            if (execOptions.isPrepared()) {
+                // Execute prepared statement, as requested by the user
                 let statement = await this.rustClient.prepareStatement(query);
-                let parsedParams = parseParams(statement, params);
-                rusty = await this.rustClient.executePreparedUnpaged(
+                let parsedParams = parseParams(
+                    statement.getExpectedTypes(),
+                    params,
+                );
```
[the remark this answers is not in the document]
adespawn: If more arguments are expected, they are filled with undefined value

src/types/type_hints.rs
adespawn: I don't think we tested this explicitly. Here I assumed, it's better to call rust function once, and then trying to serialize values, rather than calling rust function multiple times. The remaining option here is to flatten this type hint object in js, but I believe the difference would be invisible unless users create types nested more then one - two in depth.

src/types/type_hints.rs
```diff
@@ -0,0 +1,93 @@
...
+                    ComplexType::two_support(base_type, first_support, second_support)
+                }
```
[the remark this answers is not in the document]
adespawn: Up to the limit we impose.

lib/client.js
adespawn: https://docs.datastax.com/en/developer/nodejs-driver/4.7/api/class.ExecutionOptions/index.html#is-prepared

lib/client.js
adespawn: Null, undefined, false boolean and similar...

lib/client.js
adespawn: Is it purely informative

lib/client.js
adespawn: Yes. It's changing the type. From array of arbitrary type elements, to array of QueryParameterWrapper

test/integration/supported/client-batch-tests.js
adespawn: #92

lib/client.js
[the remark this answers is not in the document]
wprzytula: :question: :thinking: Is `undefined` OK for an argument where we expect a boolean to be passed...? Ugh, I hate those dynamically typed languages...

lib/client.js
wprzytula: And what cases are expected here?
I mean, we error out if some invalid object that is present on your list is provided. But we don't error out if some invalid object **not** present on your list is provided, e.g., `{"tata": "kot"}` or whatever else.

If this is supposed to be an error check, then it's a much leaky error check (in a sense that it only catches a subset of error situations).

lib/client.js
wprzytula: In such case, I encourage you to add more type hints, as you did above with the doc comment.

lib/client.js
```diff
@@ -795,24 +801,34 @@ class Client extends events.EventEmitter {
...
-            // Optimize: If we have no parameters we can skip preparation step
-            if (!params) {
-                rusty = await this.rustClient.queryUnpagedNoValues(query);
-            }
-            // Otherwise prepare the statement to know the types
-            else {
+            let result;
+            if (execOptions.isPrepared()) {
+                // Execute prepared statement, as requested by the user
                 let statement = await this.rustClient.prepareStatement(query);
```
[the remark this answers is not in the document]
adespawn: Again, https://docs.datastax.com/en/developer/nodejs-driver/4.7/api/class.ExecutionOptions/index.html#is-prepared

lib/client.js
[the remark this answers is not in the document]
adespawn: If we check it like ``if(boolean)`` or ``if(boolean === true)`` (which is the case in all / most parts of the code), then it will work as intended

lib/client.js
```diff
@@ -795,24 +801,34 @@ class Client extends events.EventEmitter {
...
+                result = await this.rustClient.executePreparedUnpaged(
                     statement,
                     parsedParams,
                 );
+            } else if (!params) {
+                // Optimize: If we have no parameters we can skip preparation step
+                result = await this.rustClient.queryUnpagedNoValues(query);
+            } else {
+                let expectedTypes = execOptions.getHints() || [];
+                let parsedParams = parseParams(expectedTypes, params, true);
```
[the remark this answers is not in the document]
adespawn: for the documentation of this function I updated it so it shows as optional argument. If you believe this is still not enough, we can talk about it during the meeting

lib/client.js
adespawn: What we want to accept here, will be ``"just a Query String"`` and ``{query: "Object with query", params: ["and", "some", 4, "parameters"]}``.
Here objects evaluating to false will throw an error, while other invalid values will throw errors in on of the few following places:
``if (!query) {`` few lines down, and in places where types are explicitly checked like when calling rust.

lib/types/cql-utils.js
adespawn: Covert hint takes already guessed type, in a uniform format, and converts it to the object recognized by rust part of the code, while guessType takes an object and tries to guess it's type. Convert hint is also used when we have type suggestions, not only when we guess a type, and with the current implementation it cannot throw errors when it encounters empty type hint.

lib/types/cql-utils.js
```diff
@@ -179,7 +179,7 @@ function getWrapped(type, value) {
 /**
  * Parses array of params into rust objects according to preparedStatement expected types
  * Throws ResponseError when received different amount of parameters than expected
- * @param {Array<rust.ComplexType>} expectedTypes
+ * @param {Array<rust.ComplexType | null>} expectedTypes
```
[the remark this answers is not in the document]
adespawn: It's somewhere in between. For this reason I decided to put it in this commit with other changes to this file

lib/types/cql-utils.js
```diff
@@ -218,15 +218,56 @@ function parseParams(expectedTypes, params, allowGuessing) {
...
+            if (typeof objectHint.code !== "number") {
+                throw new TypeError(
+                    "Type information not valid, only String and Number values are valid hints",
+                );
+            }
+            result.push(rust.convertHint(objectHint));
+        } else {
+            result.push(null);
+        }
+    }
```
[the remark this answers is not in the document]
adespawn: getHints provides the value of hint field given by the user (this is [public API of datastax](https://docs.datastax.com/en/developer/nodejs-driver/4.7/api/class.ExecutionOptions/index.html#get-hints)), and this converts them to internal type representation

src/session.rs
```diff
@@ -70,6 +70,27 @@ impl SessionWrapper {
...
+    /// If the provided types will not be correct, this query will fail.
+    #[napi]
+    pub async fn query_unpaged(
+        &self,
+        query: String,
+        params: Vec<Option<&QueryParameterWrapper>>,
+    ) -> napi::Result<QueryResultWrapper> {
+        let params_vec: Vec<Option<CqlValue>> = QueryParameterWrapper::extract_parameters(params);
+        let query_result = self
+            .internal
```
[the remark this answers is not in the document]
adespawn: This name was created some time ago (and it passed first iteration review). If you believe this would be meaningful I can change this name.

src/types/type_hints.rs
adespawn: At some point, we need to verify the type of the value. To take advantage of napi-rs converting the values, we would need to create a custom struct and (probably) convert current object into this new struct, as we do all over the code now. With this approach I hoped to simplify this process. We still rely on rust typing, but instead of getting the values as function arguments, and throwing errors there, we do almost the same in the body of the function

lib/client.js
```diff
@@ -795,24 +801,34 @@ class Client extends events.EventEmitter {
...
+                );
+                result = await this.rustClient.executePreparedUnpaged(
                     statement,
                     parsedParams,
                 );
+            } else if (!params) {
+                // Optimize: If we have no parameters we can skip preparation step
+                result = await this.rustClient.queryUnpagedNoValues(query);
+            } else {
+                let expectedTypes = execOptions.getHints() || [];
```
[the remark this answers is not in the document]
adespawn: This, at least at the extend used in Datastax code, is the syntax similar to rust ``unwrap_or`` on option and I believe it would be counterproductive to explain ech occurrence of this syntax

src/types/type_hints.rs
adespawn: This part of code was modified in the following commit and I believe this no longer works for the code supporting partial hints.

adespawn:
All questions should be either fixed, or addressed. If I missed some, let me know

lib/client.js
wprzytula: :question: Does it make sense at all to have such explicit check here, if later on those errors will be caught anyway as you said? E.g., when Rust code is called.

In general, let's have as concise set of checks as possible. Otherwise, they might be redundant and/or confusing (as it happened to me here).

lib/types/cql-utils.js
wprzytula: Please put these considerations in respective comments and docstrings.

lib/client.js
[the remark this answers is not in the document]
wprzytula: OK, disregard the part about the syntax being cryptic - it's probably me not being familiar with JS.

My question are still valid, though:
> What's the consequence of passing an empty parameter list in such case?

If I guess correctly, there are two cases:
1. `expectedTypes` is empty. Then ... (please describe the semantics and the logic following)
2. `expectedTypes` contain correct expected types. Then ... (ditto)

lib/client.js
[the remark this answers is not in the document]
wprzytula: :question: Is it common for JS code to use optional arguments? Rust doesn't have them at all, that's why I'm that sceptical towards them.

src/session.rs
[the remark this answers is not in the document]
wprzytula: Yes please, let's make our code more idiomatic as part of next iteration efforts.

src/types/type_hints.rs
wprzytula: Correct me if I get you wrong.
There are two approaches that you describe:
1. (the approach taken here) Passing a `JsObject` to Rust and dynamically retrieving needed data.
2. (the approach you claim to be taken elsewhere in the code) Building up (how?) a new Rust struct by converting (how? where? In our JS code or hidden in napi-rs?) JS object.

:question: Isn't it feasible to simply pass a JS object to napi layer, putting the whole conversion and error handling burden on it, and on the Rust side get a correctly populated Rust struct designed for the case?

wprzytula:
> All questions should be either fixed, or addressed. If I missed some, let me know

Indeed, everything was addressed. I've left another round of comments.

src/types/type_hints.rs
adespawn: 2. structs: QueryParameterWrapper, ComplexTypes, ...

Isn't it feasible? - Not that I'm aware of

lib/client.js
adespawn: In this case, without this typecheck ``TypeError: undefined has no properties`` will be thrown, (if the value is undefined) - so this increases the readability of the errors thrown. While errors thrown "later" are still not the most informative, they provide a bit more context what is wrong.

lib/client.js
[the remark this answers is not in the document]
adespawn: Explained this in the parseParams docs:

If ``allowGuessing`` is true, then for each missing field of ``expectedTypes``, this function will try to guess a type.
If the type cannot be guessed, error will be thrown.
Field is missing if it is null, undefined (or if the ``expectedTypes`` list is to short)

src/types/type_hints.rs
wprzytula: There are questions in my above comment that I'd like you to answer, because I feel I lack something.

lib/client.js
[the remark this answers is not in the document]
adespawn: ~_Note to myself. always double check comments on train XD_~

Yes. Even our driver most important functions have optional parameters: ``execute``,  ``batch`` and many more

src/types/type_hints.rs
```diff
@@ -0,0 +1,104 @@
...
+                value
+            )))
+        }
+    })
+}
+
+#[napi]
+/// Convert single hint from the format expected by API of the driver
+/// to the ``ComplexType`` that is used internally in the rust part of the code
+pub fn convert_hint(hint: Option<JsObject>) -> napi::Result<Option<ComplexType>> {
```
wprzytula, line 49: :pick: `#[napi]` should be after docstring.

wprzytula (changes requested):

src/types/type_hints.rs
```diff
@@ -46,22 +50,31 @@ fn type_code_to_cql_type(value: i32) -> Result<CqlType, napi::Error> {
...
-            let support_type = convert_hint(hint.get_named_property("info").ok())?;
+            // Ensure there is no value in info field, or there is exactly one support type
+            let mut support_types = support_types.unwrap_or_default();
+            if support_types.len() > 1 {
+                return Err(js_error(format!(
+                    "Invalid number of support types. Got {}, expected no more than one.",
+                    support_types.len()
+                )));
+            }
+            let support_type = convert_hint(support_types.pop())?;
```
wprzytula, line 73: :question: Should this belong to this commit? (`Make use of napi object`)

src/types/type_hints.rs
```diff
@@ -0,0 +1,95 @@
...
+            let support_types: Vec<JsObject> = hint.get_named_property("info")?;
+
+            let support_types: Vec<Result<ComplexType, napi::Error>> =
+                support_types.into_iter().map(convert_hint).collect();
+
+            if support_types.iter().any(|e| e.is_err()) {
+                return Err(js_error("Failed to convert one of the subtypes"));
+            }
+            let mut support_types: Vec<ComplexType> =
+                support_types.into_iter().filter_map(|e| e.ok()).collect();
```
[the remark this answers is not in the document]
wprzytula: Ah, it's actually done in a later commit.
:recycle: Let's not introduce something and then refactor in the same PR; please squash these commits.

wprzytula (changes requested):

src/session.rs
```diff
@@ -160,14 +160,23 @@ impl SessionWrapper {
...
+pub fn create_prepared_batch(queries: Vec<&PreparedStatementWrapper>) -> BatchWrapper {
     let mut batch: Batch = Default::default();
     queries
         .iter()
         .for_each(|q| batch.append_statement(q.prepared.clone()));
     BatchWrapper { inner: batch }
 }
 
+#[napi]
+pub fn create_unprepared_batch(queries: Vec<String>) -> BatchWrapper {
```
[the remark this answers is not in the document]
adespawn: > ❓ Can we accept `Vec<&str>` here?

https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/pull/140#discussion_r1973402561

src/types/type_hints.rs
adespawn: Yes...?
Adding NAPI object means we get Vec instead of single element, so we need to check the amount of elements

src/session.rs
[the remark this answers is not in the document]
wprzytula: :clown_face:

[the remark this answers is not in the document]
wprzytula: > 📌 Well... the proper fix would be to add `From<String> for BatchStatement` impl in the Rust driver. Will you?

This is my best suggestion. WDYT?

src/types/type_hints.rs
```diff
@@ -0,0 +1,120 @@
...
+use super::type_wrappers::{ComplexType, CqlType};
+
+// When converting from JS object, it the field is not provided (ex: {code: 3})
+// Then info field becomes None.
+// When the field is set to null (ex: {code: 3, info: null})
+// THen info field becomes Some(None).
+// For this reason we need to have nested option in this object.
+
+#[napi(object)]
+pub struct TypeHint {
```
wprzytula, line 12:
```suggestion
// When converting from JS object, if a field is not provided (ex: {code: 3}),
// then the field becomes None.
// When the field is set to null (ex: {code: 3, info: null}),
// then the field becomes Some(None).
// As we want to support both lack of `info` field and `info` field set to null,
// we need to have nested Option in this object.
#[napi(object)]
pub struct TypeHint {
```

wprzytula (changes requested):

src/session.rs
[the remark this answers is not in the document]
adespawn: We can implement this, but I would say at a later time.

lib/types/type-guessing.js
```diff
@@ -0,0 +1,91 @@
...
+        code = typeValues.time;
+    } else if (value instanceof types.Duration) {
+        code = typeValues.duration;
+    }
+    // Map JS TypedArrays onto vectors
+    else if (isTypedArray(value)) {
+        // TODO: Add support for typed arrays
+        throw new errors.DriverInternalError(
+            "No support for typed array guessing type",
+        );
```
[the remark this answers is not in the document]
adespawn: Now we have: #153

lib/types/cql-utils.js
```diff
@@ -21,6 +22,22 @@ function ensureValue(val, message) {
         throw new TypeError(message);
 }
 
+/**
+ * Guess the type, and if the type cannot be guessed, throws an error.
```
ZuzaOsa, line 26: Either change `guess` to `guesses` or `throws` to `throw`

ZuzaOsa (approved):
One small comment, otherwise lgtm

PR #146, opened 2025-03-04
[DRAFT] Encoder parity

adespawn:


lib/types/cql-utils.js
```diff
@@ -129,12 +130,32 @@ function getWrapped(type, value) {
...
+            }
+            if (value instanceof Date) {
+                // milliseconds since epoch
+                value = value.getTime();
+                if (isNaN(value)) {
+                    throw new TypeError("Invalid date: " + supportElement);
+                }
+            }
+
             return rust.QueryParameterWrapper.fromTimestamp(BigInt(value));
```
wprzytula, line 159: :recycle: I can see `supportElement` used only in this `case`. Then, why not define the variable (or rather constant) as `const` only in this `case`?

lib/types/cql-utils.js
```diff
@@ -65,6 +65,16 @@ function encodeMap(value, parentType) {
...
+/**
+ * Ensures the provided value is non NaN. If it is, throws an error
+ * @param {any} value
+ */
+function ensureNumber(value) {
+    if (isNaN(value)) {
+        throw new TypeError("Expected Number, obtained " + util.inspect(value));
+    }
+}
+
```
wprzytula, line 78: :question: Have you considered [the quirks of `isNaN()`])https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN)?
:thought_balloon: Maybe we should instead use `Number.isNaN()`, as suggested in the link?

test/integration/supported/numeric-tests.js
```diff
@@ -96,31 +96,15 @@ module.exports = function (keyspace, prepare) {
...
-            const params = [
-                id,
-                [intValue],
-                [decimalValue],
-                { a: decimalValue },
-                [decimalValue],
-                { a: intValue },
-                [intValue],
-            ];
+            const params = [id, [intValue], [decimalValue], { a: intValue }];
```
wprzytula, line 107: :wrench: Please explain here and in the commit message why each of these encodings is no longer supported.

wprzytula: BTW: if a commit breaks some tests, those should be adjusted in the same commit. _**All commits should compile and pass tests.**_

wprzytula (changes requested):

test/integration/supported/numeric-tests.js
adespawn: Those tests are currently, so this changes behavior of not enabled (but planned) integration tests.

lib/types/cql-utils.js
adespawn: This is how it was done in Datastax driver, so I will look into it and we can discuss this tomorrow

lib/types/cql-utils.js
adespawn: It's against the eslint rules: https://eslint.org/docs/latest/rules/no-case-declarations

lib/new-utils.js
```diff
@@ -88,8 +89,24 @@ function longToBigint(from) {
     return r;
 }
 
+/**
+ * Ensure the value is one of the accepted numeric types, and convert them to BigInt
+ * @param {string | number | Long | BigInt} value
```
ZuzaOsa, line 94: I think buffer should also be accepted as a value convertible to BigInt

adespawn: If I understand the datastax code correctly, any value can be provided as a buffet, so I would leave this to a separate PR.

lib/types/cql-utils.js
```diff
@@ -101,6 +101,7 @@ function getWrapped(type, value) {
         return null;
     }
     let encodedElement;
+    let supportElement;
```
ZuzaOsa, line 115: Maybe `supportElement` is not the best name?

lib/types/cql-utils.js
```diff
@@ -180,20 +201,31 @@ function getWrapped(type, value) {
...
+            }
+
             return rust.QueryParameterWrapper.fromLocalTime(
                 value.getInternal(),
             );
         case rust.CqlType.TinyInt:
             return rust.QueryParameterWrapper.fromTinyInt(value);
         case rust.CqlType.Uuid:
-            if (!(value instanceof Uuid))
+            // Providing UUID as string is a parity with the old driver
```
ZuzaOsa, line 235: This comment is not consistent with the comment about Inet.

lib/types/cql-utils.js
```diff
@@ -176,22 +213,35 @@ function getWrapped(type, value) {
             encodedElement = encodeMap(value, type);
             return rust.QueryParameterWrapper.fromMap(encodedElement);
         case rust.CqlType.SmallInt:
+            ensureNumber(value);
             return rust.QueryParameterWrapper.fromSmallInt(value);
         case rust.CqlType.Time:
-            if (!(value instanceof LocalTime))
+            if (typeof value == "string") {
```
ZuzaOsa, line 219: Comment with an explanation is missing.

lib/types/cql-utils.js
wprzytula: _support types_ is not a good name in general. I, however, have failed to think up a better one... :/

wprzytula (approved):

lib/types/cql-utils.js
ZuzaOsa: Here `supportElement` is not the same as in the Complex Type, it is used as a name for a temporary variable, so I think calling it for example `tmp` would be better.

wprzytula: Just `tmp` is IMO bad... variable names need to be descriptive, especially in JS, where variables have no types to serve as a hint.

wprzytula (approved):

PR #152, opened 2025-03-08
Fix integration tests problems

adespawn:
Integration tests appear to have issues in certain conditions, when run on github CI, which depends on the order in which they are run, which tests are enabled (there is no single failing test). I was unable to reproduce the issue on any other machine than Github workers.

The issue appears as one of the following problems:
- capacity overflow
- memory allocation failure

After some investigation, while I wasn't able to pinpoint a specific issue, I determined that the issue arises from concurrent JS code interacting with the NAPI-RS layer.

This commit attempts to stop the problems, by removing the concurrency. This severely limits the speed at which executeConcurrent is performed, but appears to fully resolve the issue (test size = 30).

With this in mind, any concurrency should be implemented on Rust side, or the root issue should be determined before removing the concurrency restrictions.

### DIFF
--- a/lib/concurrent/index.js
+++ b/lib/concurrent/index.js
@@ -2,6 +2,8 @@
 
 const { Stream } = require("stream");
 const utils = require("../utils");
+const { Mutex } = require("async-mutex");
+const { env } = require("process");
 
 /**
  * Utilities for concurrent query execution with the DataStax Node.js Driver.
@@ -109,6 +111,9 @@ class ArrayBasedExecutor {
             options.concurrencyLevel || 100,
             this._parameters.length,
         );
+        if (env.NO_CONCURRENCY) {
+            this._concurrencyLevel = 1;
+        }
         this._queryOptions = {
             prepare: true,
             executionProfile: options.executionProfile,
@@ -176,6 +181,7 @@ class StreamBasedExecutor {
      * @private
      */
     constructor(client, query, stream, options) {
+        this._mutex = new Mutex();
         this._client = client;
         this._query = query;
         this._stream = stream;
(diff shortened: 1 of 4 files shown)


adespawn:
@wprzytula I've updated this PR, to make the concurrency fix more meaningfull

adespawn:
Note to myself: Make this optional

adespawn:
@wprzytula Added discussed changes. It should be ready for a review

lib/concurrent/index.js
```diff
@@ -109,6 +111,9 @@ class ArrayBasedExecutor {
             options.concurrencyLevel || 100,
             this._parameters.length,
         );
+        if (env.STOP_CONCURRENCY) {
+            this._concurrencyLevel = 1;
+        }
```
wprzytula, line 116: :pick: maybe `NO_CONCURRENCY`?

wprzytula: Also, please actually enable this mitigation in the CI, in a new commit.

lib/concurrent/index.js
```diff
@@ -117,14 +122,15 @@ class ArrayBasedExecutor {
...
+    async execute() {
         const promises = new Array(this._concurrencyLevel);
 
         for (let i = 0; i < this._concurrencyLevel; i++) {
             promises[i] = this._executeOneAtATime(i, 0);
         }
 
-        return Promise.all(promises).then(() => this._result);
+        await Promise.all(promises);
+        return this._result;
```
wprzytula, line 133: :wrench: This seem unrelated to this commit, and even PR. Either delete this change from this PR or make it a separate commit.

wprzytula (changes requested):

wprzytula:
Well...
image

adespawn:
That's strange. When testing on 40 separate runs there were no failures - maybe it wasn't correctly rebased on the second iteration?

wprzytula (approved):
Has anything changed since my last review?

PR #155, opened 2025-03-10
Introduce CachingSession instead of Session

ZuzaOsa:
This PR introduces caching session for optimizing the execution of prepared statements. It also adds support for the maxPrepared option of ``Client`` and changes the default value of the cache from ``500`` to ``512``.

Slightly improves performance and significantly improves amount of transferred data (number of packet measured with wireshark):
Before:

2.74 milion packets transferred

```bash
sudo perf stat node ./benchmark/logic/concurrent_insert.js scylladb-javascript-driver 1000000


 Performance counter stats for 'node ./benchmark/logic/concurrent_insert.js scylladb-javascript-driver 1000000':

        105,883.59 msec task-clock                       #    1.908 CPUs utilized             
         3,518,881      context-switches                 #   33.233 K/sec                     
           116,370      cpu-migrations                   #    1.099 K/sec                     
           191,451      page-faults                      #    1.808 K/sec                     
   426,824,192,831      cycles                           #    4.031 GHz                       
   129,342,909,399      stalled-cycles-frontend          #   30.30% frontend cycles idle      
   310,650,304,423      instructions                     #    0.73  insn per cycle            
                                                  #    0.42  stalled cycles per insn   
    59,643,874,497      branches                         #  563.297 M/sec                     
     2,479,172,913      branch-misses                    #    4.16% of all branches           

      55.494683305 seconds time elapsed

      72.364812000 seconds user
      35.192590000 seconds sys
```

After:

1.47 milion packets transferred

```bash
sudo perf stat node ./benchmark/logic/concurrent_insert.js scylladb-javascript-driver 1000000

 Performance counter stats for 'node ./benchmark/logic/concurrent_insert.js scylladb-javascript-driver 1000000':

         89,110.66 msec task-clock                       #    1.644 CPUs utilized             
         3,253,277      context-switches                 #   36.508 K/sec                     
            79,698      cpu-migrations                   #  894.371 /sec                      
           190,646      page-faults                      #    2.139 K/sec                     
   357,511,281,194      cycles                           #    4.012 GHz                       
    95,049,526,873      stalled-cycles-frontend          #   26.59% frontend cycles idle      
   272,851,008,791      instructions                     #    0.76  insn per cycle            
                                                  #    0.35  stalled cycles per insn   
    52,113,507,640      branches                         #  584.818 M/sec                     
     1,816,498,178      branch-misses                    #    3.49% of all branches           

      54.199610935 seconds time elapsed

      62.085423000 seconds user
      28.553035000 seconds sys
```

### DIFF
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -96,8 +96,7 @@ const errors = require("./errors.js");
  * @property {Number} [maxPrepared] Determines the maximum amount of different prepared queries before evicting items
  * from the internal cache. Reaching a high threshold hints that the queries are not being reused, like when
  * hard-coding parameter values inside the queries.
- * Default: ``500``.
- * [TODO: Add support for this field]
+ * Default: `512`.
  * @property {Object} [policies]
  * [TODO: Add support for this field]
  * @property {LoadBalancingPolicy} [policies.loadBalancing] The load balancing policy instance to be used to determine
@@ -330,7 +329,7 @@ function defaultOptions() {
         authProvider: null,
         requestTracker: null,
         metrics: new metrics.DefaultMetrics(),
-        maxPrepared: 500,
+        maxPrepared: null, // Default is 512, defined on the Rust side
         refreshSchemaDelay: 1000,
         isMetadataSyncEnabled: true,
         prepareOnAllHosts: true,
@@ -668,6 +667,7 @@ function setRustOptions(options) {
     rustOptions.applicationName = options.applicationName;
     rustOptions.applicationVersion = options.applicationVersion;
     rustOptions.keyspace = options.keyspace;
+    rustOptions.cacheSize = options.maxPrepared;
     if (options.credentials) {
         rustOptions.credentialsUsername = options.credentials.username;
         rustOptions.credentialsPassword = options.credentials.password;
--- a/src/session.rs
+++ b/src/session.rs
@@ -1,4 +1,4 @@
-use scylla::client::session::Session;
+use scylla::client::caching_session::CachingSession;
 use scylla::client::session_builder::SessionBuilder;
 use scylla::client::SelfIdentity;
 use scylla::response::PagingState;
@@ -16,6 +16,8 @@ use crate::{
     requests::request::PreparedStatementWrapper, result::QueryResultWrapper, utils::err_to_napi,
 };
 
+const DEFAULT_CACHE_SIZE: u32 = 512;
+
 #[napi]
 pub struct SessionOptions {
     pub connect_points: Vec<String>,
@@ -24,6 +26,7 @@ pub struct SessionOptions {
     pub application_version: Option<String>,
     pub credentials_username: Option<String>,
     pub credentials_password: Option<String>,
+    pub cache_size: Option<u32>,
 }
 
 #[napi]
@@ -33,7 +36,7 @@ pub struct BatchWrapper {
 
 #[napi]
 pub struct SessionWrapper {
-    inner: Session,
+    inner: CachingSession,
 }
 
 #[napi]
@@ -48,6 +51,7 @@ impl SessionOptions {
             application_version: None,
             credentials_username: None,
             credentials_password: None,
+            cache_size: None,
         }
     }
 }
@@ -59,13 +63,21 @@ impl SessionWrapper {
     pub async fn create_session(options: &SessionOptions) -> napi::Result<Self> {
         let builder = configure_session_builder(options);
         let session = builder.build().await.map_err(err_to_napi)?;
+        let session: CachingSession = CachingSession::from(
+            session,
+            options.cache_size.unwrap_or(DEFAULT_CACHE_SIZE) as usize,
+        );
         Ok(SessionWrapper { inner: session })
     }
 
     /// Returns the name of the current keyspace
     #[napi]
     pub fn get_keyspace(&self) -> Option<String> {
-        self.inner.get_keyspace().as_deref().map(ToOwned::to_owned)
+        self.inner
+            .get_session()
+            .get_keyspace()
+            .as_deref()
+            .map(ToOwned::to_owned)
     }
 
     /// Executes unprepared statement. This assumes the types will be either guessed or provided by user.
@@ -85,6 +97,7 @@ impl SessionWrapper {
         let params_vec = MaybeUnsetQueryParameterWrapper::extract_parameters(params);
         let query_result = self
             .inner
+            .get_session()
             .query_unpaged(statement, params_vec)
             .await
             .map_err(err_to_napi)?;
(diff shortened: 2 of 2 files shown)


src/session.rs
```diff
@@ -53,12 +53,17 @@ impl SessionWrapper {
             .build()
             .await
             .map_err(err_to_napi)?;
+        let s: CachingSession = CachingSession::from(s, 100);
```
adespawn, line 56: Magic constant?

wprzytula (changes requested):

src/session.rs
```diff
@@ -128,6 +140,7 @@ impl SessionWrapper {
         let query = apply_options(query.prepared.clone(), options)?;
         QueryResultWrapper::from_query(
             self.inner
+                .get_session()
                 .execute_unpaged(&query, params_vec)
                 .await
                 .map_err(err_to_napi)?,
```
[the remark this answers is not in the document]
ZuzaOsa: It is introduced in places where we would call `prepare()` on the statement. Instead of preparing it every time we check if the statement has not been prepared before. Currently it happened only in `prepare_statement`.

src/session.rs
```diff
@@ -69,6 +74,7 @@ impl SessionWrapper {
     ) -> napi::Result<QueryResultWrapper> {
         let query_result = self
             .inner
+            .get_session()
             .query_unpaged(query, &[])
             .await
```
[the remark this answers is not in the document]
ZuzaOsa: Introduced issue #169

src/session.rs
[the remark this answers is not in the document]
wprzytula: Then I'm lost.
I thought the Datastax driver works like this:
- Keeps prepared statement cache: `Map<StatementString, PreparedStatement>`
- Upon each execution of a statement (provided by the user as a statement string):
  - looks up the statement in the prepared cache
    - if it's there, it's returned
    - if it's not there, it's prepared first, inserted into the cache and then returned
  - executes the `PreparedStatement`.

This is exactly what `CachingSession` does. It's enough to pass a statement string to `CachingSession::execute{_unpaged,_iter}`  and it'll transparently handle preparation if needed.

Why do we fiddle with all this in our wrapper? I don't see yet.

src/session.rs
```diff
@@ -103,8 +110,13 @@ impl SessionWrapper {
...
         statement: String,
     ) -> napi::Result<PreparedStatementWrapper> {
+        let query: Statement = statement.into();
         Ok(PreparedStatementWrapper {
-            prepared: self.inner.prepare(statement).await.map_err(err_to_napi)?,
+            prepared: self
+                .inner
+                .add_prepared_statement(&query)
+                .await
+                .map_err(err_to_napi)?,
```
wprzytula, line 119: :wrench: Use `add_prepared_statement_owned()` for less cloning.

wprzytula (changes requested):

src/session.rs
wprzytula: Let's write a TODO then to move to that function once it gets public.

src/session.rs
```diff
@@ -55,16 +59,37 @@ impl SessionWrapper {
...
+            .inner
+            .get_session()
+            .query_unpaged(query, &[])
+            .await
+            .map_err(err_to_napi)?;
+        QueryResultWrapper::from_query(query_result)
     }
 
-    /// Executes unprepared statement. This assumes the types will be either guessed or provided by user.
+    /// Executes unprepared query. This assumes the types will be either guessed or provided by user.
```
wprzytula, line 92:
```suggestion
    /// Executes unprepared statement. This assumes the types will be either guessed or provided by user.
```

src/session.rs
```diff
@@ -14,11 +14,14 @@ use crate::{
     requests::request::PreparedStatementWrapper, result::QueryResultWrapper, utils::err_to_napi,
 };
 
+const DEFAULT_CACHE_SIZE: u32 = 500;
```
wprzytula, line 17: :pick: For better alignment, let's use multiples of 2.

```suggestion
const DEFAULT_CACHE_SIZE: u32 = 512;
```

src/session.rs
```diff
@@ -94,8 +120,13 @@ impl SessionWrapper {
         &self,
         statement: String,
     ) -> napi::Result<PreparedStatementWrapper> {
+        let query: Statement = statement.into();
```
wprzytula, line 123:
```suggestion
        let statement: Statement = statement.into();
```

src/session.rs
```diff
@@ -94,8 +120,13 @@ impl SessionWrapper {
...
         Ok(PreparedStatementWrapper {
-            prepared: self.inner.prepare(statement).await.map_err(err_to_napi)?,
+            prepared: self
+                .inner
+                .add_prepared_statement(&query) // TODO: change for add_prepared_statement_to_owned after it is made public
```
wprzytula, line 127:
```suggestion
                .add_prepared_statement(&statement) // TODO: change for add_prepared_statement_to_owned after it is made public
```

wprzytula (changes requested):

lib/client-options.js
```diff
@@ -96,8 +96,7 @@ const errors = require("./errors.js");
  * @property {Number} [maxPrepared] Determines the maximum amount of different prepared queries before evicting items
  * from the internal cache. Reaching a high threshold hints that the queries are not being reused, like when
  * hard-coding parameter values inside the queries.
- * Default: ``500``.
- * [TODO: Add support for this field]
+ * Default: ``512``.
```
adespawn, line 99: 😂

wprzytula (approved):

lib/client-options.js
```diff
@@ -668,6 +667,8 @@ function setRustOptions(options) {
     rustOptions.applicationName = options.applicationName;
     rustOptions.applicationVersion = options.applicationVersion;
     rustOptions.keyspace = options.keyspace;
+    if (options.maxPrepared !== undefined)
+        rustOptions.cacheSize = options.maxPrepared;
```
adespawn, line 671:
```suggestion
    rustOptions.cacheSize = options.maxPrepared;
```

src/session.rs
```diff
@@ -59,13 +63,36 @@ impl SessionWrapper {
...
+        query: String,
+        _options: &QueryOptionsWrapper,
+    ) -> napi::Result<QueryResultWrapper> {
+        let query_result = self
+            .inner
+            .get_session()
+            .query_unpaged(query, &[])
+            .await
+            .map_err(err_to_napi)?;
+        QueryResultWrapper::from_query(query_result)
```
adespawn, line 95: This appears to be a leftover code, that should not be in this PR as it was removed here: https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/commit/fc3ad70e5571c624f8cfd5bb2193364545081f3c

adespawn (changes requested):

lib/client-options.js
```diff
@@ -330,7 +329,7 @@ function defaultOptions() {
         authProvider: null,
         requestTracker: null,
         metrics: new metrics.DefaultMetrics(),
-        maxPrepared: 500,
+        maxPrepared: null,
```
adespawn, line 332: Mention this value is defined in rust code.

adespawn (approved):
One small thing. Otherwise LGTM

adespawn:
I will try to run benchmarks for this PR in a moment

PR #156, opened 2025-03-11
Fix casting values to CqlDate

Stapox35:
Fix casting values to `CqlDate`.

### DIFF
--- a/src/types/local_date.rs
+++ b/src/types/local_date.rs
@@ -69,11 +69,11 @@ impl LocalDateWrapper {
     }
 
     pub fn get_cql_date(&self) -> CqlDate {
-        CqlDate(((1 << 31) + self.value) as u32)
+        CqlDate(((1 << 31) + self.value as i64) as u32)
     }
 
     pub fn from_cql_date(date: CqlDate) -> Self {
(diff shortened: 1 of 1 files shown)


src/types/local_date.rs
```diff
@@ -69,11 +69,11 @@ impl LocalDateWrapper {
     }
 
     pub fn get_cql_date(&self) -> CqlDate {
-        CqlDate(((1 << 31) + self.value) as u32)
+        CqlDate(((1 << 31) + self.value as i64) as u32)
     }
 
     pub fn from_cql_date(date: CqlDate) -> Self {
-        let value: i32 = date.0 as i32 - (1 << 31);
+        let value: i32 = (date.0 as i64 - (1 << 31)) as i32;
```
adespawn, line 76: Do you need do have i32 twice here?

Stapox35: But below I need `i32` for `LocalDateWrapper` (in two places). I think better casting once for `i64` than twice for `i32`.

adespawn: So just remove one of the i32, as I believe one of them is redundant

adespawn (approved):

PR #157, opened 2025-03-11
Typescript tests

ZuzaOsa:
This PR introduces a workflow for running typescript tests of the driver. The tests are taken from the DataStax driver. Entities that are deprecated were excluded from them.


.github/workflows/typescript-tests.yml
```diff
@@ -0,0 +1,49 @@
+name: Typescript tests
```
wprzytula, line 1: Is this workflow taken from the Datastax driver? There neither PR cover letter nor commit message, nor any comments...

test/typescript/api-generation-test.ts
```diff
@@ -37,9 +37,9 @@ export async function generatedFn() {
 `);
 
   printClasses(root, "root", new Set(["Encoder"]));
-  printObjects(root, "root", new Set(["token"]));
+  printObjects(root, "root", new Set(["token", "datastax", "geometry"]));
 
-  printClasses(auth, "auth", new Set(["NoAuthProvider"]));
+  printClasses(auth, "auth", new Set(["NoAuthProvider", "DseGssapiAuthProvider", "DsePlainTextAuthProvider"]));
```
wprzytula, line 42: :question: What's the semantics of these functions? Why are we adding obsolete entities here?

ZuzaOsa: Those functions test generation of API from JS and we need to exclude the deprecated classes or objects that are available in JS (and throw a deprecated error) but throw a compilation error in TS, as we discussed. In the definition of those functions the last argument is a set of entities to exclude.

.github/workflows/typescript-tests.yml
ZuzaOsa: I added a PR descirption. Only the tests are taken from the DataStax driver, this workflow was created by me as DataStax uses a different CI than Github Actions.

test/typescript/api-generation-test.ts
wprzytula: Thanks. Could you please put this explanation in the commit message and in the code here?

adespawn:
Why was the old directory broken?

ZuzaOsa:
> Why was the old directory broken?

Changed message to `not-supported directory`

PR #161, opened 2025-03-18
Unset

PiotrJunior:
Since elements of complex types, such as Lists, consist of
QueryParameterWrapper elements, and according to the CQL protocol,
complex type elements cannot contain Unsets, the structure of
QueryParameterWrapper has been split into QueryParameterWrapper and
MaybeUnsetQueryParameterWrapper to enforce this rule.

### DIFF
--- a/lib/types/cql-utils.js
+++ b/lib/types/cql-utils.js
@@ -60,7 +60,7 @@ function encodeListLike(values, subtype) {
             values[i],
             "A collection can't contain null or unset values",
         );
-        res.push(getWrapped(subtype, values[i]));
+        res.push(wrapValueBasedOnType(subtype, values[i]));
     }
     return res;
 }
@@ -84,8 +84,8 @@ function encodeMap(value, parentType) {
         ensureValue(val, "A collection can't contain null or unset values");
         ensureValue(key, "A collection can't contain null or unset values");
         res.push([
-            getWrapped(keySubtype || guessTypeChecked(key), key),
-            getWrapped(valueSubtype || guessTypeChecked(val), val),
+            wrapValueBasedOnType(keySubtype || guessTypeChecked(key), key),
+            wrapValueBasedOnType(valueSubtype || guessTypeChecked(val), val),
         ]);
     }
     return res;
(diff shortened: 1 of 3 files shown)


lib/types/cql-utils.js
```diff
@@ -81,23 +81,37 @@ function encodeMap(value, parentType) {
...
     return res;
 }
 
 /**
- * Wrap value into QueryParameterWrapper based on the type
+ * Wrap value into MaybeUnsetQueryParameterWrapper based on the type
  * @param {rust.ComplexType} type
  * @param {*} value
- * @returns {rust.QueryParameterWrapper}
+ * @returns {rust.MaybeUnsetQueryParameterWrapper}
```
adespawn, line 95: ~While this is also partially blunder on my part~, this should return ``rust.MaybeUnsetQueryParameterWrapper?`` instead of ``rust.MaybeUnsetQueryParameterWrapper``

lib/types/cql-utils.js
```diff
@@ -81,23 +81,37 @@ function encodeMap(value, parentType) {
...
  */
 function getWrapped(type, value) {
-    if (value === null || value === types.unset) {
+    if (value === null) {
         return null;
+    } else if (value === types.unset) {
+        return rust.MaybeUnsetQueryParameterWrapper.unset();
```
adespawn, line 101: As @wprzytula mentioned, rust driver may not support CQL protocol with no unset values. Remember to check that, and add appropriate comment

lib/types/cql-utils.js
```diff
@@ -81,23 +81,37 @@ function encodeMap(value, parentType) {
...
     }
+    return rust.MaybeUnsetQueryParameterWrapper.fromType(
+        getWrappedType(type, value),
+    );
```
wprzytula, line 105: :thinking: Here, we have a mess about naming. `QueryParameterWrapper` is **not** a type; it's a value. Of course, the value has some type, too, but it's more than just a type. `CqlType` is a type; `CqlValue` is a value.

As the naming you chose has brought me the same confusion many times, I believe it will be the same for other future readers. Let's protect them.

```suggestion
    return rust.MaybeUnsetQueryParameterWrapper.fromNonNullNonUnsetValue(
        wrapValueBasedOnType(type, value),
    );
```

wprzytula (changes requested):

wprzytula (approved):

adespawn (approved):

PR #165, opened 2025-03-23
Batch options

adespawn:
Apply query options to batch statements.


src/session.rs
```diff
@@ -171,51 +171,58 @@ pub fn create_unprepared_batch(queries: Vec<String>) -> BatchWrapper {
...
-    // Currently there is no support for paging, so there is no need for this option
-    /* if let Some(o) = options.fetch_size {
-        if o.is_negative() {
-            return Err(js_error("fetch size cannot be negative"));
+macro_rules! apply_options {
+    (
+    $typ: ty,
+    $name: ident
+
+) => {
```
wprzytula, line 188: :pick:

wprzytula:
```suggestion
macro_rules! make_apply_options {
    ($statement_type: ty, $fn_name: ident) => {
```

src/session.rs
```diff
@@ -171,51 +171,58 @@ pub fn create_unprepared_batch(queries: Vec<String>) -> BatchWrapper {
...
+        fn $name(mut query: $typ, options: &QueryOptionsWrapper) -> napi::Result<$typ> {
```
wprzytula, line 189: :wrench:

wprzytula:
```suggestion
        fn $fn_name(mut statement: $statement_type, options: &QueryOptionsWrapper) -> napi::Result<$statement_type> {
```

wprzytula (changes requested):
What about unprepared statements? Do they have options applied?

adespawn:
> What about unprepared statements? Do they have options applied?

Now they have options applied

src/session.rs
```diff
@@ -154,21 +154,30 @@ impl SessionWrapper {
...
+) -> napi::Result<BatchWrapper> {
     let mut batch: Batch = Default::default();
     queries
         .into_iter()
         .for_each(|q| batch.append_statement(q.as_str()));
-    BatchWrapper { inner: batch }
+
+    batch = apply_batch_options(batch, options)?;
+    Ok(BatchWrapper { inner: batch })
 }
```
wprzytula: :pick: `queries` -> `statements`

src/session.rs
```diff
@@ -65,11 +65,12 @@ impl SessionWrapper {
     pub async fn query_unpaged_no_values(
         &self,
         query: String,
-        _options: &QueryOptionsWrapper,
+        options: &QueryOptionsWrapper,
     ) -> napi::Result<QueryResultWrapper> {
```
wprzytula, line 69: :question: What is the use case for `query_unpaged_no_values()`? It seems equivalent to `query_unpaged()`, being just a special trivial case of it.

test/integration/supported/client-batch-tests.js
```diff
@@ -415,61 +413,60 @@ describe("Client @SERVER_API", function () {
                     params: [id2, "value 2 with timestamp"],
                 },
             ];
+            let assert1 = function (next) {
+                client.execute(
```
wprzytula, line 417: :pick: `let` -> `const` ?

test/integration/supported/client-batch-tests.js
```diff
@@ -415,61 +413,60 @@ describe("Client @SERVER_API", function () {
...
+                        assert.strictEqual(
+                            result.first()["writetime(text_sample)"].toString(),
+                            timestamp.toString(),
+                        );
+                        next();
+                    },
+                );
+            };
+            let assert2 = function (next) {
+                client.execute(
```
wprzytula, line 440: ditto

wprzytula (changes requested):

src/session.rs
adespawn: https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/pull/129

wprzytula:
```rust
            } else if (!params) {
                // Optimize: If we have no parameters we can skip preparation step
                result = await this.rustClient.queryUnpagedNoValues(
                    query,
                    rustOptions,
                );
            } else {
                let expectedTypes = convertHints(execOptions.getHints() || []);
                let parsedParams = parseParams(expectedTypes, params, true);
                result = await this.rustClient.queryUnpaged(
                    query,
                    parsedParams,
                );
            }
```
:question:
1. Why `queryUnpagedNoValues` accepts `rustOptions` and `queryUnpaged` does not?
2. Why can't we just have one case (`queryUnpaged`) that will be executed on empty parameter list? This way we reduce code complexity, by handling the special trivial case in the more general branch.

adespawn: 1. It's fixed in this PR
2. If I remember correctly you suggested adding this special case to improve the performance

wprzytula (approved):

PR #174, opened 2025-04-06
Remove usages of client.shutdown

adespawn:
Removes (most?) usages of client.shutdown from examples and integration tests. This is a follow up to https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/commit/d9cbe8f3023332c3d298791c61db6ba9167af3f9

### DIFF
--- a/test/integration/broken/pool-simulator-tests.js
+++ b/test/integration/broken/pool-simulator-tests.js
@@ -1196,7 +1196,5 @@ function newInstance(simulacronCluster, options) {
         ),
     );
 
-    helper.shutdownAfterThisTest(client);
-
     return client;
 }
--- a/test/integration/broken/pool-simulator-tests.js
+++ b/test/integration/broken/pool-simulator-tests.js
@@ -240,8 +240,7 @@ describe("pool", function () {
                         isIdempotent: true,
                     }),
                 )
-                .then(() => assert.strictEqual(retryCount, 2))
-                .then(() => client.shutdown());
+                .then(() => assert.strictEqual(retryCount, 2));
         });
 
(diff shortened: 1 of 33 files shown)


adespawn:
Tokio runtime is fully managed by NAPI-RS so we (cannot / shouldn't) drop the runtime manually - meaning it's allocated for the whole duration NAPI-RS is imported.

This is not the case for the data allocated to the client - this is deleted as soon as the variable is no longer used and the garbage collector activates.

wprzytula:
> it's allocated for the whole duration NAPI-RS is imported.

What does it mean? Can the GC deallocate the tokio runtime at some point?

test/integration/broken/pool-simulator-tests.js
```diff
@@ -784,7 +770,8 @@ describe("pool", function () {
                 await promiseUtils.delay(20);
 
                 // Shutdown while there are queries in progress
-                await client.shutdown();
+                // Note: this will likely no longer work in our driver
+                // await client.shut down();
```
wprzytula, line 774: :wrench:

wprzytula: 1. `shut down` -> `shutdown`
2. Why `likely`? It's just not supported, let's be blunt clear about that.

wprzytula (changes requested):
:question: In person, I requested `shutdown()` documentation that is relevant to our implementation, but I can't see it written yet.

adespawn:
I've added documentation as a separate PR (#184)

wprzytula (approved):

PR #175, opened 2025-04-06
Remove some unused code

adespawn:
This PR removes most of the old files, that are not part of the exposed API - supported or not. 

This breaks some functionality that was working (probably not fully correct) with the old code


adespawn:
Merged as #180

PR #180, opened 2025-04-09
Remove some unused code v2

adespawn:
#175, but with different split into commits, and even more code removed.

Removes only code, that is not exposed in the API. May break some functionality that was working (probably not fully correct) with the old code.


lib/metadata/index.js
```diff
@@ -130,13 +129,13 @@ class Metadata {
...
     setPartitioner(partitionerName) {
-        if (/RandomPartitioner$/.test(partitionerName)) {
+        /* if (/RandomPartitioner$/.test(partitionerName)) {
             return (this.tokenizer = new t.RandomTokenizer());
         }
         if (/ByteOrderedPartitioner$/.test(partitionerName)) {
             return (this.tokenizer = new t.ByteOrderedTokenizer());
         }
-        return (this.tokenizer = new t.Murmur3Tokenizer());
+        return (this.tokenizer = new t.Murmur3Tokenizer()); */
```
wprzytula, line 136: :question: Why are you commenting this code out? What's the purpose? Shouldn't we rather delete the code altogether and provide a `deprecated` annotation with an explanation?

.github/workflows/integration-tests.yml
```diff
@@ -69,6 +69,4 @@ jobs:
         shell: bash
       - name: Run integration tests
         run: npm run integration
-      - name: Run integration not supported
-        run: npm run integration-not-supported
```
wprzytula, line 72: :question: Can you remind me what was the purpose of `not-supported` integration workflow?

lib/host.js
```diff
@@ -48,14 +46,14 @@ class Host extends events.EventEmitter {
          * @type {HostConnectionPool}
          */
         Object.defineProperty(this, "pool", {
-            value: new HostConnectionPool(this, protocolVersion),
+            value: [],
             enumerable: false,
         });
```
wprzytula, line 50: :question: ??? Why replacing an object with an empty array? How is it supposed to work?!
Why not remove the property altogether?

lib/host.js
```diff
@@ -48,14 +46,14 @@ class Host extends events.EventEmitter {
...
 
-        this.pool.on("open", (err) =>
+        /* this.pool.on("open", (err) =>
             promiseUtils.toBackground(this._onNewConnectionOpen(err)),
         );
-        this.pool.on("remove", () => this._checkPoolState());
+        this.pool.on("remove", () => this._checkPoolState()); */
```
wprzytula, line 55: :question: Why comment this out instead of deleting?

wprzytula, line 55: :question: Why comment this out instead of deleting?

test/test-helper.js
```diff
@@ -1033,8 +1033,6 @@ const helper = {
              * Shutdowns the hosts and invoke the optional callback.
              */
             shutdown: function (cb) {
-                hosts.forEach((h) => h.shutdown(false));
-
                 if (cb) {
                     cb();
                 }
```
wprzytula, line 957: :thinking: I though we do not support shutdown at all. The whole thing should therefore be deleted, shouldn't it?

lib/metadata/index.js
```diff
@@ -968,9 +909,9 @@ class Metadata {
...
                 throw new TypeError(
                     "User defined type not found: " +
-                        type.info.keyspace +
-                        "." +
-                        type.info.name,
+                    type.info.keyspace +
+                    "." +
+                    type.info.name,
                 );
             }
```
wprzytula, line 916: :question: Why is this in this commit? Shouldn't it have been formatted months ago, when the formatter was run against the whole codebase?

adespawn: Hmmm... It should have been - but for some unknown reasons it hasn't been

.github/workflows/integration-tests.yml
adespawn: There were some integration tests (or just a single one) that were working with DS code.

lib/host.js
adespawn: I cleaned up the host file even more. My initial intention for this PR was to just remove files that didn't expose anything to API, and then do a second iteration of cleanup - removing the old logic that was no longer usefull.

examples/DataStax/metadata/metadata-hosts.js.broekn
PiotrJunior, line 1: `[...].broekn`
Shoudn't it be `broken`

lib/client.js
```diff
@@ -57,24 +56,10 @@ class Client extends events.EventEmitter {
...
-        /**
-         * Gets the schema and cluster metadata information.
-         * @type {Metadata}
-         */
-        this.metadata = this.controlConnection.metadata;
-        /**
-         * Gets an associative array of cluster hosts.
-         * @type {HostMap}
-         */
-        this.hosts = this.controlConnection.hosts;
```
PiotrJunior, line 77: Can you explain, why are we removing metadata and hosts, despite that it is exposed by DataStax API https://docs.datastax.com/en/drivers/nodejs-dse/1.0/Client.html#hosts

PiotrJunior (changes requested):

test/test-helper.js
wprzytula: OK; what I mean is that if we removed the code that actually shuts hosts down (and its caller at 1036), we should remove the remaining logic 1035-1038+ too.

wprzytula: Don't you agree?

lib/host.js
wprzytula: This does not answer my question:
> How is it supposed to work?!
Array is not an object, and object is not an array. Then, replacing an object with an arrays looks like a bug.

adespawn: That was a placeholder code -- it wasn't supposed to work -- it wasn't just yet deleted.

lib/metadata/index.js
```diff
@@ -1025,29 +966,8 @@ class Metadata {
...
-                const value = row["schema_version"];
-                if (!value) {
-                    continue;
-                }
-                versions.add(value.toString());
-            }
-        }
-        return versions.size === 1;
+    async compareSchemaVersions() {
+        throw new Error(`TODO: Not implemented`);
```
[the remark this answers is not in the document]
adespawn: This is some internal function, that no longer works, but is not deleted at this moment

lib/host.js
wprzytula: OK. Just... I'd use an empty object as a placeholder for another object. But nvm.

wprzytula (approved):

PR #192, opened 2025-04-24
[WIP] rust concurrent

adespawn:


src/session.rs
```diff
@@ -237,6 +287,62 @@ impl SessionWrapper {
...
+            Result<scylla::response::query_result::QueryResult, scylla::errors::ExecutionError>,
+            usize,
+        )> = JoinSet::new();
+        let mut i = 0;
+        let mut in_flight = 0;
+
+        let mut res: Vec<Option<QueryResultWrapper>> = Vec::new();
+        while res.len() < queries.len() {
+            res.push(None);
+        }
```
wprzytula, line 312:
```suggestion
        res.resize(queries.len(), None);
```

wprzytula (changes requested):

src/session.rs
wprzytula: Ugh, really? If this is the reason you did it this way, then please write that in a comment.

PR #183, opened 2025-04-12
FromNapiValue trait for parameters

adespawn:
Before, we created QueryParameterWrapper with from_... functions, and then cloning this object when executing query.

Now, with implementation of FromNapiValue trait, we pass parameters to queries directly, without any additional conversion steps. In order to correctly handle value types, each parameter is passed as a pair (type, value). Information about type is then used on the rust side to create correct CqlValue object.

This significantly improves performance for concurrent inserts benchmark, and slightly for plain inserts benchmark.

Concurrent inset. Around 30% improvement in time elapsed, 20% in task clock

After:

```bash
sudo perf stat node ./benchmark/logic/concurrent_insert.js scylladb-javascript-driver 1000000

 Performance counter stats for 'node ./benchmark/logic/concurrent_insert.js scylladb-javascript-driver 1000000':

         70,343.46 msec task-clock                       #    1.892 CPUs utilized             
         2,909,393      context-switches                 #   41.360 K/sec                     
            40,946      cpu-migrations                   #  582.087 /sec                      
           190,074      page-faults                      #    2.702 K/sec                     
   273,230,048,329      cycles                           #    3.884 GHz                       
    72,133,512,748      stalled-cycles-frontend          #   26.40% frontend cycles idle      
   205,423,999,244      instructions                     #    0.75  insn per cycle            
                                                  #    0.35  stalled cycles per insn   
    39,462,664,735      branches                         #  561.000 M/sec                     
     1,481,648,921      branch-misses                    #    3.75% of all branches           

      37.183329662 seconds time elapsed

      46.102799000 seconds user
      25.690728000 seconds sys
```

Before

```bash
sudo perf stat node ./benchmark/logic/concurrent_insert.js scylladb-javascript-driver 1000000

 Performance counter stats for 'node ./benchmark/logic/concurrent_insert.js scylladb-javascript-driver 1000000':

         86,873.69 msec task-clock                       #    1.633 CPUs utilized             
         3,221,740      context-switches                 #   37.085 K/sec                     
            48,446      cpu-migrations                   #  557.660 /sec                      
           190,596      page-faults                      #    2.194 K/sec                     
   344,373,066,097      cycles                           #    3.964 GHz                       
    87,340,988,626      stalled-cycles-frontend          #   25.36% frontend cycles idle      
   246,727,230,093      instructions                     #    0.72  insn per cycle            
                                                  #    0.35  stalled cycles per insn   
    47,143,750,188      branches                         #  542.670 M/sec                     
[...]

PR #197, opened 2025-04-29
Decimal

PiotrJunior:


src/result.rs
```diff
@@ -299,6 +299,20 @@ impl CqlValueWrapper {
         }
     }
 
+    #[napi]
+    pub fn get_decimal(&self) -> napi::Result<Buffer> {
+        match &self.inner {
+            CqlValue::Decimal(r) => {
+                let dec = r.as_signed_be_bytes_slice_and_exponent();
```
adespawn, line 306:
```suggestion
                let (value, len) = r.as_signed_be_bytes_slice_and_exponent();
```

src/result.rs
```diff
@@ -299,6 +299,20 @@ impl CqlValueWrapper {
         }
     }
 
+    #[napi]
+    pub fn get_decimal(&self) -> napi::Result<Buffer> {
+        match &self.inner {
```
adespawn, line 304: Add a description, of how the buffer is filled

lib/types/cql-utils.js
```diff
@@ -161,6 +163,22 @@ function wrapValueBasedOnType(type, value) {
...
+
+            if (value instanceof BigDecimal) {
+                return rust.QueryParameterWrapper.fromDecimal(
+                    value._scale,
+                    Integer.toBuffer(value._intVal),
+                );
+            }
+            throw new TypeError(
+                "Expected BigDecimal type to parse into Cql Decimal",
+            );
```
adespawn, line 181: I would prefer this logic reversed

adespawn: (Throw an error in if condition)

test/unit/query-parameter-tests.js
```diff
@@ -19,7 +20,12 @@ const testCases = [
     ["Boolean", false],
     ["Blob", utils.allocBufferFromArray([0, 1, 2, 3])],
     ["Counter", 921],
-    ["Double", 21.37],
+    [
+        "Decimal",
+        BigDecimal.fromString(
+            "2137.213721372137213721372137213721372137213721372137213721372137213721372",
```
adespawn, line 26: 😂

lib/types/big-decimal.js
```diff
@@ -3,264 +3,272 @@ const Integer = require("./integer");
...
-     * @type {Integer}
-     * @private
+     * Constructs an immutable arbitrary-precision signed decimal number.
+     * A `BigDecimal` consists of an [arbitrary precision integer]{@link module:types~Integer}
+     * *unscaled value* and a 32-bit integer *scale*.  If zero
+     * or positive, the scale is the number of digits to the right of the
+     * decimal point.  If negative, the unscaled value of the number is
+     * multiplied by ten to the power of the negation of the scale.  The
+     * value of the number represented by the `BigDecimal` is
+     * therefore <tt>(unscaledValue &times; 10<sup>-scale</sup>)</tt>.
```
adespawn, line 22: Is it possible to remove those html tags?

lib/types/big-decimal.js
```diff
@@ -3,264 +3,272 @@ const Integer = require("./integer");
...
-    if (typeof unscaledValue === "number") {
-        unscaledValue = Integer.fromNumber(unscaledValue);
-    }
+class BigDecimal {
     /**
-     * @type {Integer}
-     * @private
+     * Constructs an immutable arbitrary-precision signed decimal number.
+     * A `BigDecimal` consists of an [arbitrary precision integer]{@link module:types~Integer}
+     * *unscaled value* and a 32-bit integer *scale*.  If zero
```
adespawn, line 17:
```suggestion
     * *unscaled value* and a 32-bit integer *scale*. If zero
```

lib/types/big-decimal.js
```diff
@@ -3,264 +3,272 @@ const Integer = require("./integer");
...
+     * or positive, the scale is the number of digits to the right of the
+     * decimal point.  If negative, the unscaled value of the number is
```
adespawn, line 19:
```suggestion
     * decimal point. If negative, the unscaled value of the number is
```

lib/types/big-decimal.js
```diff
@@ -3,264 +3,272 @@ const Integer = require("./integer");
...
+     * multiplied by ten to the power of the negation of the scale.  The
```
adespawn, line 20:
```suggestion
     * multiplied by ten to the power of the negation of the scale. The
```

adespawn (changes requested):

lib/types/cql-utils.js
PiotrJunior: I think its more readable then `if (!(value instanceof BigDecimal))`

lib/types/big-decimal.js
PiotrJunior: I don;t think so. Maybe @Chlor87 knows the answer

adespawn:
image

lib/types/cql-utils.js
wprzytula: > I think its more readable then `if (!(value instanceof BigDecimal))`

I disagree. It's a common way to write code in a way that there's a series of error checks, finalized by the success case on the happy path:
```
if bad_one {
  return error1;
}
if bad_two {
  return error2;
}
return good;
```

adespawn:
As previously discussed, please re-base onto #183

lib/types/big-decimal.js
```diff
@@ -3,264 +3,279 @@ const Integer = require("./integer");
...
 
-/**
- * Returns a Number representation of this `BigDecimal`.
- * @returns {Number}
- */
-BigDecimal.prototype.toNumber = function () {
-    return parseFloat(this.toString());
-};
+    /**
+     * Returns a Number representation of this <code>BigDecimal</code>.
```
adespawn, line 265: Why did you add a code tag here?

adespawn (changes requested):

adespawn (commented):
Few things about comments, otherwise LGTM

src/result.rs
```diff
@@ -254,7 +254,20 @@ impl ToNapiValue for CqlValueWrapper {
...
                 CqlType::Counter,
             ),
-            CqlValue::Decimal(_) => todo!(),
+            CqlValue::Decimal(val) => {
+                // Returned buffer is in format XXXXYYYY...YYY
+                // where XXXX is the exponent of the decimal in big endian
+                // and YYYY...YYY is the value of the decimal in big endian
+                let (value, len) = val.as_signed_be_bytes_slice_and_exponent();
+                let mut buf = vec![0u8; 4 + value.len()];
+                buf[0..4].copy_from_slice(&len.to_be_bytes()); // Big endian notation is used in CQL protocol
```
adespawn, line 263: CQL protocol has nothing to do with this. When big-decimal no longer interfaces with cql protocol, but with values provided by us, this comment does not provide useful information and it doesn't satisfy my previous comment

src/result.rs
```diff
@@ -254,7 +254,20 @@ impl ToNapiValue for CqlValueWrapper {
                 BigInt::to_napi_value(env, val.0.into()),
                 CqlType::Counter,
             ),
-            CqlValue::Decimal(_) => todo!(),
+            CqlValue::Decimal(val) => {
+                // Returned buffer is in format XXXXYYYY...YYY
+                // where XXXX is the exponent of the decimal in big endian
+                // and YYYY...YYY is the value of the decimal in big endian
+                let (value, len) = val.as_signed_be_bytes_slice_and_exponent();
```
adespawn, line 261: We can have a single comment for both places where we use be here

lib/types/cql-utils.js
```diff
@@ -185,6 +185,19 @@ function wrapValueBasedOnType(type, value) {
...
+            } else if (typeof value === "string") {
+                value = BigDecimal.fromString(value);
+            }
+            if (!(value instanceof BigDecimal)) {
+                throw new TypeError(
+                    "Expected BigDecimal type to parse into Cql Decimal",
+                );
+            }
+            value = BigDecimal.toBuffer(value);
+            break;
```
adespawn, line 200: Why change ordering in switch case?

src/requests/parameter_wrappers.rs
```diff
@@ -106,7 +106,13 @@ fn cql_value_from_napi_value(typ: &ComplexType, elem: &Array, pos: u32) -> napi:
             get_element!(BigInt),
             "Value cast into counter type shouldn't overflow i64",
         )?)),
-        CqlType::Decimal => todo!(),
+        CqlType::Decimal => {
+            let buf = get_element!(&[u8]);
+            CqlValue::Decimal(CqlDecimal::from_signed_be_bytes_slice_and_exponent(
+                &buf[4..],
+                i32::from_be_bytes(buf[0..4].try_into().unwrap()), // Buffer is guaranteed to be at least 4 bytes by CQL protocol
```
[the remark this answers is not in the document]
adespawn: Again, CQL protocol has nothing to do with our internal representation

src/tests/result_tests.rs
```diff
@@ -42,6 +44,39 @@ pub fn tests_get_cql_wrapper_counter() -> CqlValueWrapper {
...
+#[napi]
+/// Test function returning sample CqlValueWrapper with Decimal type with negative exponent
+pub fn tests_get_cql_wrapper_decimal_negative_exponent() -> CqlValueWrapper {
+    let element = CqlValue::Decimal(CqlDecimal::from_signed_be_bytes_slice_and_exponent(
+        &[69],
+        -10,
+    ));
+    CqlValueWrapper { inner: element }
+}
+
```
adespawn, line 79: Nit: we can compress this to a single function (see: ``request_values_tests``): let's not make the rust API overcrowded

test/unit/big-decimal-tests.js
adespawn, line 1: Enable Decimal tests -> Enable Decimal unit tests

src/result.rs
PiotrJunior: But the Buffer moved between Rust part and JS part is in the same format as in the CQL protocol

src/tests/result_tests.rs
PiotrJunior: I think then all other test can be compressed like in `request_values_tests`

src/result.rs
```diff
@@ -244,7 +244,21 @@ impl ToNapiValue for CqlValueWrapper {
...
-            CqlValue::Decimal(_) => todo!(),
+            CqlValue::Decimal(val) => {
+                // JS driver expects decimal to be in the format of Decimal from CQL protocol.
+                // Returned buffer is in format XXXXYYYY...YYY
+                // where XXXX is the exponent of the decimal in big endian
+                // and YYYY...YYY is the value of the decimal in big endian.
+                let (value, len) = val.as_signed_be_bytes_slice_and_exponent();
+                let mut buf = vec![0u8; 4 + value.len()];
+                buf[0..4].copy_from_slice(&len.to_be_bytes());
+                buf[4..].copy_from_slice(value);
```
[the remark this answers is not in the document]
PiotrJunior: I think it is better to have 4 as a constant then using `std::mem::size_of::<T>()` because in case of changing internal logic of CqlValue::Decimal, this buffer won't be in CQL protocol format

PiotrJunior:
@ZuzaOsa  It looks like cargo clippy was updated and many files not regarding my PR need to be changed.

src/result.rs
```diff
@@ -218,7 +218,22 @@ impl ToNapiValue for CqlValueWrapper {
                 BigInt::to_napi_value(env, val.0.into()),
                 CqlType::Counter,
             ),
-            CqlValue::Decimal(_) => todo!(),
+            CqlValue::Decimal(val) => {
+                let exp_size = 4;
```
wprzytula, line 222: :pick: Why `let`? It's a constant.

wprzytula: This should be made `const EXP_SIZE: usize = 4`.

lib/types/cql-utils.js
wprzytula, line 200: :wrench: Please include related DataStax JS Driver's code that handles decimal input. The goal is to have reference for what types are accepted and what requirements are imposed on them.

The same should be done for other types, I think, but it's ofc out of scope of this PR.

src/requests/parameter_wrappers.rs
```diff
@@ -106,7 +106,19 @@ fn cql_value_from_napi_value(typ: &ComplexType, elem: &Array, pos: u32) -> napi:
             get_element!(BigInt),
             "Value cast into counter type shouldn't overflow i64",
         )?)),
-        CqlType::Decimal => todo!(),
+        CqlType::Decimal => {
+            let exp_size = 4;
```
wprzytula, line 110: :pick: Ditto: please use `const`.

src/requests/parameter_wrappers.rs
```diff
@@ -106,7 +106,19 @@ fn cql_value_from_napi_value(typ: &ComplexType, elem: &Array, pos: u32) -> napi:
...
+            let buf = get_element!(&[u8]);
+            // JS driver returns decimal in the same format as Decimal from CQL protocol.
+            // Returned buffer is in format XXXXYYYY...YYY
+            // where XXXX (4 bytes) is the exponent of the decimal in big endian
+            // and YYYY...YYY is the value of the decimal in big endian.
+            CqlValue::Decimal(CqlDecimal::from_signed_be_bytes_slice_and_exponent(
+                &buf[exp_size..],
+                i32::from_be_bytes(buf[0..exp_size].try_into().unwrap()), // Buffer is always guaranteed to be
+                                                                          // at least 4 bytes by JS part of the driver (it always returns at least 4 bytes).
```
wprzytula, line 119: :pick: Please put the comment in the line above. The way it's now it's not readable.

wprzytula (changes requested):

src/result.rs
PiotrJunior: I’m not sure I fully understand what you’d like me to do here.

lib/types/cql-utils.js
PiotrJunior: I’m not sure I fully understand what you’d like me to do here.

wprzytula: I don't understand myself either. Let's forget about this one.

adespawn (approved):
LGTM!

wprzytula (approved):

PR #207, opened 2025-05-14
Introduce repeat capped

adespawn:
Limit a single execute concurrent t q queries (q = 10^6),
in order to avoid "JavaScript heap out of memory" problem


wprzytula (approved):

benchmark/logic/utils.js
```diff
@@ -1,10 +1,23 @@
...
 
+// call callback, each with up to singleStepCount,
+// multiple times, so that sum of all called callback is equal to n
+async function repeatCapped(callback, n) {
+    for (let rep = 0; rep * singleStepCount < n; rep++) {
+        const finalStep = Math.min(n, (rep + 1) * singleStepCount);
+        await callback(finalStep - rep * singleStepCount);
+    }
+
+}
```
wprzytula, line 20: Please put the explanation what this is introduced for.

PR #211, opened 2025-05-17
Remove datacenter from benchmarks, examples and docs

adespawn:
We currently do not support this field -- meaning it had no effect in those situations.


adespawn:
This only removes the usage examples of unimplemented features

wprzytula (approved):

PR #218, opened 2025-05-25
Batch benchmark

adespawn:


benchmark/logic/batch.js
Stapox35, line 1: This benchmark does not end for `cassandra-driver`.

Stapox35 (changes requested):

wprzytula (approved):

PR #222, opened 2025-05-27
Generate UUID in rust

adespawn:
While this appears to be slightly faster, more testing is required, as changes are minimal


wprzytula (dismissed):

adespawn:
> > looking back at this code I'm not fully sure if the call to rust will never throw an error
> 
> What are your doubts? What pieces of Rust code do you suspect might throw an error?

There may be some edgecase where NAPI-RS could thrown an error

lib/types/uuid.js
```diff
@@ -72,20 +72,14 @@ class Uuid {
...
-            });
-        } else {
-            const buffer = getRandomBytes();
-            return createUuidFromBuffer(buffer);
+            try {
+                return callback(null, new Uuid(rust.getRandomUuidV4()));
+            } catch (err) {
+                return callback(err);
+            }
         }
```
wprzytula, line 81: :wrench: Please write a comment that you are not sure what error could be thrown here, and you add it just in case.

wprzytula (changes requested):

PR #224, opened 2025-05-28
UUID generateRandom

adespawn:
Extend the UUID API to add UUID.generateRandom static value.

When providing `generateRandom` as a query parameter,
each time a new random UUID will be generated.

While using `generateRandom` is faster than `random`,
it's not possible to retrieve a value of such UUID,
or use the same UUID in multiple places.

### DIFF
--- a/benchmark/logic/concurrent_insert.js
+++ b/benchmark/logic/concurrent_insert.js
@@ -8,6 +8,7 @@ const assert = require("assert");
 
 const client = new cassandra.Client(utils.getClientArgs());
 const iterCnt = parseInt(process.argv[3]);
+const uuidGenerator = utils.getRandomUUID(process.argv[2], cassandra.types.Uuid);
 
 async.series(
     [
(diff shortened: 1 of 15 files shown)


wprzytula:
> Extend the UUID API to add UUID.generateRandom static value.
> 
> When providing `generateRandom` as a query parameter, each time a new random UUID will be generated.
> 
> While using `generateRandom` is faster than `random`, it's not possible to retrieve a value of such UUID, or use the same UUID in multiple places.

Having read this, I don't undestand:
- what is the motivation of this,
- what do we want to achieve by this.

Too little context in general.

adespawn:
> what is the motivation of this,
> what do we want to achieve by this.

We want to reduce the time spent on generating UUIDs and as a result speedup the driver

wprzytula:
> > what is the motivation of this,
> > what do we want to achieve by this.
> 
> We want to reduce the time spent on generating UUIDs and as a result speedup the driver

What you achieve is not a speedup _of the driver_. This is a speedup of _an application that uses the driver_, because it's the application's responsibility to provide ready statement parameters.

Your PR shifts this responsibility to the driver, without particular reason. The only argument I can see for doing that is to make benchmarks run faster - but it's about the benchmarks, not the driver itself.

For now, I reject this PR as a whole.

PR #225, opened 2025-05-29
UUID result

adespawn:
Update UUID, so that we return raw bytes instead of object wrapper.

Removes CqlTypeClass, to reduce the overhead from it's type.
This means, that array will be treated as ambiguous types in result wrapping.


wprzytula:
> This means, that array will be treated as ambiguous types in result wrapping.

1. What array?
2. What does it mean that it will be treated as ambiguous? Can you provide an example?

wprzytula (approved):

Issue #148, opened 2025-03-04
Flaeky tests


adespawn:
List of known flake tests executions:
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/actions/runs/13652709566/attempts/1?pr=144
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/actions/runs/13640799157/job/38130001125
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/actions/runs/12236679366/job/34130672494?pr=120
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/actions/runs/12225798378/job/34100215931?pr=95
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/actions/runs/12905363429/job/35984375578?pr=56
https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/actions/runs/11702787093/job/32591689078?pr=43

wprzytula:
Could we please have a listing of flaky tests (their names) here? @adespawn

adespawn:
That is the funny part. From what I can see, there is no single flaky test. There is just a combination of tests that are flaky (see Discord)

adespawn:
At this point in time, we have found out the reason for all memory issues. The only thing that remains is (long unseen) ccm timeouts

wprzytula:
What's the current status of this?

PR #243, opened 2025-10-02
[WIP] Introduce temporary logging

adespawn:
Introduce temporary logging in integration tests, around the before step. The goal of this commit is to narrow down the cause of occasional timeouts in "before" step on CI (those timeouts mentioned in #148).


adespawn:
This PR does not solve the problem, but changes tested here lead to this bug: #244

PR #291, opened 2025-10-23
Update napi rs

adespawn:
This PR updates NapiRs to V3, introduces some changes that came up from updating the library and does some small cleaning of the code

### DIFF
--- a/src/metadata/host.rs
+++ b/src/metadata/host.rs
@@ -1,14 +1,15 @@
 use napi::bindgen_prelude::Buffer;
 
-use crate::session::SessionWrapper;
+use crate::{session::SessionWrapper, utils::to_napi_obj::define_rust_to_js_convertible_object};
 
-#[napi]
-pub struct HostWrapper {
-    pub host_id: Buffer,
-    pub address: String,
-    pub datacenter: Option<String>,
-    pub rack: Option<String>,
-}
+define_rust_to_js_convertible_object!(
+    HostWrapper {
+        host_id, hostId: Buffer,
+        address, address: String,
+        datacenter, datacenter: Option<String>,
+        rack, rack: Option<String>
+    }
+);
 
 #[napi]
 impl SessionWrapper {


test/unit/type-guessing-tests.js
```diff
@@ -54,7 +53,7 @@ describe("Encoder.guessDataType()", function () {
         );
         assertGuessed(
             types.TimeUuid.now(),
-            rust.CqlType.Timeuuid,
+            rust.CqlType.Uuid,
             "Guess type for a TimeUuid value failed",
         );
```
wprzytula: Justify this change.

src/tools/from_napi_obj.rs
```diff
@@ -0,0 +1,70 @@
...
+///     {some_field, bool},
+///     {other_field, i32});
+/// ```
+///
+/// Will create the following struct:
+/// ```rust
+/// struct Example{
+///     some_field: Option<bool>,
+///     other_field: Option<i32>,
+/// }
```
wprzytula: :pick:

src/tools/from_napi_obj.rs
```diff
@@ -0,0 +1,70 @@
...
+/// ```
+///
+/// That can be passed from JS with the following Object:
```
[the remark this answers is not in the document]
wprzytula:
```suggestion
/// Which can be passed from JS with the following Object:
```

src/tools/from_napi_obj.rs
```diff
@@ -0,0 +1,70 @@
...
+/// ```js
+/// {someField: false}
+/// ```
+///
+/// and will be converted to:
+/// ```rust
+/// Example {some_field: Some(false), other_field: None}
```
wprzytula: :question: Will `{someField: null}` be equivalent to `{}` on the Rust side? I.e., is passing a null value for a field equivalent for not passing the field at all?

src/tools/from_napi_obj.rs
```diff
@@ -0,0 +1,70 @@
...
+/// ```
+#[macro_export]
```
wprzytula: :thought_balloon: Do we really have to `#[macro_export]`? If the defining crate is the only consumer, we can export it by re-exporting it **under** its definition: `pub(crate) use create_serializable_object;`.

lib/query-options.js
```diff
@@ -1,6 +1,7 @@
 const Long = require("long");
 const _execOptions = require("./execution-options");
 const { longToBigint } = require("./new-utils");
+const rust = require("../index");
```
wprzytula: :question: in commit message `Options wrapping (caching)`: `across the langues` ? Do you mean `across the languages`?

wprzytula (changes requested):

src/tools/from_napi_obj.rs
adespawn: No, I just didn't know we can do that

src/lib.rs
```diff
@@ -10,5 +10,6 @@ pub mod requests;
 pub mod result;
 pub mod session;
 pub mod tests;
+pub mod tools;
 pub mod types;
 pub mod utils;
```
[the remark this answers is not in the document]
adespawn: Missed that

test/unit/type-guessing-tests.js
adespawn: This was a deviation from the DSx unit test, that was not marked appropriately. Now it's back to the DSx version

src/requests/request.rs
```diff
@@ -1,37 +1,41 @@
...
+    {keyspace, String},
+    {logged, bool},
+    {prepare, bool},
+    {read_timeout, i32},
+    {routing_indexes, Vec<i32>},
+    {routing_names, Vec<String>},
+    {serial_consistency, i16},
+    {timestamp, BigInt},
+    {trace_query, bool}
+);
```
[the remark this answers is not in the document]
wprzytula: Why not? We're in charge of specifying the syntactic parameters of the macro.

src/tools/from_napi_obj.rs
adespawn: Yes

src/tools/from_napi_obj.rs
```diff
@@ -0,0 +1,70 @@
...
+macro_rules! create_serializable_object {
+    ($struct_name: ident, $({$field_name:ident, $field_type:ty}),*) => {
```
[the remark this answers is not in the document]
wprzytula: If for some reason I wanted to create a serializable objects with 0 fields, then I'd have to write the trailing comma as I did in the example above. This is not very interesting case, though, because I don't imagine anyone wanting 0-fields serializable objects :P

package.json
```diff
@@ -6,24 +6,22 @@
...
-        "armv7-unknown-linux-musleabihf",
-        "x86_64-unknown-linux-musl",
-        "x86_64-unknown-freebsd",
-        "i686-pc-windows-msvc",
-        "armv7-linux-androideabi",
-        "universal-apple-darwin",
-        "riscv64gc-unknown-linux-gnu"
-      ]
-    }
+    "binary-name": "scylladb-javascript-driver",
```
wprzytula: 🌱 We should probably rename `scylladb-javascript-driver` to `scylladb-nodejs-rs-driver`.

wprzytula (changes requested):

wprzytula (approved):

PR #297, opened 2025-11-04
Address translation policy

adespawn:
This PR adds support for pre-defined address translation policies. (Fixes: #299)

Custom policies will be added in a separate PR.

~~Unit tests will be added in this PR, once https://github.com/scylladb/scylla-rust-driver/pull/1466 is merged and the driver is updated to the release containing that commit~~

DataStax driver did not have any integration tests for this policy

### DIFF
--- a/src/policies/address_translation.rs
+++ b/src/policies/address_translation.rs
@@ -66,3 +66,36 @@ impl AddressTranslator for EC2MultiRegionTranslator {
         }
     }
 }
+
+#[cfg(test)]
+use std::str::FromStr;
+
+#[tokio::test]
+async fn return_same_address_when_not_resolved() {
+    let policy = EC2MultiRegionTranslator {};
+    let addr = SocketAddr::from_str("192.0.2.1:9042").unwrap();
+    let peer = UntranslatedPeer::new(addr);
+    let res: Result<SocketAddr, TranslationError> = policy.translate_address(&peer).await;
+    assert_eq!(res.expect("Expected to obtain the address"), addr);
+}
+
+#[tokio::test]
+async fn return_the_other_ip_associated_with_the_domain() {
+    let policy = EC2MultiRegionTranslator {};
+    let addr = SocketAddr::from_str("1.1.1.1:9042").unwrap();
+    let addr2 = SocketAddr::from_str("1.0.0.1:9042").unwrap();
+    let peer = UntranslatedPeer::new(addr);
+    for _ in 0..100 {
+        let res: SocketAddr = policy
+            .translate_address(&peer)
+            .await
+            .expect("Expected to obtain the address");
+        // Success: we got different address than provided.
+        if res == addr2 {
+            return;
+        }
+    }
+    unreachable!(
+        "Unlikely to happen: with each repetition we have 50:50 chance to get the other address"
+    );
+}
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,3 +3,7 @@ linker = "aarch64-linux-musl-gcc"
 rustflags = ["-C", "target-feature=-crt-static"]
 [target.x86_64-pc-windows-msvc]
 rustflags = ["-C", "target-feature=+crt-static"]
+
+[build]
+# To enable NodeJS RS Driver only features from Rust driver.
+rustflags = ["--cfg", "nodejs_rs_unstable"]
--- a/test/unit/address-resolution-tests.js
+++ b/test/unit/address-resolution-tests.js
@@ -5,6 +5,8 @@ const dns = require("dns");
 const addressResolution = require("../../lib/policies/address-resolution");
 const EC2MultiRegionTranslator = addressResolution.EC2MultiRegionTranslator;
 
+// Those unit tests check the JS side version of the logic.
+// Rust version of this policy is tested on the rust side with cargo test
 describe("EC2MultiRegionTranslator", function () {
     this.timeout(10000);
     describe("#translate()", function () {


lib/client-options.js
```diff
@@ -669,6 +669,18 @@ function setRustOptions(options) {
     rustOptions.applicationVersion = options.applicationVersion;
     rustOptions.keyspace = options.keyspace;
     rustOptions.cacheSize = options.maxPrepared;
+    // We can either:
+    // - do not have policies,
```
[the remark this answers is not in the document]
wprzytula:
```suggestion
    // We can either:
    // - not have policies,
```

adespawn:
And the Napi-rs funny approach (that we encountered at some point last academic year) strikes again.  Here we need to  nest AddressTranslatorPolicies in two Options, to properly convert it to Rust value

lib/policies/address-resolution.js
```diff
@@ -39,6 +40,18 @@ class AddressTranslator {
     translate(address, port, callback) {
         callback(address + ":" + port);
     }
+
+    /**
+     * This method is used to determine if the provided policy is one of the policies with the default implementation.
+     * This is meant to be for internal use only - overriding it by uses may lead to UB.
```
wprzytula: :wrench:

lib/client-options.js
```diff
@@ -665,6 +666,18 @@ function setRustOptions(options) {
...
+    // - have policies, but not have address resolution policy,
+    // - have a address resolution policy, that works as a policy, but is not inherited from the default policy (duck typing),
+    // - have a proper address resolution policy.
+    // In the first two cases, we treat that as lack of policy, and the third case as custom policy.
+    rustOptions.addressTranslationPolicy =
+        options.policies && options.policies.addressResolution
+            ? options.policies.addressResolution.isDefaultPolicy
+                ? options.policies.addressResolution.isDefaultPolicy()
+                : rust.AddressTranslatorPolicies.CustomPolicy
+            : null;
```
wprzytula: :thinking: This looks like `isDefaultPolicy()` should rather be `makeDefaultPolicy()`, no?
`is*` suggests that the function should return boolean.

src/utils/from_napi_obj.rs
```diff
@@ -40,7 +40,6 @@
 /// ```
 macro_rules! define_js_to_rust_convertible_object {
     ($struct_name: ident{$($field_name:ident, $js_name:ident: $field_type:ty),*}) => {
-        #[derive(Debug)]
         pub struct $struct_name {
```
wprzytula: :question: Why? Why not in a dedicated commit together with an explanation?

.github/workflows/unit-tests.yml
```diff
@@ -30,3 +30,5 @@ jobs:
         run: npm run unit
       - name: Run not fully integrated unit tests
         run: npm run unit-not-supported
+      - name: Run rust unit tests
+        run: cargo test
```
wprzytula: :thought_balloon:  Perhaps, for better debug experience, we should enable `RUST_BACKTRACE=full`?

wprzytula: I can see we have the following in our workflow yamls in Rust Driver's CI:

```yaml
env:
  CARGO_TERM_COLOR: always
  RUST_BACKTRACE: full
```

src/policies/address_translation.rs
```diff
@@ -66,3 +66,32 @@ impl AddressTranslator for EC2MultiRegionTranslator {
...
+    let policy = EC2MultiRegionTranslator {};
+    let addr = <SocketAddr as std::str::FromStr>::from_str("1.1.1.1:9042").unwrap();
+    let addr2 = <SocketAddr as std::str::FromStr>::from_str("1.0.0.1:9042").unwrap();
+    let peer = UntranslatedPeer::new(addr);
+    for _ in 0..100 {
+        let res: SocketAddr = policy
+            .translate_address(&peer)
+            .await
+            .expect("Expected to obtain the address");
+        // Success, we got other address, then inputted
```
wprzytula: :wrench:

wprzytula:
```suggestion
        // Success: we got different address than provided.
```

src/policies/address_translation.rs
```diff
@@ -0,0 +1,62 @@
+use std::net::{IpAddr, SocketAddr, SocketAddrV4, SocketAddrV6};
+
+use async_trait::async_trait;
+
+use scylla::{
+    errors::TranslationError,
+    policies::address_translator::{AddressTranslator, UntranslatedPeer},
+};
+
+pub struct EC2MultiRegionTranslator {}
```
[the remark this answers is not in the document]
adespawn: As this struct is internal only, I've added "redirection" to the public part of the docs, instead of duplicating the comments

src/policies/address_translation.rs
```diff
@@ -0,0 +1,62 @@
...
+        &self,
+        untranslated_peer: &UntranslatedPeer,
+    ) -> Result<SocketAddr, TranslationError> {
+        let translation = async move || {
+            let reverse_lookup =
+                dns_lookup::lookup_addr(&untranslated_peer.untranslated_address().ip())
+                    .map_err(|e| TranslationError::IoError(e.into()))?;
+            let looked_up = dns_lookup::lookup_host(&reverse_lookup)
+                .map_err(|e| TranslationError::IoError(e.into()))?
+                .nth(1);
```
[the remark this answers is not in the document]
adespawn: Off by one

adespawn:
Addressed most of the comments

adespawn:
I'm not sure why mdbook now complains about `multilingual = false`

lib/client-options.js
wprzytula: Once again. This is definition of `isDefaultPolicy`. It does not return a boolean, so it's not a predicate. I don't understand why it should be called `is*` then.
```js
isDefaultPolicy() {
        if (
            Object.getPrototypeOf(this) === EC2MultiRegionTranslator.prototype
        ) {
            return rust.AddressTranslatorPolicies.EC2MultiRegion;
        }
        return rust.AddressTranslatorPolicies.CustomPolicy;
    }
```

adespawn:
Rebase

PR #307, opened 2025-11-05
Add basic ssl support

adespawn:
Adds support for using ssl, when connecting to the database. For now, the driver does not parse any of the ssl options, other than rejectUnauthorized, that is used in the integration tests.

### DIFF
--- a/test/integration/supported/ssl-tests.js
+++ b/test/integration/supported/ssl-tests.js
@@ -8,12 +8,14 @@ const errors = require("../../../lib/errors");
 const utils = require("../../../lib/utils");
 const types = require("../../../lib/types");
 
-// Test disabled due to ssl error
-// INVESTIGATE(@wprzytula)
-// https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/actions/runs/11703077607/job/32592642939#step:12:741
-/* describe("Client @SERVER_API", function () {
+describe("Client @SERVER_API", function () {
     this.timeout(60000);
     context("with ssl enabled", function () {
+        // CCM does not support Scylla with ssl
+        // Would require merging of https://github.com/scylladb/scylla-ccm/pull/682
+        if (helper.getServerInfo().isScylla) {
+            return;
+        }
         const keyspace = helper.getRandomName("ks");
         const table = keyspace + "." + helper.getRandomName("table");
         const setupQueries = [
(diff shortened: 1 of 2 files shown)


adespawn:
Hmm... It appears, that ccm fails to setup scylla correctly with ssl, while it has no problem setting up cassandra db with ssl.
 Probably this needs to be reported to CCM repo...

adespawn:
@wprzytula I can either add support for more ssl options (and appropriate test) either in this PR or in a separate one. 
In case it's okay with you to do that in the next PR, this one is redy for review

src/session.rs
```diff
@@ -17,14 +18,24 @@ use crate::{requests::request::PreparedStatementWrapper, result::QueryResultWrap
 
 const DEFAULT_CACHE_SIZE: u32 = 512;
 
+// For now, ssl options include only rejectUnauthorized.
+// In practice, user can provide more options to configure
+// the ssl connection (see: ConnectionOptions typescript class)
+// This specific option is added, as it's used in the existing integration tests
+define_js_to_rust_convertible_object!(SslOptions {
+    reject_unauthorized,
+    rejectUnauthorized: bool
```
wprzytula: :recycle: Let's keep the convention of having specification of a single field in a single line.

src/session.rs
```diff
@@ -240,7 +251,20 @@ fn configure_session_builder(options: &SessionOptions) -> SessionBuilder {
...
+
+    if let Some(ssl_options) = &options.ssl_options {
+        let mut ssl_context_builder =
+            SslContextBuilder::new(SslMethod::tls()).map_err(err_to_napi)?;
+
+        ssl_context_builder.set_verify(if ssl_options.reject_unauthorized == Some(false) {
+            SslVerifyMode::NONE
+        } else {
+            SslVerifyMode::PEER
+        });
```
wprzytula: :pick: I'd use a `match` here, for clarity.

wprzytula (changes requested):

src/session.rs
adespawn: I've tried to do so, but cargo fmt had other plans

test/integration/supported/ssl-tests.js
```diff
@@ -8,12 +8,14 @@ const errors = require("../../../lib/errors");
...
-// https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/actions/runs/11703077607/job/32592642939#step:12:741
-/* describe("Client @SERVER_API", function () {
+describe("Client @SERVER_API", function () {
     this.timeout(60000);
     context("with ssl enabled", function () {
+        // CCM does not support Scylla with ssl
+        // Would require merging of https://github.com/scylladb/scylla-ccm/pull/682
+        if (helper.getServerInfo().isScylla) {
+            return;
+        }
```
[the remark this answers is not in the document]
adespawn: I mentioned the PR that fixes it in the comment above

Issue #66, opened 2024-11-06
Make rust parts of unit tests as a feature 


adespawn:
Or use other way to allow disabling those parts in deployment compilation

PR #313, opened 2025-11-07
New error handling

### LINKED ISSUE #119: Add support for throwing errors with special parameters from napi
Some custom errors require special parameters when being thrown. Currently error handling does not support that.

### LINKED ISSUE #216: Incorrect usage of the error throwing wrapper
In some places -  for example batch statements - the error wrapper is used only on the Rust side, but not on the JavaScript side. This results to the error not being unwrapped on the JS side.

adespawn:
The idea for this PR, is to include the name of the Rust error type, in the thrown error.
In the future, current approach can be extended to contain more information,
that cannot be passed with the current error handling implementation.

The other benefit of this PR is that all error construction will happen on the Rust side,
so we no longer need to remember which function may return errors, and handle them (see #216)

I turns out https://doc.rust-lang.org/std/ops/trait.FromResidual.html is still experimental, so we cannot implement a struct that we can use `?` on an we need to do it a bit more creativly.

If this were to be merged, after updating all places where we return napi::Error, this would fix:

Fixes #119
Fixes #216 (We update the whole process, so this is no longer necessary)


src/errors.rs
```diff
@@ -47,6 +50,145 @@ impl Display for ErrorType {
...
+///
+/// You can either use Converted error, that supports some extra fields,
+/// or you can add a new error type. Note that `with_custom_error!` macro
+/// supports only ConvertedError conversion. Usage with different error requires manual handling.
+/// This is important, because we cannot implement ToNapiValue for Result<T, E> directly,
+/// And we cannot implement FromResidual for this class, as this is unstable feature.
+///
+/// When converting this class to JS value through ToNapiValue
+/// cases will be handled in the following way
+/// - Ok(T): T::to_napi_value will be return
```
wprzytula: :wrench:

wprzytula:
```suggestion
/// - Ok(T): T::to_napi_value will be returned
```

src/errors.rs
```diff
@@ -47,6 +50,145 @@ impl Display for ErrorType {
...
+/// Returning Error variants will result in throwing errors in JS.
+///
+/// You can either use Converted error, that supports some extra fields,
+/// or you can add a new error type. Note that `with_custom_error!` macro
+/// supports only ConvertedError conversion. Usage with different error requires manual handling.
+/// This is important, because we cannot implement ToNapiValue for Result<T, E> directly,
+/// And we cannot implement FromResidual for this class, as this is unstable feature.
+///
+/// When converting this class to JS value through ToNapiValue
+/// cases will be handled in the following way
```
wprzytula: :wrench:

wprzytula: @copilot 
```suggestion
/// When converting this class to JS value through ToNapiValue
/// cases will be handled in the following way:
```

src/errors.rs
```diff
@@ -47,6 +50,145 @@ impl Display for ErrorType {
...
+/// - Ok(T): T::to_napi_value will be return
+/// - Error(ConvertedError): We will create napi::Error based on ConvertedError ToNapiValue trait and return Err(Error)
+/// - NapiError(Error): Err(Error) will be returned
+pub enum JsResult<T> {
+    Ok(T),
+    Error(ConvertedError),
```
wprzytula: :thought_balloon: Shouldn't this rather be `ConvertibleError`? It's not converted at the point of construction, right?

src/errors.rs
```diff
@@ -47,6 +50,145 @@ impl Display for ErrorType {
...
+    unsafe fn to_napi_value(
+        env: napi::sys::napi_env,
+        val: Self,
+    ) -> napi::Result<napi::sys::napi_value> {
+        match val {
+            JsResult::Ok(val) => unsafe { T::to_napi_value(env, val) },
+            JsResult::Error(error) => {
+                // # Safety - Valid pointer to napi env is provided by the caller
+                let v = unsafe { ConvertedError::to_napi_value(env, error) }?;
+                // # Safety - We are creating Unknown from valid napi env (validity ensured by the caller) and value (created just above)
```
wprzytula: :pick: Why the `#` mark here?

src/errors.rs
```diff
@@ -47,6 +50,145 @@ impl Display for ErrorType {
...
+                let err = unsafe { Unknown::from_raw_unchecked(env, v) };
```
wprzytula: :question: What is `Unknown`? There's no documentation, at least none that I could found. How did you learn how to use it?

src/errors.rs
wprzytula: OK, I think `ConvertedError` is a valid name.

src/errors.rs
```diff
@@ -47,6 +50,145 @@ impl Display for ErrorType {
...
+            }
+            JsResult::NapiError(err) => Err(err),
+        }
+    }
+}
+
+pub struct ConvertedError {
+    msg: String,
+    name: String,
+}
```
wprzytula: :wrench: Please provide at least minimal docstrings.

src/errors.rs
```diff
@@ -47,6 +50,145 @@ impl Display for ErrorType {
...
+            //
+            // Why using such obscure method?
+            // Rust driver maintainer did not agree to add a code that would allow
+            // systematic conversion from the errors as they are into their names and so on...
+            //
+            // For this reason, we also only expose the error class name,
+            // and not the name of the specific kind, or the inner values of those errors.
+            name: std::any::type_name::<T>()
+                .split(":")
+                .last()
```
wprzytula: :pick: You can use `rsplit_once` for this.

src/errors.rs
```diff
@@ -47,6 +50,145 @@ impl Display for ErrorType {
...
+            // The split is done to cope with the fact that the type name
+            // may return type with full package or without.
+            //
+            // Why using such obscure method?
+            // Rust driver maintainer did not agree to add a code that would allow
+            // systematic conversion from the errors as they are into their names and so on...
+            //
+            // For this reason, we also only expose the error class name,
+            // and not the name of the specific kind, or the inner values of those errors.
+            name: std::any::type_name::<T>()
```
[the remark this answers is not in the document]
wprzytula: If you fear to lose flexibility of accepting `name` which is not `'static`, you can use `Cow`.

src/errors.rs
```diff
@@ -47,6 +50,145 @@ impl Display for ErrorType {
...
+macro_rules! with_custom_error {
+    ($code: block) => {
+        async {
+            let c: Result<_, ConvertedError> = async { $code }.await.into();
+
+            c.into()
+        }
+        .await
+    };
+}
```
wprzytula: :100: Good idea! But does this really need to be a macro? Can't we make this a function?

test/unit/js-results-tests.js
```diff
@@ -0,0 +1,87 @@
...
+            assert.strictEqual(e.name, "BadKeyspaceName");
+        }
+    });
+    it("should get the correct error from JsResult::NapiError", function () {
+        try {
+            rust.testsReturnJsResult(3);
+            assert.fail("This function should throw");
+        } catch (e) {
+            helper.assertInstanceOf(e, Error);
+            assert.strictEqual(e.message, "Error#Napi-error");
```
wprzytula: :question: Where does this `#` come from?

src/tests/js_results_tests.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[napi]
+pub async fn tests_return_js_result_async(kind: i32) -> JsResult<i32> {
+    match kind {
+        1 => JsResult::Ok(1),
+        2 => JsResult::Error(scylla::errors::BadKeyspaceName::Empty.into()),
+        3 => JsResult::NapiError(js_error("Napi-error")),
+        _ => unimplemented!("Unexpected test kind"),
+    }
+}
```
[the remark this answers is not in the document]
adespawn: #66

src/errors.rs
adespawn: https://std-dev-guide.rust-lang.org/policy/safety-comments.html#inside-unsafe-elements

I assumed we must write `# Safety` before calls to unsafe, but I guess I was wrong about this one and it turns out only `Safety` is recomended

src/errors.rs
adespawn:
```rs
#[derive(Clone, Copy)]
/// Represents a raw JavaScript value
pub struct Unknown<'env>(
  pub(crate) Value,
  pub(crate) std::marker::PhantomData<&'env ()>,
);
```
https://github.com/napi-rs/napi-rs/blob/dc4d475eeec4249cb77928c8800369ae890e713e/crates/napi/src/js_values/unknown.rs#L8-L13

I've found this class among the classes that have implemented `From<T> for Error` trait. This was the only option that I could find to create Error with custom fields - through creating custom JS value (unknown) and then converting it to error

wprzytula: Definitely. This is not intuitive at all.

src/session.rs
```diff
@@ -169,22 +168,23 @@ impl SessionWrapper {
...
+            let (result, paging_state_response) = self
+                .inner
+                .get_session()
+                .query_single_page(statement, params, paging_state)
+                .await?;
+
+            Ok(PagingResult {
+                result: QueryResultWrapper::from_query(result)?,
+                paging_state: paging_state_response.into(),
+            })
```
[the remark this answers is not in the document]
adespawn: I've addressed all the comments. Let me know if it looks ok now

src/errors.rs
```diff
@@ -47,6 +50,168 @@ impl Display for ErrorType {
...
+        match value {
+            Ok(v) => JsResult::Ok(v),
+            Err(e) => JsResult::Error(e.into()),
+        }
+    }
+}
+
+pub(crate) trait ToConvertedError<Final> {
+    fn cnv(self) -> Final;
+}
```
wprzytula: :wrench: Customary name for such conversion would be `to_converted_error`.

:wrench: Also, `to_*` methods take `&self`, while `into_*` take `self`. Once again please take a look at https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv.

:thought_balloon: Shouldn't `Final` be an associated type instead of generic type parameter?

src/errors.rs
wprzytula: Then let's use `rsplit().next()`.

PR #318, opened 2025-11-14
Re-enable full tuple example

adespawn:


examples/other/tuple-insert-select-simple.js
```diff
@@ -1,46 +0,0 @@
-"use strict";
-const cassandra = require("scylladb-javascript-driver");
```
wprzytula: :question: Are we still importable by this name?
:camping: We should switch to `scylladb-nodejs-rs-driver` or perhaps `scylladb-driver`, if the language is obvious from the context.

wprzytula (approved):

Issue #186, opened 2025-04-15
Implement load balancing policies


ZuzaOsa:
There are several [implemented load balancing policies](https://docs.datastax.com/en/developer/nodejs-driver/4.7/api/module.policies/module.loadBalancing/index.html) in the DataStax driver.

- [ ] RoundRobinPolicy 
- [ ] DCAwareRoundRobinPolicy 
- [ ] TokenAwarePolicy 
- [ ] AllowListPolicy 
- [ ] WhiteListPolicy

adespawn:
Currently, we support some of the policies. There were some changes to the behaviour of those policies that need to be documented. Allow List policy can be implemented without the implementation of custom load balancing policies. We may consider removing LegacyDefault policy, as it's unlikely we can implement its current behaviour.

PR #326, opened 2025-11-21
Partial load balancing

adespawn:
This PR adds a new load balancing policy that is based on the Rust Load Balancing policy. I also attempt to add support for some of the old built-in policies, although proper testing and, as a result, resolution of #186 will be done in a future PR.

Right now, this is not tested, as all of the current integration tests rely on info.queried hosts, which are not available in the rust driver.

### DIFF
--- a/lib/policies/index.js
+++ b/lib/policies/index.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const { changedBehaviorWarning } = require("../new-utils");
+
 /**
  * Contains driver tuning policies to determine [load balancing]{@link module:policies/loadBalancing},
  *  [retrying]{@link module:policies/retry} queries, [reconnecting]{@link module:policies/reconnection} to a node,
@@ -29,14 +31,23 @@ exports.defaultAddressTranslator = function () {
 /**
  * Returns a new instance of the default load-balancing policy used by the driver.
  * @param {string} [localDc] When provided, it sets the data center that is going to be used as local for the
- * load-balancing policy instance.
+ * load-balancing policy instance, with dc failover disabled.
  *
- * When localDc is undefined, the load-balancing policy instance will use the `localDataCenter`
- * provided in the {@link ClientOptions}.
- * @returns {LoadBalancingPolicy}
+ * When localDc is undefined, the load balancing policy will not be data-center aware.
+ * @returns {DefaultLoadBalancingPolicy}
  */
 exports.defaultLoadBalancingPolicy = function (localDc) {
-    return new loadBalancing.LegacyDefaultLoadBalancingPolicy(localDc);
+    // TODO: Document the differences and upgrade instructions for load balancing policy.
+    // And then link it here.
+    changedBehaviorWarning("The default load balancing policy has changed.");
+
+    if (!localDc) {
+        changedBehaviorWarning("The behavior of defaultLoadBalancingPolicy with undefined localDc has changed significantly.");
+        return new loadBalancing.DefaultLoadBalancingPolicy();
+    }
+
+    return new loadBalancing.DefaultLoadBalancingPolicy({ localDc: localDc, permitDcFailover: false });
+
 };
 
 /**
--- a/test/unit-broken/default-load-balancing-policy-tests.js
+++ b/test/unit-broken/default-load-balancing-policy-tests.js
@@ -510,19 +510,6 @@ describe("DefaultLoadBalancingPolicy", () => {
     });
 });
 
-describe("policies.defaultLoadBalancingPolicy()", () => {
-    [
-        { title: "without the local dc", localDc: undefined },
-        { title: "with a local dc", localDc: "my_dc" },
-    ].forEach(({ title, localDc }) => {
-        it(`should support creating a new instance ${title}`, () => {
-            const lbp = policies.defaultLoadBalancingPolicy(localDc);
-            assert.instanceOf(lbp, DefaultLoadBalancingPolicy);
-            assert.strictEqual(lbp.localDc, localDc);
-        });
-    });
-});
-
 function testPreferredHost(useReplicas, isPreferredRemote) {
     const localDcLength = 7;
     const replicaCondition = (h) => lastOctetOf(h) > 1;
--- a/test/unit/default-policies-tests.js
+++ b/test/unit/default-policies-tests.js
@@ -1,0 +1,17 @@
+"use strict";
+
+const { assert } = require("chai");
+
+const {
+    DefaultLoadBalancingPolicy,
+    LoadBalancingConfig,
+} = require("../../lib/policies/load-balancing");
+const { policies } = require("../../main");
+
+describe("policies.defaultLoadBalancingPolicy()", () => {
+    it(`should support creating a new instance`, () => {
+        const lbp = policies.defaultLoadBalancingPolicy();
+        assert.instanceOf(lbp, DefaultLoadBalancingPolicy);
+        assert.instanceOf(lbp.getRustConfiguration(), LoadBalancingConfig);
+    });
+});


lib/policies/index.d.ts
```diff
@@ -65,6 +65,18 @@ export namespace policies {
...
+      preferDatacenter?: string;
+      preferRack?: string;
+      tokenAware?: boolean;
+      permitDcFailover?: boolean;
+      enableShufflingReplicas?: boolean;
+    }
+    
+    class NewDefaultLoadBalancingPolicy extends LoadBalancingPolicy {
+      constructor(config?: LoadBalancingConfig);
+    }
```
wprzytula: This will at some point no longer be _new_.
I'd call it simply `DefaultLoadBalancingPolicy`.

lib/policies/load-balancing.js
```diff
@@ -903,6 +923,139 @@ function setLocalDc(lbp, client, hosts) {
     }
 }
 
+/**
+ * This load balancing policy replicates the behavior of the Rust Driver Default Load Balancing Policy.
+ *
+ * It can be configured to be datacenter-aware, rack-aware and token-aware.
+ * Datacenter failover (sending query to a node from a remote datacenter)
+ * for queries with non local consistency mode is also supported.
```
wprzytula: AFAIR we changed the logic of DC failover to be possible also for queries with local consistency.
Please update the docs here. If you could be that kind and update Rust docs as well, it would be great :pray:

lib/policies/index.d.ts
wprzytula: Ugh, but there already is something called `DefaultLoadBalancingPolicy`...

lib/policies/index.js
```diff
@@ -28,15 +28,10 @@ exports.defaultAddressTranslator = function () {
...
- * When localDc is undefined, the load-balancing policy instance will use the `localDataCenter`
- * provided in the {@link ClientOptions}.
- * @returns {LoadBalancingPolicy}
+ * @returns {NewDefaultLoadBalancingPolicy}
  */
-exports.defaultLoadBalancingPolicy = function (localDc) {
-    return new loadBalancing.DefaultLoadBalancingPolicy(localDc);
+exports.defaultLoadBalancingPolicy = function () {
+    return new loadBalancing.NewDefaultLoadBalancingPolicy();
 };
```
wprzytula: :thought_balloon: For maintaining as much compatibility as possible, shouldn't we accept `localDc` arg? Dunno.

lib/policies/load-balancing.js
```diff
@@ -922,10 +922,13 @@ class NewDefaultLoadBalancingPolicy extends LoadBalancingRustImplemented {
     #config;
 
     /**
-     * @param {LoadBalancingConfig} config
+     * @param {LoadBalancingConfig} [config]
```
wprzytula: :recycle:  This should belong to an earlier commit, right?

lib/policies/load-balancing.js
```diff
@@ -54,6 +54,15 @@ class LoadBalancingPolicy {
...
     }
+
+    /**
+     * @returns {LoadBalancingConfig}
+     * @package
+     */
+    getRustConfiguration() {
+        // This error will be thrown by all policies, that do not override this method.
+        throw new Error("Currently this policy is not supported by the driver");
+    }
```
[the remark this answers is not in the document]
wprzytula:
```suggestion
    /**
     * @returns {LoadBalancingConfig}
     * @package
     */
    getRustConfiguration() {
        // This error will be thrown by all policies, that do not override this method.
        throw new Error("Currently this load balancing policy is not supported by the driver");
    }
```

lib/policies/index.d.ts
adespawn: If we were to have a name `DefaultLoadBalancingPolicy`, then some user programs that rely on the old default may break. But on the other hand, currently we don't support the old default, so their programs will still break, but at least we can be vocal about it, instead of risking some UB.

lib/policies/index.js
adespawn: On first glance, the local Dc parameter seems to have similar semantics to preferDc with permitDcFailover = false, but also, this policy does a lot more stuff, which I didn't spend the time on trying to understand

src/session.rs
```diff
@@ -46,6 +60,7 @@ pub struct BatchWrapper {
 #[napi]
 pub struct SessionWrapper {
     pub(crate) inner: CachingSession,
+    pub(crate) session_load_balancing_policy: Option<Arc<dyn LoadBalancingPolicy>>,
 }
```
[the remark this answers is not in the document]
adespawn: Should that be configuerd through `default_execution_profile_handle` in session?
I may have missed that

adespawn:
Updated the docs

adespawn:
Rebased on main

lib/new-utils.js
```diff
@@ -104,9 +104,21 @@ function arbitraryValueToBigInt(value) {
...
+ * The goal of this functions is to generate (potentially toggle-able in the future) warnings,
+ * when someone uses a feature, which behavior has changed in comparison to the DSx driver.
+ * @param {string} message
+ */
+function changedBehaviorWarning(message) {
+    console.warn(
+        `Using feature, which behavior has changed in comparison to the DataStax driver: ${message}`,
+    );
+}
+
```
wprzytula: ⛏️

lib/policies/load-balancing.js
```diff
@@ -903,6 +923,139 @@ function setLocalDc(lbp, client, hosts) {
...
 
+/**
+ * This load balancing policy replicates the behavior of the Rust Driver Default Load Balancing Policy.
+ *
+ * It can be configured to be datacenter-aware, rack-aware and token-aware.
+ * When the policy is datacenter-aware, you can configure whether to allow datacenter failover
+ * (sending query to a node from a remote datacenter).
+ */
+// TODO: come up with a better name...
+class NewDefaultLoadBalancingPolicy extends LoadBalancingRustImplemented {
```
wprzytula: Yep, let's resolve this.

lib/policies/index.js
```diff
@@ -29,14 +31,23 @@ exports.defaultAddressTranslator = function () {
...
+ * @returns {NewDefaultLoadBalancingPolicy}
  */
 exports.defaultLoadBalancingPolicy = function (localDc) {
-    return new loadBalancing.DefaultLoadBalancingPolicy(localDc);
+    // TODO: Document the differences and upgrade instructions for load balancing policy.
+    // And then link it here.
+    changedBehaviorWarning("The default load balancing policy has changed.");
+
+    if (localDc === undefined) {
+        changedBehaviorWarning("The behaviour of defaultLoadBalancingPolicy with undefined localDc has changed significantly.");
```
wprzytula: You're mixing AE with BE spelling of `behavio(u)r`. Please settle on one.

wprzytula: Also, provide exhaustive information about the change here.

src/session.rs
```diff
@@ -228,23 +230,42 @@ impl SessionWrapper {
...
+        with_custom_error_sync(|| {
+            let mut batch: Batch = Default::default();
+            statements
+                .into_iter()
+                .for_each(|q| batch.append_statement(q.as_str()));
+
+            batch = self.apply_batch_options(batch, &options.options)?;
+            ConvertedResult::Ok(BatchWrapper { inner: batch })
+        })
+    }
```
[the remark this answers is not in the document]
adespawn: Resolved in #360

src/session.rs
```diff
@@ -356,9 +369,53 @@ fn configure_session_builder(options: &SessionOptions) -> ConvertedResult<Sessio
...
+            }
+            if let Some(permit_dc_failover) = config.permit_dc_failover {
+                builder = builder.permit_dc_failover(permit_dc_failover);
+            }
+            if let Some(enable_shuffling_replicas) = config.enable_shuffling_replicas {
+                builder = builder.enable_shuffling_replicas(enable_shuffling_replicas);
+            }
+            Ok(Some(builder.build()))
+        }
+        None => Ok(None),
```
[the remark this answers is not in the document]
wprzytula: `let-else`, not `if-let`.

PR #327, opened 2025-11-21
Add support for client ID

### LINKED ISSUE #322: Report application information to server
Add following keys to startup options map:

- [x]     APPLICATION_NAME - ID what application is using driver, example: repo of the application
- [x]    APPLICATION_VERSION - Version of the application, example: release version or commit id of the application
- [ ]    CLIENT_ID - unique id of the client instance, example: pod name

Migrated to Jira: [DRIVER-242](https://scylladb.atlassian.net/browse/DRIVER-242)

[DRIVER-242]: https://scylladb.atlassian.net/browse/DRIVER-242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

adespawn:
Compared to the DSx driver, this supports ids both as uuid and strings. When uuid is provided, it's converted to string, as this is how client id is represented in the Rust driver.

Fixes #322

Closes [DRIVER-242](https://scylladb.atlassian.net/browse/DRIVER-242)

[DRIVER-242]: https://scylladb.atlassian.net/browse/DRIVER-242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -1,3 +1,4 @@
 pub mod error_throwing_tests;
+pub mod option_tests;
 pub mod result_tests;
 pub mod utils_tests;
--- a/src/tests/option_tests.rs
+++ b/src/tests/option_tests.rs
@@ -1,0 +1,45 @@
+use crate::session::{SessionOptions, SslOptions};
+
+#[napi]
+pub fn tests_check_client_option(options: SessionOptions, case: i32) {
+    match case {
+        1 => {
+            assert_eq!(
+                options,
+                SessionOptions {
+                    connect_points: Some(vec![
+                        "Contact point 1".to_owned(),
+                        "Contact point 2".to_owned()
+                    ]),
+                    keyspace: Some("keyspace name".to_owned()),
+                    application_name: Some("App name".to_owned()),
+                    application_version: Some("App version".to_owned()),
+                    client_id: Some("Client id".to_owned()),
+                    credentials_username: Some("Unique username".to_owned()),
+                    credentials_password: Some("Unique password".to_owned()),
+                    cache_size: Some(2137),
+                    ssl_options: Some(SslOptions {
+                        reject_unauthorized: Some(false)
+                    })
+                }
+            )
+        }
+        2 => {
+            assert_eq!(
+                options,
+                SessionOptions {
+                    connect_points: None,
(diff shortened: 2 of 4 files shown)


lib/client-options.js
```diff
@@ -46,10 +46,9 @@ const errors = require("./errors.js");
  *
  * @property {String} [credentials.username] The username to use for plain-text authentication.
  * @property {String} [credentials.password] The password to use for plain-text authentication.
- * @property {Uuid} [id] A unique identifier assigned to a {@link Client} object, that will be communicated to the
- * server (DSE 6.0+) to identify the client instance created with this options. When not defined, the driver will
+ * @property {Uuid | string} [id] A unique identifier assigned to a {@link Client} object, that will be communicated to the
+ * server to identify the client instance created with this options. When not defined, the driver will
```
wprzytula: > When not defined, the driver will generate a random identifier.

This does not seem to be true.

wprzytula (changes requested):

lib/client-options.js
adespawn: It's generated in client constructor

wprzytula: This proves we lack tests for this.

lib/client-options.js
```diff
@@ -664,15 +664,17 @@ function setRustOptions(options) {
     rustOptions.applicationName = options.applicationName;
     rustOptions.applicationVersion = options.applicationVersion;
     rustOptions.keyspace = options.keyspace;
-    rustOptions.cacheSize = options.maxPrepared;
+    if (rustOptions.cacheSize) {
+        rustOptions.cacheSize = options.maxPrepared;
+    }
```
wprzytula: Shouldn't this condition be `if (options.maxPrepared)`?

wprzytula (changes requested):

lib/client-options.js
wprzytula: I believe the error is still there.
Which convinces me to delay merging this until we have proper testing.

Issue #294, opened 2025-10-29
add benchmark case for large paged SELECTs


wprzytula:
Let's have another benchmark case for multi-page SELECTs returning large amounts of data.
Suggestion: a single query would return hundreds of pages, each with hundreds of rows.
Such benchmark would measure performance of deserialization process well, diminishing impact of other sources of overhead.

PR #331, opened 2025-11-25
Paging benchmark

adespawn:
Add sync paging benchmark and large select benchmark.

large select benchmark fixes: #294 

Example results 
graph

### DIFF
--- a/benchmark/runner.py
+++ b/benchmark/runner.py
@@ -227,6 +227,5 @@ commit = data.stdout.replace('\n', '')
 wh = SyncWebhook.from_url(os.environ['DISCORD_BENCHMARKS_WEBHOOK'])
 
 wh.send(content="Branch: " + branch +
-        " commit: https://github.com/scylladb-zpp-2024-javascript-driver/" +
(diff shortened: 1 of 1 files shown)


benchmark/logic_rust/paging.rs
```diff
@@ -0,0 +1,74 @@
...
+        let mut iter = session
+            .get_session()
+            .execute_iter(prepared.clone(), &[])
+            .await?
+            .rows_stream::<(Uuid, i32)>()
+            .unwrap();
+        let mut sm = 0;
+        while let Some(r) = iter.next().await {
+            sm += r.unwrap().1;
+        }
```
wprzytula: :thought_balloon: We're using `execute_iter` here but `execute_single_page` API in Node.js Driver, right?
If so, we should stick to `execute_single_page` in Rust to compare them more fairly.

adespawn: For n = 10_000
Rust-iterator: 47s
Rust-single-page: 57s
DSx: 61s
Node-rs: 102s

Issue #345, opened 2025-12-10
Update Error integration tests


adespawn:
Because #313 is merged, we can update all of the integration tests to check the name field in the error tests

PR #90, opened 2024-11-15
Add eslint

ZuzaOsa:
Added eslint to check for issues in code for example space after ``//``.
- Rule about using prototypes was switched off beacuse Datastax code violates it. After switching the code to use classes the rule should be switched on.
- Added camelcase, no-var and spaced-comment rules.
- no-unused-vars rule was switched off for error catches and unused function parameters.
- While changing things to satisfy linter some geometry leftovers were removed in ``encoder.js``. Spaces were added in the comments. In ``integer.js`` some variables were switched from using snake_case to camelCase and ``var`` replaced with ``let`` (i have *several* questions about that code by the way).

PR #354, opened 2025-12-11
Update integration tests to check the error name 

adespawn:
This PR takes the new field of the errors - name that was added in #313 and updates the test that check for specific errors, to use this filed, instead of trying to filter for a specific error message, as was done in some of the tests (and some tests didn't check for anything more then the fact that the error was thrown)

Additionally, this PR fixes a bug with error handling in execute concurrent that I discovered when enabling some additional tests.

Fixes: #345

### DIFF
--- a/test/integration/supported/ssl-tests.js
+++ b/test/integration/supported/ssl-tests.js
@@ -4,7 +4,6 @@ const util = require("util");
 
 const helper = require("../../test-helper");
 const Client = require("../../../lib/client");
-const errors = require("../../../lib/errors");
 const utils = require("../../../lib/utils");
 const types = require("../../../lib/types");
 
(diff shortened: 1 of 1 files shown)


test/integration/supported/client-batch-tests.js
```diff
@@ -1075,18 +1041,11 @@ describe("Client @SERVER_API", function () {
...
-                    assert.ok(err);
-                    assert.ok(
-                        err.message.includes(
-                            "Preparation failed on every connection from the selected pool.",
-                        ),
-                    );
-                    // helper.assertInstanceOf(err, errors.ResponseError);
-                    return client.shutdown();
+                    helper.assertErrorWithName(err, "PrepareError");
                 });
```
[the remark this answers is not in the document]
adespawn: (For now) shutown is deprecated

test/test-helper.js
```diff
@@ -297,6 +297,12 @@ const helper = {
         );
     },
 
+    assertErrorWithName: function (err, name) {
+        assert.ok(err, "Expected to receive an error");
```
wprzytula: :thinking: Shouldn't this message say more about what we received instead of an error?

test/integration/supported/client-execute-tests.js
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable no-unused-vars */
+/* eslint camelcase: ["error", {"properties": "never"}] */
```
wprzytula: :question: Please explain this - here and in a comment.

adespawn: It's a linter change that allows having sanke_case in fields that represent udt objects. A small progress towards: #90

Issue #361, opened 2025-12-18
Queried host integration test failure


adespawn:
For some reason, the newly introduced integration test (#357) fails CI on the fork repo:

https://github.com/adespawn/nodejs/actions/runs/20342270140/job/58444686760

PR #362, opened 2025-12-19
Update queried host flakey test

adespawn:
The integration test introduced in #357 turned out to be flaky when testing against Scylla.

For Scylla, we may get the address with shard aware port, while the client.hosts returns the hosts with non shard aware ports. (See https://github.com/scylladb/scylla-rust-driver/issues/1513)

For this reason, I updated the test when testing with Scylla, to strip the port part of the address, fixing the flakiness.

Fixed: #361

### DIFF
--- a/test/integration/supported/client-execute-tests.js
+++ b/test/integration/supported/client-execute-tests.js
@@ -183,15 +183,29 @@ describe("Client @SERVER_API", function () {
             client.execute(helper.queries.basic, function (err, result) {
                 assert.equal(err, null);
                 assert.notEqual(result, null);
+                let hosts = client.hosts.values().map((h) => h.address);
+                let queriedHost = result.info.queriedHost;
 
-                for (const host of client.hosts.values()) {
-                    if (host.address === result.info.queriedHost) {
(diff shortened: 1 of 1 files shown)


test/integration/supported/client-execute-tests.js
```diff
@@ -183,15 +183,27 @@ describe("Client @SERVER_API", function () {
...
+                let queriedHost = result.info.queriedHost;
 
-                for (const host of client.hosts.values()) {
-                    if (host.address === result.info.queriedHost) {
+                if (helper.getServerInfo().isScylla) {
+                    // For Scylla, we may get the address with shard aware port,
+                    // while the client.hosts returns the hosts with non shard aware ports.
+                    // (See https://github.com/scylladb/scylla-rust-driver/issues/1513)
+                    // For this reason, when testing with Scylla, we strip the port part of the address.
+                    hosts = hosts.map((address) => address.split(":")[0]);
```
wprzytula: BUG POTENTIAL if IPv6 with 0s shortened to `:`.

wprzytula (approved):

Issue #99, opened 2024-11-18
Add support for named parameters


adespawn:

adespawn:
Moving past GA. Support for named parameters in prepared statements was added in #421, and to close this issue we would need to support named parameters in uprepared statements. 

Since it's not necessery for GA, moving this issue past GA

PR #421, opened 2026-03-26
Named parameters

adespawn:
This PR introduces support for named parameters in prepared (queries/statements). By reusing handling logic (code in `utils.js`), I aim to keep the API compatibility with DSx driver.

At a later point, we can have a discussion about whether we want to still accept case-insensitive parameters, or do we want to enforce case sensitivity.

This does not fully close​ #99, as the DSx driver had support for named parameters in unprepared (queries/statements) in some special cases (when type hints were used in a specific way). Such support will be added in the following PRs.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -176,8 +176,9 @@ class Client extends events.EventEmitter {
      */
     async prepareStatement(statement) {
         let expectedTypes = await this.rustClient.prepareStatement(statement);
-        let types = expectedTypes.map((t) => convertComplexType(t));
-        return new PreparedInfo(types, statement);
+        let types = expectedTypes.map((t) => convertComplexType(t[0]));
+        let boundParamNames = expectedTypes.map((t) => t[1].toLowerCase());
+        return new PreparedInfo(types, statement, boundParamNames);
     }
 
     /**
@@ -322,7 +323,7 @@ class Client extends events.EventEmitter {
     /**
      * Wrapper for executing queries by rust driver
      * @param {string | PreparedInfo} query
-     * @param {Array} params
+     * @param {Array | Object} params
      * @param {ExecutionOptions} execOptions
      * @returns {Promise<ResultSet>}
      * @package
@@ -350,6 +351,10 @@ class Client extends events.EventEmitter {
                 prepared = await this.prepareStatement(query);
             } else {
                 prepared = query;
+            }
+
+            if (withNamedParameters) {
+                params = utils.adaptNamedParamsPrepared(params, prepared);
             }
 
             let encoded = encodeParams(prepared.types, params, this.#encoder);
@@ -500,7 +505,7 @@ class Client extends events.EventEmitter {
     /**
      * Execute a single page of query
      * @param {string} query
-     * @param {Array} params
+     * @param {Array | Object} params
      * @param {ExecutionOptions} execOptions
      * @param {rust.PagingStateWrapper|Buffer} [pageState]
      * @returns {Promise<Array<rust.PagingStateResponseWrapper, ResultSet>>} should be Promise<[rust.PagingStateResponseWrapper, ResultSet]>
@@ -542,6 +547,10 @@ class Client extends events.EventEmitter {
                 prepared = await this.prepareStatement(query);
             } else {
                 prepared = query;
+            }
+
+            if (withNamedParameters) {
+                params = utils.adaptNamedParamsPrepared(params, prepared);
             }
 
             let encoded = encodeParams(prepared.types, params, this.#encoder);
@@ -729,6 +738,10 @@ class Client extends events.EventEmitter {
                 }
                 types = prepared.types;
                 statement = prepared.statement;
+
+                if (isNamedParameters(params, execOptions)) {
+                    params = utils.adaptNamedParamsPrepared(params, prepared);
+                }
             } else {
                 types = hints[i] || [];
             }
--- a/lib/client.js
+++ b/lib/client.js
@@ -3,7 +3,11 @@
 
 const events = require("events");
 const util = require("util");
-const { throwNotSupported, PreparedInfo } = require("./new-utils.js");
+const {
+    throwNotSupported,
+    isNamedParameters,
+    PreparedInfo,
+} = require("./new-utils.js");
 
 const utils = require("./utils.js");
 const errors = require("./errors.js");
@@ -324,18 +328,7 @@ class Client extends events.EventEmitter {
      * @package
      */
     async rustyExecute(query, params, execOptions) {
-        if (
-            // !execOptions.isPrepared() &&
-            params &&
-            !Array.isArray(params)
-            // && !types.protocolVersion.supportsNamedParameters(version)
-        ) {
-            throw new Error(`TODO: Implement any support for named parameters`);
-            // // Only Cassandra 2.1 and above supports named parameters
-            // throw new errors.ArgumentError(
-            //   "Named parameters for simple statements are not supported, use prepare flag",
-            // );
-        }
+        let withNamedParameters = isNamedParameters(params, execOptions);
 
         if (!this.connected) {
             // TODO: Check this logic and decide if it's needed. Probably do it while implementing (better) connection
@@ -514,18 +507,7 @@ class Client extends events.EventEmitter {
      * @private
      */
     async #rustyPaged(query, params, execOptions, pageState) {
-        if (
-            !execOptions.isPrepared() &&
-            params &&
-            !Array.isArray(params)
-            // && !types.protocolVersion.supportsNamedParameters(version)
-        ) {
-            throw new Error(`TODO: Implement any support for named parameters`);
-            // // Only Cassandra 2.1 and above supports named parameters
-            // throw new errors.ArgumentError(
-            //   "Named parameters for simple statements are not supported, use prepare flag",
-            // );
-        }
+        let withNamedParameters = isNamedParameters(params, execOptions);
 
         if (!this.connected) {
             // TODO: Check this logic and decide if it's needed. Probably do it while implementing (better) connection
--- a/lib/client.js
+++ b/lib/client.js
@@ -167,14 +167,14 @@ class Client extends events.EventEmitter {
 
     /**
      * Manually prepare query into prepared statement
-     * @param {string} query
+     * @param {string} statement
      * @returns {Promise<list<Object | string>>}
      * Returns a tuple of type object (the format expected by the encoder) and prepared statement wrapper
      * @package
      */
-    async prepareQuery(query) {
-        let expectedTypes = await this.rustClient.prepareStatement(query);
-        let res = [expectedTypes.map((t) => convertComplexType(t)), query];
+    async prepareStatement(statement) {
+        let expectedTypes = await this.rustClient.prepareStatement(statement);
+        let res = [expectedTypes.map((t) => convertComplexType(t)), statement];
         return res;
     }
 
@@ -258,7 +258,7 @@ class Client extends events.EventEmitter {
      *
      * It returns a `Promise` when a `callback` is not provided.
      *
-     * @param {string} query The query to execute.
+     * @param {string} statement The query to execute.
      * @param {Array|Object} [params] Array of parameter values or an associative array (object) containing parameter names
      * as keys and its value.
      * @param {QueryOptions} [options] The query options for the execution.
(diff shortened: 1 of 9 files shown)


wprzytula:
> support for named parameters in unprepared (queries/statements) in some special cases (when type hints were used in a specific way).

Please describe how this is supposed to work in the documentation.

lib/concurrent/index.js
```diff
@@ -155,7 +156,7 @@ class ArrayBasedExecutor {
         try {
             let prepared = this._cache.getElement(query);
             if (!prepared) {
-                prepared = await (this._client.prepareQuery(query));
+                prepared = await PreparedInfo.create(query, this._client.rustClient);
```
wprzytula: ♻️ Let's retain the `prepareStatement` method, which will now return `PreparedInfo`.

lib/client.js
```diff
@@ -554,26 +537,21 @@ class Client extends events.EventEmitter {
         const rustOptions = execOptions.getRustOptions();
         let result;
         if (execOptions.isPrepared()) {
+            /**
+             * @type {PreparedInfo}
+             */
+            let prepared = query;
```
wprzytula:
```suggestion
            /**
             * @type {PreparedInfo}
             */
            let prepared;

File: lib/client.js
Comment (wprzytula) [inline lib/client.js:507]: ❗ You've just told me this type hint is invalid, because some internal calls pass something (`PreparedInfo`) other than `string`!

File: lib/prepared.js [hunk 3]
```diff
@@ -0,0 +1,26 @@
+const _rust = require("../index");
+const { convertComplexType } = require("./types/cql-utils");
+
+class PreparedInfo {
+    /**
+     * @param {Array<type>} types
+     * @param {string} query
```
wprzytula: ♻️ `statement`

lib/new-utils.js
wprzytula, line 1: ♻️ Can we merge `utils.js` with `new-utils.js`?

lib/new-utils.js
```diff
@@ -137,10 +137,23 @@ function ensure64SignedInteger(number, name) {
...
+    if (params && !Array.isArray(params)) {
+        if (!execOptions.isPrepared()) {
+            throw new customErrors.ArgumentError(
+                "Named parameters for simple statements are not supported, use prepare flag",
+            );
+        }
+        return true;
+    }
+    return false;
+}
```
wprzytula, line 170: 🔧 Please check that the provided value is an Object. `number`s make no sense as named parameters...

lib/utils.js
```diff
@@ -288,26 +288,20 @@ function validateFn(fn, name) {
...
-        return params;
-    }
-    const paramsArray = new Array(columns.length);
+    const paramsArray = new Array(columns.types.length);
     params = toLowerCaseProperties(params);
-    const keys = {};
-    for (let i = 0; i < columns.length; i++) {
-        const name = columns[i].name;
+    for (let i = 0; i < columns.types.length; i++) {
+        const name = columns.colNames[i];
```
wprzytula: ♻️ Extract a separate commit that replaces use of the magic unnamed type with `PreparedInfo`. Only then introduce new functionality (support for named parameters).

lib/prepared.js
```diff
@@ -5,10 +5,12 @@ class PreparedInfo {
...
      * @param {Array<type>} types
      * @param {string} query
+     * @param {Array<string>} colNames
      */
-    constructor(types, query) {
+    constructor(types, query, colNames) {
         this.types = types;
         this.query = query;
+        this.colNames = colNames;
     }
```
wprzytula: 🔧 These are bound variable names (bound parameter names), not columns' names.

src/requests/request.rs
```diff
@@ -51,12 +51,17 @@ impl QueryOptionsWrapper {
...
             .get_variable_col_specs()
             .iter()
-            .map(|e| ComplexType::new_owned(e.typ().clone()))
+            .map(|e| {
+                (
+                    ComplexType::new_owned(e.typ().clone()),
+                    e.name().to_string(),
+                )
+            })
             .collect()
```
wprzytula, line 65: 📌 Note to myself: use Claude to optimise allocations on ToNapiValue for strings and vecs.

wprzytula (changes requested):

lib/client.js
adespawn: It turns out I was wrong. Only `rustyExecute` can be called with already prepared statements. The code that assumes query can be already prepared is mostly copied from `rustyExecute` - something #422 will resolve.

lib/new-utils.js
adespawn: I can do it in #422, as that PR is focused on refators

lib/client.js
```diff
@@ -165,16 +165,14 @@ class Client extends events.EventEmitter {
...
      */
-    async prepareQuery(query) {
-        let expectedTypes = await this.rustClient.prepareStatement(query);
-        let res = [expectedTypes.map((t) => convertComplexType(t)), query];
-        return res;
+    async prepareQuery(statement) {
+        let expectedTypes = await this.rustClient.prepareStatement(statement);
+        let types = expectedTypes.map((t) => convertComplexType(t));
+        return new PreparedInfo(types, statement);
     }
```
wprzytula: 🔧
```suggestion
    async prepareStatement(statement) {
        let expectedTypes = await this.rustClient.prepareStatement(statement);
        let types = expectedTypes.map((t) => convertComplexType(t));
        return new PreparedInfo(types, statement);
    }
```

lib/client.js
```diff
@@ -318,7 +316,7 @@ class Client extends events.EventEmitter {
 
     /**
      * Wrapper for executing queries by rust driver
-     * @param {string | list<Object | string>} query
+     * @param {string | PreparedInfo} query
```
wprzytula, line 325:
```suggestion
     * @param {string | PreparedInfo} statement
```

lib/client.js
```diff
@@ -348,26 +346,21 @@ class Client extends events.EventEmitter {
...
+            /**
+             * @type {PreparedInfo}
+             */
+            let prepared = query;
             // If the statement is already prepared, skip the preparation process
             // Otherwise call Rust part to prepare a statement
             if (typeof query === "string") {
-                query = await this.prepareQuery(query);
+                prepared = await this.prepareQuery(query);
             }
```
wprzytula: 🔧 Please don't assign a value of bad type (`query`, being `string`, to `prepared`, being `PreparedInfo`).

lib/utils.js
```diff
@@ -285,29 +285,23 @@ function validateFn(fn, name) {
...
-    if (!params || Array.isArray(params) || !columns || columns.length === 0) {
-        // params is an array or there aren't parameters
-        return params;
-    }
-    const paramsArray = new Array(columns.length);
+    const paramsArray = new Array(columns.types.length);
     params = toLowerCaseProperties(params);
-    const keys = {};
-    for (let i = 0; i < columns.length; i++) {
-        const name = columns[i].name;
```
wprzytula, line 304: 🔧 These are not columns, as we already discussed!
These are bound parameters.

lib/client.js
wprzytula: I see you fixed this in a later commit, but not in the original one.

lib/client.js
```diff
@@ -349,18 +343,24 @@ class Client extends events.EventEmitter {
...
+            if (withNamedParameters) {
+                params = utils.adaptNamedParamsPrepared(params, prepared);
             }
 
             let encoded = encodeParams(prepared.types, params, this.#encoder);
 
             // Execute query
             result = await this.rustClient.executePreparedUnpagedEncoded(
-                prepared.query,
+                prepared.statement,
```
wprzytula, line 364: ditto: bad commit for fixing this.

wprzytula: Such renames should constitute a separate commit.

lib/client.js
wprzytula, line 521: ❓ 🌱  Unrelated to this PR: what about this?

lib/utils.js
```diff
@@ -296,7 +296,7 @@ function adaptNamedParamsPrepared(params, columns) {
     const paramsArray = new Array(columns.types.length);
     params = toLowerCaseProperties(params);
     for (let i = 0; i < columns.types.length; i++) {
-        const name = columns.colNames[i];
+        const name = columns.boundParamNames[i];
```
wprzytula: Again, fixed in bad commit.

wprzytula (changes requested):

lib/client.js
adespawn: Probably we can take a look at it after GA

wprzytula: Maybe let's open an issue about this?

lib/utils.js
adespawn: Sorry, that was leftover from the DSx naming. Didn't catch it

lib/client.js
adespawn: To make it consistent I would need to update all endpoints in the client. Maybe let's move it to #422, as this is a refactor PR?

lib/client.js
adespawn: #439

lib/client.js
wprzytula: OK.

wprzytula: Be careful not to introduce API-breaking changes on the go.

lib/client.js
```diff
@@ -165,16 +165,14 @@ class Client extends events.EventEmitter {
...
      */
-    async prepareQuery(query) {
-        let expectedTypes = await this.rustClient.prepareStatement(query);
-        let res = [expectedTypes.map((t) => convertComplexType(t)), query];
-        return res;
+    async prepareStatement(statement) {
+        let expectedTypes = await this.rustClient.prepareStatement(statement);
+        let types = expectedTypes.map((t) => convertComplexType(t));
+        return new PreparedInfo(types, statement);
     }
```
wprzytula, line 182: ♻️ commit `Introduce Prepared Info class` should be split into two:
1. rename `query` -> `statement`,
2. actually introduce `PreparedInfo`.

lib/utils.js
```diff
@@ -285,29 +285,23 @@ function validateFn(fn, name) {
...
  * @throws {Error} In case a parameter with a specific name is not defined
  */
-function adaptNamedParamsPrepared(params, columns) {
-    if (!params || Array.isArray(params) || !columns || columns.length === 0) {
-        // params is an array or there aren't parameters
-        return params;
-    }
-    const paramsArray = new Array(columns.length);
+function adaptNamedParamsPrepared(params, prepared) {
+    const paramsArray = new Array(prepared.types.length);
```
wprzytula, line 296: 🔧
1. Docstring still mentions _columns_.
2. Missing period.

lib/client.js
```diff
@@ -171,8 +175,9 @@ class Client extends events.EventEmitter {
      */
     async prepareStatement(statement) {
         let expectedTypes = await this.rustClient.prepareStatement(statement);
-        let types = expectedTypes.map((t) => convertComplexType(t));
-        return new PreparedInfo(types, statement);
+        let types = expectedTypes.map((t) => convertComplexType(t[0]));
+        let colNames = expectedTypes.map((t) => t[1].toLowerCase());
+        return new PreparedInfo(types, statement, colNames);
```
wprzytula: 🔧 `colNames` is a wrong name.

lib/client.js
```diff
@@ -317,24 +322,13 @@ class Client extends events.EventEmitter {
...
-            params &&
-            !Array.isArray(params)
-            // && !types.protocolVersion.supportsNamedParameters(version)
-        ) {
-            throw new Error(`TODO: Implement any support for named parameters`);
-            // // Only Cassandra 2.1 and above supports named parameters
-            // throw new errors.ArgumentError(
-            //   "Named parameters for simple statements are not supported, use prepare flag",
-            // );
-        }
```
wprzytula, line 340: ♻️ Extracting this as a function should be done in a separate precommit, for clean diff and not to confuse readers.

wprzytula (changes requested):

lib/client.js
adespawn: Renaming public API endpoints arguments will not be a breaking change

adespawn:
Rebased (no changes)

wprzytula (dismissed):

lib/new-utils.js
```diff
@@ -138,10 +138,42 @@ function ensure64SignedInteger(number, name) {
     }
 }
 
+class PreparedInfo {
+    /**
+     * @param {Array<type>} types
```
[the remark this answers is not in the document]
wprzytula: #426

wprzytula (approved):

PR #422, opened 2026-03-27
Client refactors

adespawn:
The main motivation for this refactor is that in the integration tests, we only tested the rusty paged version of the internal interface. This increases the chance of improper updates to the code, as we need to remember to update 2 places instead of 1, and we do not need to unnecessarily extend integration tests to handle both branches.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -299,9 +299,7 @@ class Client extends events.EventEmitter {
             const execOptions = this.createOptions(options);
             if (execOptions.isPaged()) {
                 return promiseUtils.optionalCallback(
-                    this.#rustyPaged(query, params, execOptions).then(
-                        (e) => e[1],
-                    ),
+                    this.#rustyPaged(query, params, execOptions),
                     callback,
                 );
             }
@@ -458,8 +456,7 @@ class Client extends events.EventEmitter {
 
         /**
          * @param {Error} err
-         * @param {Array<rust.PagingStateResponseWrapper, ResultSet>} result
-         * Should be [rust.PagingStateResponseWrapper, ResultSet]
+         * @param {ResultSet} result
          */
         function pageCallback(err, result) {
             if (err) {
@@ -468,32 +465,29 @@ class Client extends events.EventEmitter {
             /**
              * Next requests in case paging (auto or explicit) is used
              */
-            let lastPagingState = result[0];
-            let queryResult = result[1];
+            rowLength += result.rowLength;
 
-            rowLength += queryResult.rowLength;
-
-            if (queryResult.rows) {
-                queryResult.rows.forEach((value, index) => {
+            if (result.rows) {
+                result.rows.forEach((value, index) => {
                     rowCallback(index, value);
                 });
             }
 
-            if (lastPagingState) {
+            if (result.innerPageState) {
                 // Use new page state as next request page state
-                pagingState = lastPagingState;
+                pagingState = result.innerPageState;
 
                 if (execOptions.isAutoPage()) {
                     // Issue next request for the next page
                     return nextPage();
                 }
                 // Allows for explicit (manual) paging, in case the caller needs it
-                queryResult.nextPage = nextPage;
+                result.nextPage = nextPage;
             }
 
             // Finished auto-paging
-            queryResult.rowLength = rowLength;
-            callback(null, queryResult);
+            result.rowLength = rowLength;
+            callback(null, result);
         }
 
         promiseUtils.toCallback(
@@ -508,7 +502,7 @@ class Client extends events.EventEmitter {
      * @param {Array | Object} params
      * @param {ExecutionOptions} execOptions
      * @param {rust.PagingStateWrapper|Buffer} [pageState]
-     * @returns {Promise<Array<rust.PagingStateResponseWrapper, ResultSet>>} should be Promise<[rust.PagingStateResponseWrapper, ResultSet]>
+     * @returns {Promise<ResultSet>}
      * @private
      */
     async #rustyPaged(query, params, execOptions, pageState) {
@@ -597,9 +591,7 @@ class Client extends events.EventEmitter {
                 );
             };
         }
-        result[1] = resultSet;
-
-        return result;
+        return resultSet;
     }
 
     /**
--- a/lib/client.js
+++ b/lib/client.js
@@ -332,6 +332,7 @@ class Client extends events.EventEmitter {
         // This should probably be re-worked when stabilizing the API. For now I do not want to restrict the API to avoid
         // accidentally breaking compatibility...
         const paged = pageState !== undefined;
+        pageState = normalizePagingState(pageState, execOptions);
 
         let withNamedParameters = isNamedParameters(params, execOptions);
 
@@ -339,23 +340,6 @@ class Client extends events.EventEmitter {
             // TODO: Check this logic and decide if it's needed. Probably do it while implementing (better) connection
             // // Micro optimization to avoid an async execution for a simple check
             await this.#connect();
-        }
-
-        if (paged) {
-            if (pageState instanceof Buffer) {
-                pageState = rust.PagingStateWrapper.fromBuffer(pageState);
-            } else if (pageState == null) {
-                // Take the page state option into account only when we don't pass it explicitly (it's done only in eachRow)
-                if (execOptions.getPageState() instanceof Buffer) {
-                    pageState = rust.PagingStateWrapper.fromBuffer(
-                        execOptions.getPageState(),
-                    );
-                } else if (typeof execOptions.pageState === "string") {
-                    pageState = rust.PagingStateWrapper.fromBuffer(
-                        Buffer.from(execOptions.getPageState(), "hex"),
-                    );
-                }
-            }
         }
 
         let rustOptions = execOptions.getRustOptions();
@@ -798,6 +782,49 @@ class Client extends events.EventEmitter {
      * @param {ResultSet | PromiseLike<ResultSet>} value
      * @returns {void}
      */
+}
+
+/**
+ * This function normalizes the type, and combines all possible page state sources.
+ *
+ * Page state can come from two sources:
+ * - pageState variable: when we execute the following page of the query
+ * - execOptions: when the user start a new query from requested state
+ *
+ * PageState has a priority over execOptions, as execOptions.pageState
+ * will not be cleared for the following pages of the execution.
+ *
+ * @param {rust.PagingStateWrapper|Buffer|null|undefined} pageState
+ * @param {ExecutionOptions} execOptions
+ * @returns {rust.PagingStateWrapper|undefined}
+ */
+function normalizePagingState(pageState, execOptions) {
+    let optionsPageState = execOptions.getPageState();
+    switch (true) {
+        // Paging is disabled, so we do not care about value of paging state
+        case pageState === undefined:
+            return undefined;
+        // Paging is enabled, but no one has provided starting page state - undefined represents such case
+        case pageState === null && optionsPageState == undefined:
+            return undefined;
+        // Normalize pageState first, since it has a priority over execOptions
+        case pageState instanceof rust.PagingStateWrapper:
+            return pageState;
+        case pageState instanceof Buffer:
+            return rust.PagingStateWrapper.fromBuffer(pageState);
+        // If we have paging enabled, but no page state from last page, normalize exec options page state
+        case pageState === null && optionsPageState instanceof Buffer:
+            return rust.PagingStateWrapper.fromBuffer(optionsPageState);
+        case pageState === null && typeof optionsPageState === "string":
+            return rust.PagingStateWrapper.fromBuffer(
+                Buffer.from(optionsPageState, "hex"),
+            );
+        default:
+            // Only cases with invalid types will be caught here: all accepted types are accepted in the earlier branches
+            throw new TypeError(
+                `Invalid paging state. Provided PageState ${pageState}, execOptions pageState ${optionsPageState}`,
+            );
+    }
 }
 
 /**
--- a/lib/client.js
+++ b/lib/client.js
@@ -345,85 +345,33 @@ class Client extends events.EventEmitter {
         let rustOptions = execOptions.getRustOptions();
 
         /**
-         * @type {rust.QueryResultWrapper}
+         * @type {rust.PagingResultWithExecutor}
          */
-        let result;
-        /**
-         * @type {rust.PagingStateWrapper?}
-         */
-        let resultPageState;
-        /**
-         * @type {rust.QueryExecutor?}
-         */
-        let executor;
+        let resultTuple;
 
         if (execOptions.isPrepared()) {
-            /**
-             * @type {PreparedInfo}
-             */
-            let prepared;
-            // If the statement is already prepared, skip the preparation process
-            // Otherwise call Rust part to prepare a statement
-            if (typeof query === "string") {
-                prepared = await this.prepareStatement(query);
-            } else {
-                prepared = query;
-            }
-
-            if (withNamedParameters) {
-                params = utils.adaptNamedParamsPrepared(params, prepared);
-            }
-
-            let encoded = encodeParams(prepared.types, params, this.#encoder);
-
-            if (paged) {
-                let temp = await this.rustClient.executeSinglePageEncoded(
-                    prepared.statement,
-                    encoded,
-                    rustOptions,
-                    pageState,
-                );
-                result = temp[1];
-                resultPageState = temp[0];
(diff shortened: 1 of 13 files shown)


lib/types/result-set.js
```diff
@@ -41,10 +41,17 @@ class ResultSet {
...
+     * @type {rust.PagingStateWrapper?}
+     * @package
+     */
+    innerPageState;
+
     /**
      * @param {rust.QueryResultWrapper} result
      * @param {_Encoder} encoder
-     * @param {rust.PagingStateResponseWrapper} [pagingState]
+     * @param {rust.PagingStateWrapper} [pagingState]
```
wprzytula: ♻️ This fix should constitute a separate commit.

lib/client.js
```diff
@@ -432,155 +494,49 @@ class Client extends events.EventEmitter {
...
                 // Allows for explicit (manual) paging, in case the caller needs it
-                queryResult.nextPage = nextPage;
+                result.nextPage = nextPage;
             }
 
             // Finished auto-paging
-            queryResult.rowLength = rowLength;
-            callback(null, queryResult);
+            result.rowLength = rowLength;
+            callback(null, result);
```
wprzytula: ♻️ Unrelated change - should go in a separate commit.

wprzytula (changes requested):
Please do the following in separate commits:
1. First do unrelated smaller refactors (each in a separate commit).
2. Then refactor `rustyExecute` to be usable in the multi-paged code path.
3. Then replace `rustyPaged` with `rustyExecute`.

lib/client.js
```diff
@@ -319,14 +314,24 @@ class Client extends events.EventEmitter {
     }
 
     /**
-     * Wrapper for executing queries by rust driver
+     * Wrapper for executing queries by rust driver.
+     * When called with a pageState argument (including null), executes a single-page query.
+     * When called without pageState, executes an unpaged query.
      * @param {string | PreparedInfo} query
-     * @param {Array | Object} params
+     * @param {Array} params
```
wprzytula: ❓ 
This shouldn't belong to this commit.

lib/client.js
```diff
@@ -319,14 +314,24 @@ class Client extends events.EventEmitter {
...
      * @param {ExecutionOptions} execOptions
+     * @param {rust.PagingStateWrapper|Buffer|null} [pageState] When provided (including null), enables paged execution.
+     * When unprovided (undefined), executes an unpaged query.
      * @returns {Promise<ResultSet>}
      * @package
      */
-    async rustyExecute(query, params, execOptions) {
+    async rustyExecute(query, params, execOptions, pageState) {
+        // Why not just take execOptions.isPaged()?
+        // When executing through eachRow, this may not be set properly, so we need to account for that case.
```
wprzytula: 🔧 Elaborate why this may be not set properly on `eachRow`.

lib/client.js
```diff
@@ -335,8 +340,37 @@ class Client extends events.EventEmitter {
...
+                    pageState = rust.PagingStateWrapper.fromBuffer(
+                        execOptions.getPageState(),
+                    );
+                } else if (typeof execOptions.pageState === "string") {
+                    pageState = rust.PagingStateWrapper.fromBuffer(
+                        Buffer.from(execOptions.getPageState(), "hex"),
+                    );
+                }
+            }
+        }
```
wprzytula: ♻️ Try using `switch` here for readability.

wprzytula: ♻️ Consider extracting this as `normalizePagingState`.

lib/client.js
wprzytula: ♻️ `switch` please.

lib/client.js
```diff
@@ -357,12 +391,23 @@ class Client extends events.EventEmitter {
...
+                result = temp[1];
+                resultPageState = temp[0];
+                executor = temp[2];
+            } else {
+                result = await this.rustClient.executePreparedUnpagedEncoded(
+                    prepared.statement,
+                    encoded,
+                    rustOptions,
+                );
+            }
```
wprzytula: 📌 Let's remove the superfluous `Encoded` suffix.

wprzytula (changes requested):
Good job! 🚀

adespawn:
Addressed comments and added a new commit with changes related to TS

lib/client.js
```diff
@@ -297,15 +297,10 @@ class Client extends events.EventEmitter {
...
             const execOptions = this.createOptions(options);
-            if (execOptions.isPaged()) {
-                return promiseUtils.optionalCallback(
-                    this.#rustyPaged(query, params, execOptions),
-                    callback,
-                );
-            }
-
+            // The rusty execute will take the page state from options, but we need to indicate whether it's a paged query.
+            let pageState = execOptions.isPaged() ? null : undefined;
```
wprzytula, line 303: > page state from options

🤔 Did you actually mean this? Shouldn't this be `page state from query result`?

wprzytula: I'm confused.

lib/client.js
```diff
@@ -800,6 +784,39 @@ class Client extends events.EventEmitter {
...
+            typeof execOptions.getPageState() === "string":
+            return rust.PagingStateWrapper.fromBuffer(
+                Buffer.from(execOptions.getPageState(), "hex"),
+            );
+        default:
+            throw new TypeError(
+                `Invalid paging state. Provided PageState ${pageState}, execOptions pageState ${execOptions.getPageState()}`,
+            );
+    }
+}
```
wprzytula, line 943: 💭 I'm extremely confused by the logic here.

1. There are two sources of possible page state:
- `pageState`
- `execOptions.getPageState()`
How are they different? Naming isn't helpful.

2. No comments here, no clue what happens why.

lib/client.js
```diff
@@ -345,86 +345,34 @@ class Client extends events.EventEmitter {
         let rustOptions = execOptions.getRustOptions();
 
         /**
-         * @type {rust.QueryResultWrapper}
+         * @type {rust.PagingResultWithExecutor}
          */
-        let result;
-        /**
```
wprzytula: ♻️ It's an extremely bad name for a variable.

lib/client.js
```diff
@@ -437,6 +385,110 @@ class Client extends events.EventEmitter {
...
+        }
+        return [
+            undefined,
+            await this.rustClient.executePreparedUnpagedEncoded(
+                prepared.statement,
+                encoded,
+                rustOptions,
+            ),
+            undefined,
+        ];
```
wprzytula, line 466: ⛏️ Let's comment what these `undefined`s are.

wprzytula, line 466: ♻️ Let's use `if-else`, not an early return, because none of the variants is more unusual than the other.

lib/client.js
```diff
@@ -437,6 +385,110 @@ class Client extends events.EventEmitter {
...
+
+        return [
+            undefined,
+            await this.rustClient.queryUnpagedEncoded(
+                query,
+                encoded,
+                rustOptions,
+            ),
+            undefined,
+        ];
```
wprzytula, line 512: ditto about `if-else`

wprzytula, line 512: 🤔 It's quite strange that we have two separate endpoints for unpaged executions - one for prepared and one for unprepared statements - but only one for single page executions. Why so?

lib/client.js
```diff
@@ -375,12 +377,18 @@ class Client extends events.EventEmitter {
...
         let resultSet = new ResultSet(result, this.#encoder, resultPageState);
         if (resultPageState) {
-            resultSet.rawNextPageAsync = async (pageState) => {
-                return await executor.fetchNextPage(
-                    this.rustClient,
-                    rust.PagingStateWrapper.fromBuffer(pageState),
-                );
-            };
+            // If resultPageState then executor must be defined according to type definition.
+            assert(executor != undefined);
```
wprzytula: 💭 Maybe a stronger assertion: `executor instanceof Executor`?

lib/client.js
```diff
@@ -414,11 +422,23 @@ class Client extends events.EventEmitter {
...
+        /**
+         * @type {Array<any>}
+         */
+        let unifiedParams;
         if (withNamedParameters) {
-            params = utils.adaptNamedParamsPrepared(params, prepared);
+            unifiedParams = utils.adaptNamedParamsPrepared(params, prepared);
+        } else {
+            // This will not fail, as withNamedParameters is set false only if params is an array.
+            assert(Array.isArray(params));
```
wprzytula: 💭 Is `params` `Array` _if and only if _ `withNamedParameters`? If so, then maybe let's remove the redundant `withNamedParameters` argument and don't even need such assertion?

lib/promise-utils.js
```diff
@@ -92,7 +92,7 @@ function newQueryPlan(lbp, keyspace, executionOptions) {
  * When callback is undefined it returns the promise.
  * When using a callback, it will use it as handlers of the continuation of the promise.
  * @param {Promise} promise
- * @param {Function?} callback
+ * @param {Function} [callback]
```
wprzytula, line 95: ❓ What do the brackets mean here?

wprzytula (changes requested):

lib/client.js
adespawn: When setting up the project, we specifically disallowed this pattern:

https://eslint.org/docs/latest/rules/no-else-return#allowelseif

Unless you have good reason to - I wouldn't switch it

lib/client.js
adespawn: This assertion is only necessary fot type checks. When resultPageState is provided, rust will always return executor, but type checks do not seem to understand this assumption (see change in custom-types.d.td). That's why we need this assertion.

lib/client.js
adespawn: I'm not sure what you mean here:

We have 4 enpoints (prepared, paged), (unprepared, paged), (prepared, unpaged), (unprepared, unpaged)

lib/client.js
adespawn: No. Here we are starting to execute from the state the user has provided so we take page state from the query options

lib/promise-utils.js
adespawn: {value?} argName = {value | null}
{value} [argName] = {value | undefined}

At least according to ts type checker. And we use this function with undefined and not null

lib/client.js
wprzytula: The lint is about `else if` after `return`. Here I suggest using `else`, not `else if`. The lint guards against different situations than what we have here.

lib/client.js
wprzytula: Yep, I don't know what tricked me.

lib/client.js
wprzytula: OK, but if we anyway need to assert, then why not do a stronger assertion that still needs to pass?

lib/client.js
adespawn: image

lib/client.js
adespawn: The diff does not contain any variable name. Did you meant the PagintResult type, or something else?

wprzytula: I meant `tmp` or `temp`.

lib/client.js
wprzytula: Ugh, this is not what's specified in the lint documentation you provided...

adespawn:
Reduced the number of type errors even more from 67(initial)->37(initial changes)->~12~ 14 errors(current)

adespawn:
Aghr, JS docs is stricking again...

lib/client.js
adespawn: I've reworked the documentation (so less LLM bulshit and hopefully more meaningful comment)

adespawn:
Cut down another 4 type safety errors

lib/client.js
```diff
@@ -332,6 +332,7 @@ class Client extends events.EventEmitter {
         // This should probably be re-worked when stabilizing the API. For now I do not want to restrict the API to avoid
         // accidentally breaking compatibility...
         const paged = pageState !== undefined;
+        pageState = normalizePagingState(pageState, paged, execOptions);
```
wprzytula: 🔧 It makes no sense to me to pass two distinct separate to `normalizePagingState`, where one (`paged`) is derived from another (`pageState`). This increases number of options that `normalizePagingState` function's reader must consider (`pageState` being `undefined` and `paged` set to `true`, or the opposite case). Please remove the `paged` parameter from `normalizePagingState`.

lib/client.js
```diff
@@ -800,6 +784,45 @@ class Client extends events.EventEmitter {
...
+            typeof execOptions.getPageState() === "string":
+            return rust.PagingStateWrapper.fromBuffer(
+                Buffer.from(execOptions.getPageState(), "hex"),
+            );
+        default:
+            throw new TypeError(
+                `Invalid paging state. Provided PageState ${pageState}, execOptions pageState ${execOptions.getPageState()}`,
+            );
+    }
+}
```
wprzytula, line 943: 👍 Doccoment is now clear enough for me.

🔧 Case handling is very convoluted and not obvious. For example, if `!paged` AND `pageState instanceof Buffer`, we return `pageState` unchanged, without normalization.

wprzytula: 🔧 Enhance doccomment:
1. Write all possible inputs and all possible outputs.
2. Explain why such mapping takes place (why such precedence).

wprzytula: 🔧 Improve case handling:
Cover all cases. Imagine you're on a PMat exercises session and you need to consider all cases exhaustively. If some cases make no sense, they should be guarded with an assertion and a comment.

lib/metadata/client-state.js
```diff
@@ -39,6 +39,7 @@ class ClientState {
 
     static from(_client) {
         this.log("warning", deprecatedMsg);
+        return new ClientState();
     }
```
wprzytula, line 43: ❓ ❗ What was that? Did this return `undefined`?

lib/client.js
wprzytula, line 943: ♻️ This should be part of the original commit that introduce `normalizePagingState`.

wprzytula (changes requested):

lib/metadata/client-state.js
adespawn: Yup. This is how we handled this deprecated class

lib/client.js
adespawn: Should be better now

lib/metadata/client-state.js
```diff
@@ -39,6 +39,7 @@ class ClientState {
 
     static from(_client) {
         this.log("warning", deprecatedMsg);
+        return new ClientState();
```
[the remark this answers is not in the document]
wprzytula: @adespawn Is this concern true?

lib/client.js
```diff
@@ -288,48 +290,62 @@ class Client extends events.EventEmitter {
...
      * @package
      */
-    async rustyExecute(query, params, execOptions) {
-        let withNamedParameters = isNamedParameters(params, execOptions);
+    async rustyExecute(query, params, execOptions, pageState) {
+        // Why not just take execOptions.isPaged()?
+        // When executing through eachRow, user may not this may not be set the isPaged query option properly
+        // (set it to false - this is not checked in any place at the moment).
+        // This should probably be re-worked when stabilizing the API. For now I do not want to restrict the API to avoid
+        // accidentally breaking compatibility...
```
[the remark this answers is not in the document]
wprzytula: Is this addressed? @adespawn

lib/client.js
```diff
@@ -839,6 +899,49 @@ class Client extends events.EventEmitter {
...
+        // Paging is enabled, but no one has provided starting page state - undefined represents such case
+        case pageState === null && optionsPageState == undefined:
+            return undefined;
+        // Normalize pageState first, since it has a priority over execOptions
+        case pageState instanceof rust.PagingStateWrapper:
+            return pageState;
+        case pageState instanceof Buffer:
+            return rust.PagingStateWrapper.fromBuffer(pageState);
+        // If we paging enabled, but no page state from last page, normalize exec options page state
+        case pageState === null && optionsPageState instanceof Buffer:
```
wprzytula: > // If we paging enabled

⛏️ Yoda style 😉

wprzytula (dismissed):

lib/metadata/client-state.js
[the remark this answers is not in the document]
adespawn: We still inform the user that this endpoint is not supported, so I guess it's fine(ish).

lib/client.js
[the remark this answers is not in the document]
adespawn: Not it should be

lib/client.js
adespawn: image

wprzytula (approved):

PR #423, opened 2026-03-27
Integrate rust logging

### LINKED ISSUE #16: Integrate logging with Rust driver's logging (tracing) subsystem
Logging helps in debugging a lot. Let's make logging available and configurable, so that if we (or a user of our driver) turn logging on, then Rust driver emits logs, too.

Relevant Rust driver's documentation: https://rust-driver.docs.scylladb.com/stable/logging/logging.html

adespawn:
This PR was created with Claude Opus and reviewed by me.

This PR introduces new API for logging. As discussed in person, we cannot keep the original interface, which we change with this PR.

## Add configurable log forwarding from Rust driver to JS

Introduces a logging system that forwards internal driver log events to the
JS `Client` as `'log'` EventEmitter events, with configurable severity
filtering.

### Highlights

- **`logLevel` client option:** Controls the minimum severity of forwarded
  events. Set to one of `trace`, `debug`, `info`, `warning`, `error`, or
  `off` (default). Filtering happens on the native side before crossing the
  FFI boundary.
- **`types.logLevels` enum:** Provides type-safe access to log level values
  in both JS and TypeScript.
- **Per-client registration:** Each `Client` registers its own callback on
  `connect()` and unregisters on `shutdown()`. Multiple clients can coexist
  with independent log levels.
- **Four separate event arguments:** The `'log'` event delivers
  `(level, target, message, furtherInfo)` as separate arguments, matching
  the `cassandra-driver` event signature.
- **Native implementation:** A global `tracing` subscriber in Rust maintains
  a registry of per-client NAPI callbacks. `setupLogging()` returns an
  `Option<u32>` id (or `null` for `off`/invalid), used by `removeLogging()`
  to unregister.

### Documentation

- New standalone [Logging](docs/src/logging.md) page covering levels, event
  arguments, event sources, multi-client behavior, and usage examples.
- New "Logging" section in the [Migration guide](docs/src/migration_guide.md)
  documenting differences from `cassandra-driver`: logging off by default,
  `verbose` replaced by `trace`+`debug`, `target` replaces `className`,
  cross-client event visibility.

### Tests

- Unit tests for `setupLogging`/`removeLogging`: level filtering, all five
  levels delivered, `furtherInfo` with extra fields, multiple callbacks,
  removal stops delivery.
- Integration tests for log events during `connect()`, query execution, and
  DDL operations.
- API tests asserting `logLevels` enum values in JS and TypeScript.

Fixes: #16


wprzytula:
>     * New "Logging" section in the [Migration guide](docs/src/migration_guide.md)
>       documenting differences from `cassandra-driver`: logging off by default,

I have doubts. You say that you change the default logging level from (?) to "off". Why?
Especially ERROR level, but also WARN level logs are often used as a way to signal user
important situation in the driver. For example, Rust Driver automatically logs warnings
that come from the DB on the WARN level. It is expected that driver's users see them by default.

adespawn:
> It is expected that driver's users see them by default.

You still need to add some logic to be able to see any logs (`client.on("log"...`).

> Why?

Considering the above, I assumed that users when creating the logic will also set the desired log level.
But on the other hand, this logging should be mostly API compatible, so we could set it to WARN/ ERROR by default, so transitioning users have some default logging...

wprzytula:
> > It is expected that driver's users see them by default.
> 
> You still need to add some logic to be able to see any logs (`client.on("log"...`).
> 
> > Why?
> 
> Considering the above, I assumed that users when creating the logic will also set the desired log level. But on the other hand, this logging should be mostly API compatible, so we could set it to WARN/ ERROR by default, so transitioning users have some default logging...

Yep. Let's default to WARN.

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+use tracing_subscriber::{Layer, Registry};
+
+/// Shorthand for the NAPI callback type.
+///
+/// Generic parameters: `CalleeHandled = false`, `Weak = true` so that the
+/// callback does not prevent the Node.js event-loop from exiting.
+type LogCallback = ThreadsafeFunction<
+    /* T: */ FnArgs<(String, String, String, String)>,
+    /* Return: */ (),
+    /* CallJsBackArgs: */ FnArgs<(String, String, String, String)>,
```
wprzytula, line 20: ❓ What is `CallJsBackArgs` and why it's the same as `T`? Why are they distinct?

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+/// A callback registered by one client, together with the minimum severity
+/// level it is interested in.
+struct RegisteredCallback {
+    id: u32,
+    callback: LogCallback,
+    min_level: Level,
+}
+
+/// Monotonically increasing id counter for registered callbacks.
+static NEXT_ID: AtomicU32 = AtomicU32::new(0);
```
wprzytula: ⛏️ Maybe let's use `AtomicU64` to have a larger pool of IDs?

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+
+/// Registry of all currently-active per-client callbacks.
+/// Initialized (and the global subscriber installed) on the first call to
+/// `setup_logging`.
+static CALLBACKS: OnceLock<RwLock<Vec<RegisteredCallback>>> = OnceLock::new();
```
wprzytula, line 40: 🤔 I don't get why we need both `OnceLock` and `RwLock`. `RwLock` is enough to provide interior mutability, while it can be initialized to `RwLock::new(Vec::new))`.

What `OnceLock` does give is once-only semantics for the global tracing subscriber initialization. Perhaps we can do without that, though?

wprzytula: Ah, yes. You anyway initialize with `RwLock::new(Vec::new())`. In this case, you can replace `OnceLock` with `Once`, and use that `Once` for initializing the global subscriber only.

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+/// Maps a Rust `tracing::Level` to the JS log-level strings.
+fn level_to_js(level: &Level) -> &'static str {
+    match *level {
+        Level::TRACE => "trace",
+        Level::DEBUG => "debug",
+        Level::INFO => "info",
+        Level::WARN => "warning",
+        Level::ERROR => "error",
+    }
+}
```
wprzytula, line 64: 🔧 maybe: `rust_level_to_js_level`?

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+fn parse_level(level: &str) -> Option<Level> {
+    match level {
+        "trace" => Some(Level::TRACE),
+        "debug" => Some(Level::DEBUG),
+        "info" => Some(Level::INFO),
+        "warning" => Some(Level::WARN),
+        "error" => Some(Level::ERROR),
+        _ => None,
+    }
+}
```
wprzytula, line 77: maybe: `parse_js_level_to_rust_level`?

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+/// Returns `None` for `"off"` or unrecognized strings.
+fn parse_level(level: &str) -> Option<Level> {
+    match level {
+        "trace" => Some(Level::TRACE),
+        "debug" => Some(Level::DEBUG),
+        "info" => Some(Level::INFO),
+        "warning" => Some(Level::WARN),
+        "error" => Some(Level::ERROR),
+        _ => None,
+    }
```
wprzytula, line 76: 🤔 Shouldn't this treat unknown strings as `Result::Err`, while `off` as `Option::None`?

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+struct MessageVisitor {
+    message: String,
+    extras: String,
+}
```
wprzytula, line 81: 🔧 Missing documentation.

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+impl Visit for MessageVisitor {
+    fn record_debug(&mut self, field: &Field, value: &dyn fmt::Debug) {
+        if field.name() == "message" {
+            self.message = format!("{:?}", value);
+        } else {
+            if !self.extras.is_empty() {
+                self.extras.push_str(", ");
+            }
+            self.extras
+                .push_str(&format!("{}={:?}", field.name(), value));
```
wprzytula: 🔧 This needlessly allocates. Instead of `format!`, use `String`'s implementation of `std::fmt::Write`, with `write!(fmt, string, "{}={:?}", field.name(), value)`.

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+            message: String::new(),
+            extras: String::new(),
+        }
+    }
+}
+
+impl Visit for MessageVisitor {
+    fn record_debug(&mut self, field: &Field, value: &dyn fmt::Debug) {
+        if field.name() == "message" {
+            self.message = format!("{:?}", value);
```
wprzytula: 🔧 This needlessly allocates, while there already is a `String` created with `String::new()` in `MessageVisitor::new()`. Instead of `format!`, use `String`'s implementation of `std::fmt::Write`, with `write!(fmt, &mut self.message, "{:?}", value)`.

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+    fn record_str(&mut self, field: &Field, value: &str) {
+        if field.name() == "message" {
+            self.message = value.to_owned();
+        } else {
+            if !self.extras.is_empty() {
+                self.extras.push_str(", ");
+            }
+            self.extras.push_str(&format!("{}={}", field.name(), value));
+        }
+    }
```
wprzytula: ditto about needless allocations

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+        let mut visitor = MessageVisitor::new();
+        event.record(&mut visitor);
+
+        for cb in callbacks.iter() {
+            if *event_level <= cb.min_level {
+                cb.callback.call(
+                    FnArgs {
+                        data: (
+                            level_str.clone(),
+                            target.clone(),
```
wprzytula: 🔧 Small optimisation:
```suggestion
        // At least one callback wants this event — extract the fields.
        let level_str = level_to_js(event_level);
        let target = meta.target();

        let mut visitor = MessageVisitor::new();
        event.record(&mut visitor);

        for cb in callbacks.iter() {
            if *event_level <= cb.min_level {
                cb.callback.call(
                    FnArgs {
                        data: (
                            level_str.to_owned(),
                            target.to_owned(),
```

src/logging.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    let callbacks = get_or_init_callbacks();
+    callbacks.write().unwrap().push(RegisteredCallback {
+        id,
+        callback,
+        min_level: level,
+    });
+
+    Some(id)
+}
```
wprzytula, line 182: ❓ Assume there are two `Client`s, each with its own Rust Driver's `Session`, and both clients use logging. Will both clients receive logs originating from their own `Session` or from any `Session`?

IIUC, they will get logs from any `Session`. Therefore, setting the logging callbacks per `Client` makes no sense for me. There should only be a global callback optionally set. Otherwise, what's the use of separate per-`Client` logging settings if all `Client`s see all logs anyway?

wprzytula (changes requested):
For now, reviewed up till the second commit, inclusively.

src/logging.rs
adespawn: Very very very unlikely to overflow, but I can change it

wprzytula:
In-person review:
- useless last integration test,
- missing unit test of the visitor,
- callback per Client is OK,
- callbacks should be removed in a finalizer, not in shutdown,
- take a look at CPP-RS's impl of the logger bridge.

src/logging.rs
adespawn:
```    
    T,           // Input type passed to call()
    Return,      // Return type from JavaScript
    CallJsArgs,  // Arguments for JS function (usually same as T)
```

https://napi.rs/blog/function-and-callbacks.en#part-5-threadsafefunction-generic-parameters

We specify non default value for weak, so we need specify CallJsArgs, unless I'm missing some rust syntax

src/logging.rs
adespawn: Off is filtered at the JS code. 

https://github.com/scylladb/nodejs-rs-driver/blob/7eb8b18f9f75800449889af441dd213c2d4f2448/lib/client.js#L235

Invalid strings are also filtered
https://github.com/scylladb/nodejs-rs-driver/blob/7eb8b18f9f75800449889af441dd213c2d4f2448/lib/client-options.js#L737-L744
So this is a quiet ignoring of path that should not be taken

docs/src/logging.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Event sources
+
+Log events originate from two layers:
+
+- **Internal driver** — connection, query routing, and protocol events from
+  the driver internals. These carry module paths in `target`
+  (e.g. `"scylla::network::connection"`).
+- **Client** — high-level lifecycle events emitted by the `Client` class
+  (e.g. *"Connecting to cluster"*). These carry `"Client"` as the `target`.
```
wprzytula: 🔧 1. I dislike this distinction. We should clearly mention both Rust Driver and the JS wrapper as possible sources.
2. I dislike naming of the parts (_high-level lifecycle events_).
3. I dislike listing just a few specific parts of the Rust Driver. Let's be less specific, or be exhaustive in our specificity.

docs/src/logging.md
```diff
@@ -0,0 +1,119 @@
...
+  logLevel: types.logLevels.info,
+});
+
+client.on('log', (level, target, message, furtherInfo) => {
+  const extra = furtherInfo ? ` (${furtherInfo})` : '';
+  console.log(`[${level}] ${target}: ${message}${extra}`);
+});
+
+await client.connect();
+// [info] Client: Connecting to cluster using 'ScyllaDB JavaScript Driver' version ...
```
wprzytula: Name of the driver is incorrect.

docs/src/migration_guide.md
```diff
@@ -154,3 +154,47 @@ The `cassandra-driver` driver had some undocumented assumptions about the order
...
+## Logging
+
+See the [Logging](./logging.md) page for the full documentation of the new logging system.
+Below are the key differences from the `cassandra-driver`.
+
+### Logging is off by default
+
+The `cassandra-driver` always emitted `'log'` events. In this driver, logging
+is **disabled by default** and must be explicitly enabled via the `logLevel`
+client option. See [Enabling logging](./logging.md#enabling-logging).
```
wprzytula: 🤔 Haven't we talked about this, that we should keep the old default?

adespawn:
Rebase (no changes)

adespawn:
Rebase (again)

src/logging.rs
adespawn: API compatibility:

https://docs.datastax.com/en/developer/nodejs-driver/4.8/features/logging/index.html

adespawn:
> callbacks should be removed in a finalizer, not in shutdown,

Looking a bit more at it, I'm not sure it's a good idea: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry#notes_on_cleanup_callbacks

adespawn:
Partial changes

> useless last integration test

Done

adespawn:
> callbacks should be removed in a finalizer, not in shutdown,

Done, with a new test

adespawn:
The only remaining thing now should be logging format (which we discussed during in person review). 
Noteworthy changes:
- Change to default behaviour
- Internal notes on loggin

adespawn:
Since I'm not exactly sure about the "Rework logging" commit, I didn't merge it to a proper place yet - will do it once those changes are be accepted

lib/types/index.js
```diff
@@ -216,7 +216,7 @@ const distance = {
  * @property {String} info High-level informational messages.
  * @property {String} warning Potentially harmful situations.
  * @property {String} error Error conditions.
- * @property {String} off Disables internal driver log forwarding (default).
+ * @property {String} off Disables internal driver log forwarding.
  */
 const logLevels = {
```
wprzytula, line 221: 🔧 Which one is the default now? Let's mention it in the docstring.

lib/client-options.js
```diff
@@ -224,6 +224,18 @@ const { ExecutionProfile } = require("./execution-profile.js");
  * to represent CQL varint data type. Defaults to true.
  *
  * Note, that using Integer as Varint (`useBigIntAsVarint == false`) is deprecated.
+ * @property {String} [logLevel] The minimum severity of log events emitted by the driver.
+ *
+ * **WARNING:** While you can configure different log levels for different clients, each client will receive
+ * log messages from all clients.
```
wprzytula, line 230: 🔧 Each client will receive log messages from all clients, right, but only those that pass the log level filter, right?

lib/client-options.js
```diff
@@ -470,6 +482,7 @@ function defaultOptions() {
             useBigIntAsLong: true,
             useBigIntAsVarint: true,
         },
+        logLevel: undefined,
```
wprzytula: 🤔 This changes nothing, right? The key set to `undefined` is equivalent to no such key set, right?

lib/client-options.js
```diff
@@ -713,6 +728,24 @@ function validateEncodingOptions(encodingOptions) {
...
+function validateLogLevel(logLevel) {
+    if (
+        logLevel !== undefined &&
+        (typeof logLevel !== "string" || !validLogLevels.includes(logLevel))
+    ) {
+        throw new TypeError(
+            `logLevel must be one of ${validLogLevels.map((l) => `'${l}'`).join(", ")}`,
+        );
+    }
+}
```
wprzytula, line 747: 💭 This seems buggy.
1. What if a non-string junk object is passed?
2. What if a value of the log level enum is passed?

lib/client.js
```diff
@@ -234,13 +234,23 @@ class Client extends events.EventEmitter {
...
             const emitter = this.emit.bind(this);
+            const logLevel = this.options.logLevel || types.logLevels.trace;
+
+            if (!this.options.logLevel) {
+                this.log(
+                    "warning",
+                    "No logLevel configured. Defaulting to 'trace'. " +
+                        "Note, that this default behavior may change in future releases.",
+                );
+            }
```
wprzytula: 🤔  You're saying that no logLevel configured results in a 'trace' log level?
This is not acceptable. Production users who don't care about logs should get flooded with trace-granularity logs.
I'd prefer that no log level set here is equivalent to 'off`, BUT that a default log level ('info'?) is set earlier, in the options.

lib/client.js
```diff
@@ -252,6 +262,10 @@ class Client extends events.EventEmitter {
             // We should close the pools (if any) and reset the state to allow successive calls to connect()
             this.connected = false;
             this.connecting = false;
+            if (this.#loggingId !== undefined) {
+                rust.removeLogging(this.#loggingId);
+                this.#loggingId = undefined;
+            }
             this.emit("connected", err);
             throw err;
         }
```
wprzytula, line 278: ❓ Close what pools?

src/tests/logging_tests.rs
```diff
@@ -0,0 +1,22 @@
...
+    tracing::warn!(test_field = "warn_extra", "warn message from rust");
+    tracing::error!(test_field = "error_extra", "error message from rust");
+}
+
+/// Emit a single `tracing::info!` event with the given message.
+/// Useful for verifying that arbitrary messages arrive on the JS side.
+#[napi]
+pub fn tests_emit_log_info(message: String) {
+    tracing::info!("{}", message);
+}
```
wprzytula, line 22: ❓ Can't it be `&str`, really?

test/test-helper.js
```diff
@@ -697,7 +697,7 @@ const helper = {
             // retry all queries multiple times (for improved test resiliency).
             policies: { retry: new RetryPolicy() },
             // Default to 'warning' in tests to reduce noise.
-            logLevel: "warning",
+            logLevel: types.logLevels.warning,
```
wprzytula, line 700: :recycle: This shouldn't come in this commit (`Add logging unit and integration tests`), but rather in the original commit that introduced this.

test/typescript/api-generation-test.ts
```diff
@@ -89,6 +90,17 @@ function printEnum(enumObject: { [key: string]: any }, name: string): void {
...
+function printStringEnum(enumObject: { [key: string]: any }, name: string): void {
+  console.log(`  // ${name} enum values`);
+
+  Object.keys(enumObject)
+    .filter((k) => typeof enumObject[k] === "string")
+    .forEach((k) => {
+      console.log(`  s = ${name}.${k};`);
+    });
+  console.log();
+}
```
wprzytula, line 110: 🔧 Missing documentation. This should also contain an example: when input X, the output is Y.

test/unit/api-tests.js
```diff
@@ -51,6 +51,16 @@ describe("API", function () {
...
+
+    it("should expose logLevels enum on types", function () {
+        assert.ok(api.types.logLevels);
+        assert.strictEqual(api.types.logLevels.trace, "trace");
+        assert.strictEqual(api.types.logLevels.debug, "debug");
+        assert.strictEqual(api.types.logLevels.info, "info");
+        assert.strictEqual(api.types.logLevels.warning, "warning");
+        assert.strictEqual(api.types.logLevels.error, "error");
+        assert.strictEqual(api.types.logLevels.off, "off");
+    });
```
wprzytula, line 63: 🤔 Why is this put in `unit-not-supported`?

wprzytula: 📌  Please describe the conventions of `not-supported` directories very clearly in the internal documentation.

docs/src/logging.md
```diff
@@ -0,0 +1,132 @@
...
+```js
+client.on('log', (level, target, message, furtherInfo) => {
+  console.log(`${level} - ${target}: ${message}`);
+});
+```
+
+## Enabling logging
+
+Logging is **enabled by default**. When no `logLevel` is specified, all events
+are captured (at `trace` level).
```
wprzytula: 🔧 As said elsewhere, `trace` is too spammy. Let's use INFO or WARN as the default. Rust tracing subscribers use INFO by default.

docs/source/logging/logging.md
```diff
@@ -0,0 +1,132 @@
...
+
+```js
+const client = new Client({
+  contactPoints: ['127.0.0.1'],
+  logLevel: types.logLevels.off,
+});
+```
+
+The callback is registered when `connect()` is called and unregistered on
+`shutdown()`. No log events are emitted before the client connects.
```
wprzytula, line 46: 💭 That's a bit unfortunate. We lose any logs that are emitted earlier. Can't we somehow enable logging earlier? Probably not, because they must be tied to the client?

docs/src/logging.md
```diff
@@ -0,0 +1,132 @@
...
+| `logLevels.debug`   | `'debug'`   | Fine-grained diagnostic information (DEBUG events)   |
+| `logLevels.info`    | `'info'`    | High-level informational messages                    |
+| `logLevels.warning` | `'warning'` | Potentially harmful situations                       |
+| `logLevels.error`   | `'error'`   | Error conditions                                     |
+| `logLevels.off`     | `'off'`     | Disables logging entirely                            |
+
+The `logLevel` option acts as a **filter**: only events at or above the
+configured severity are delivered to the listener. Filtering happens on the
+native side, before crossing the FFI boundary, so suppressed events have
+virtually zero overhead.
```
wprzytula: ❓ Are you sure about this virtually zero overhead?

docs/source/logging/logging.md
```diff
@@ -0,0 +1,132 @@
...
+| `logLevels.off`     | None                                |
+| `logLevels.trace`   | All (TRACE and above)               |
+| `logLevels.debug`   | DEBUG and above                     |
+| `logLevels.info`    | INFO and above                      |
+| `logLevels.warning` | WARN and above                      |
+| `logLevels.error`   | ERROR only                          |
+
+The `trace` level is only suitable for debugging and is usually very
+noisy. We recommend gathering events from `info` and above in production
+environments.
```
wprzytula, line 78: That's why `trace` is not suitable as the default, definitely.

docs/src/migration_guide.md
```diff
@@ -154,3 +154,57 @@ The `cassandra-driver` driver had some undocumented assumptions about the order
...
+
+## Logging
+
+See the [Logging](./logging.md) page for the full documentation of the new logging system.
+Below are the key differences from the `cassandra-driver`.
+
+This driver introduces a concept of configurable logging levels.
+While logging levels were present in `cassandra-driver`, you could only filter for those levels,
+after receiving the log information. To allow for better performance, this driver allows you to
+configure received log levels at the client settings level.
```
wprzytula:
```suggestion
This driver introduces a concept of configurable logging levels.
While logging levels were already present in `cassandra-driver`, you could only filter according to those
levels after receiving the log information. To allow for better performance, this driver allows you to
configure received log levels before the logs are emitted, at the client settings level.
```

src/tests/logging_tests.rs
```diff
@@ -20,3 +20,27 @@ pub fn tests_emit_log_events() {
...
+    );
+}
+
+/// Emit an INFO event with no extra fields (message only).
+#[napi]
+pub fn tests_emit_log_message_only() {
+    tracing::info!("message with no extras");
+}
+
+/// Emit an INFO event with a single string extra field.
```
wprzytula, line 42: commit `[WIP] some unit tests - usefull or not?`
🔧 Write a better commit message: describe the tests included. This will help me and you assess if they are worth it.

test/unit/logging-tests.js
```diff
@@ -15,11 +16,13 @@ const { logLevels } = types;
...
+ * Allow any in-flight NonBlocking ThreadsafeFunction calls to drain
+ * Repeating it will heuristically reduce change of leaked events.
  */
-function drain() {
-    return new Promise((resolve) => setTimeout(resolve, 50));
+async function drain() {
+    for (let i = 0; i < 20; i++) {
+        await new Promise((resolve) => setTimeout(resolve, 0));
+    }
 }
```
wprzytula: ♻️ This should likely belong to an earlier commit.

test/unit/logging-tests.js
```diff
@@ -305,4 +308,74 @@ describe("Logging", function () {
...
+            try {
+                await client.connect();
+            } catch (_) {
+                // connect will fail without a real cluster — that's expected
+            }
+
+            await drain();
+            return { client, events };
+        }
+
```
wprzytula, line 332: 🤔 Won't this break due to spawning many clients, each of whom shares the same Rust tracing subscriber?

docs/src/internal/logging.md
```diff
@@ -0,0 +1,43 @@
+# Logging
+
+This is a document describing internal consideration when it comes to logging.
+To see the documentation about the logging feature as present in the driver, see [logging docs](../logging.md).
+
+## Default behavior
+
+In the DSx driver, there was no option to configure the logging levels. Since the logging goes through
+napi layer, I decided to introduce a client configuration option, that limits the logs going through
```
wprzytula: ⛏️ `option that` (drop the comma)

docs/src/internal/logging.md
```diff
@@ -0,0 +1,43 @@
...
+napi layer - with the performance as the main goal here. While I didn't test the exact performance impact here
+(which can be done in the future - I would recommend it, if you the developer of this driver plan to change this feature),
+I expect it to be quite high, as the rust driver generates quite a lot of logs for a single DB query.
+
+Hover, to allow for the best API compatibility, the default behavior is to pass all logs
+(default: when no log level is provided by the user). This will be desired for users transferring from
```
wprzytula: 🔧 However

docs/src/internal/logging.md
```diff
@@ -0,0 +1,43 @@
...
+DSx driver, but probably not ideal for new users of the driver. For this reason I included warnings,
+that this default behavior may change in the future. The best time for this would probably be somewhere around 1.0.
+
+## Log attaching
+
+In the DSx driver, the logs were generated on the client (so you could listen to logs generated by individual clients).
+Hover, the rust driver does not allow for distinctions between the clients, that generated the log (we only have a single source
```
wprzytula: 🔧 However

wprzytula: 🔧 the clients that

docs/src/internal/logging.md
```diff
@@ -0,0 +1,43 @@
...
+for logs, that does not allow distinguishing what client generated it).
```
wprzytula: `, that` -> `, which`

docs/src/internal/logging.md
```diff
@@ -0,0 +1,43 @@
...
+
+Since the use-cases, where someone will have 2 or more clients seem to be very rare (this is only our assumption),
```
wprzytula: `, where` -> ` where`

docs/src/internal/logging.md
```diff
@@ -0,0 +1,43 @@
...
+
+This has the benefit of keeping the old interface of logging (so logging receiver set up for DSx driver should work here),
+at the cost of duplicating the logs when multiple client sessions are active (every active client will receive single instance
+of a log, no matter what client triggered this action).
+
+Again, this is something that can be reworked during 1.0 release, while keeping the old interface as backward compatible (but probably deprecated).
+
+## Shutdown
+
+We de-attach logging in two cases:
```
wprzytula: `detach`

docs/src/internal/logging.md
```diff
@@ -0,0 +1,43 @@
...
+
+- The client is shut down
+- The client is destroyed
```
wprzytula: 🔧 Please explain what triggers both conditions.

src/logging.rs
```diff
@@ -281,13 +265,6 @@ mod tests {
         assert_eq!(parse_js_level_to_rust("warning"), Some(Level::WARN));
         assert_eq!(parse_js_level_to_rust("error"), Some(Level::ERROR));
     }
-
-    #[test]
-    fn parse_js_level_to_rust_invalid() {
-        assert_eq!(parse_js_level_to_rust("off"), None);
-        assert_eq!(parse_js_level_to_rust("unknown"), None);
-        assert_eq!(parse_js_level_to_rust(""), None);
-    }
```
wprzytula: ❓ Why is this deleted?

src/logging.rs
```diff
@@ -77,40 +77,24 @@ fn parse_js_level_to_rust(level: &str) -> Option<Level> {
...
         } else {
-            if !self.extras.is_empty() {
-                self.extras.push_str(", ");
-            }
-
-            let _ = write!(&mut self.extras, "{}={:?}", field.name(), value);
-        }
-    }
-
-    fn record_str(&mut self, field: &Field, value: &str) {
```
wprzytula: 🔧 Explain these changes in the commit message.

wprzytula (commented):
The commit structure and descriptions are bad. Please order the commits more neatly, describe them better, and make sure commits do not introduce bad code that you fix in a later commit.

lib/client-options.js
adespawn: Yes

src/tests/logging_tests.rs
adespawn: Yup. Napi-rs problem

docs/source/logging/logging.md
adespawn: We initially decided to do WARN: https://github.com/scylladb/nodejs-rs-driver/pull/423#issuecomment-4252742251

but after in person review, you suggested going back to trace, to keep the API compatibility: https://github.com/scylladb/nodejs-rs-driver/pull/423#discussion_r3093921632

src/logging.rs
adespawn: I was cleaning up some LLM generated tests, that are not necessery. Just forgot to put it in proper commits

docs/src/logging.md
adespawn: I make the assumption that the Rust logic for logging will be much faster than corssing FFI layer, and this comment is based on that assumption

docs/source/logging/logging.md
adespawn: Are there any logs generated before?

I don't think so, since the rust client is created only at the connect

test/unit/logging-tests.js
adespawn: JS tests execute one at a time

adespawn: So this shouldn't be a problem

docs/source/logging/logging.md
wprzytula: Entities other than `scylla::Session` can emit logs, too.

wprzytula: Consider execution profiles, policies, etc.

docs/src/logging.md
wprzytula: Still, log filtering on the Rust side has _some_ overhead, not _virtually zero_, right?

docs/source/logging/logging.md
wprzytula: Oh, I see the latter comment. I wrote about using the old default, but what I mainly meant was that we shouldn't set it to OFF by default if the old default was not OFF.

It's shocking that the old driver had logging ON and set to TRACE by default. This means getting flooded with trace messages, right? This must have slowed down the driver a lot. Is it even suitable for production?

After you answer above questions, I think we should choose either INFO or WARN as the new default.

adespawn: > This means getting flooded with trace messages, right

It was the user who was filtering the logs, not the driver

wprzytula: OK, so myriads of logs came to the user and they discarded it. Still a large overhead, right?

test/unit/logging-tests.js
wprzytula: What about the subscriber being initializable only once? Won't this break?

docs/source/logging/logging.md
adespawn: Yes

wprzytula: Understood. We definitely don't want such situation to occur by default.

test/unit/logging-tests.js
adespawn: Not to my understanding

docs/source/logging/logging.md
adespawn: I would say this is a task for a new logging API, that I think would be nice to have for 1.0, not the legacy API

src/logging.rs
adespawn: This was a re-work to make it match to c++. Now this is just the default change, without the middle step

lib/client-options.js
adespawn: > What if a non-string junk object is passed?

Then the typeof check fails

> What if a value of the log level enum is passed?

Then it will be treated as string, since this is what the values of the enums are

wprzytula: You're right. Could you invert the logic, i.e. put negation on the whole condition, so that the inner part means the required (correct) condition?

docs/source/logging/logging.md
wprzytula: OK, then please ensure we have an issue for that, and it mentions the above identified problems.

adespawn: #460

adespawn:
Closing in favour of smaller PRs (#472 , #468, #469, #463)

Issue #73, opened 2024-11-06
Benchmark critical paths in the driver (sending queries and receiving results)


wprzytula:
Let's write benchmarks for critical operations of the driver, to be aware of how fast we are and to gauge impact of further optimisations 
/ detect regressions.

Proposed functionality to be benchmarked:
- INSERTs with and without data,
- SELECTs with little and a lot of data, 
- both prepared and unprepared statements,
- batches,
- TODO - more ideas welcome.

adespawn:
It would  be nice if benchmark would automatically compare state on two branches and generate reports from that.

adespawn:
> INSERTs with a large amount of data

@wprzytula can we consider [ser](https://github.com/scylladb/nodejs-rs-driver/blob/benchmark-serialize/benchmark/logic/ser.js) benchmarks as this kind of benchmark?

adespawn:
Moving this out of 0.3.0 milestone. In this milestone I've added a few new benchmarks (#198 PR). The idea is to keep this issue open, for discussion about any new benchmarks ideas that may come during development. 

If we decide (we can discuss it tomorrow?) that we want some additional benchmarks as part of 0.3.0, we can track those as a sub-issues

PR #425, opened 2026-03-30
Experiment with concurrency levels

adespawn:
Refs: #73 

With planned changes that touch the concurrency, we would want to see the driver performance across different concurrency levels.

This PR introduces a refactor to the current benchmarks, which would allow us to add benchmarks parameterised on the concurrency level easily, and adds those benchmarks.

I would recommend reviewing the first commit in vscode (or similar), to better see what changes and what's kept unchanged

Introduces changes that were generated with LLM.

Example results [with not yet published PR] (x axis is the concurrency level on the benchmark) - I  need to add beter labeling support to the benchmarking tool)

throughput_vs_concurrency_lelel_deser
throughput_vs_concurrency_lelel

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -85,6 +85,10 @@ Each benchmark recreates the table before execution to ensure consistent and iso
 
 Each of the JS benchmarks has an equivalent in Rust. For JS benchmarks, `driver` parameter determines which driver is tested. It can be `scylladb-driver-alpha` or `cassandra-driver`.
 
+For every benchmark you can use default values for either number of queries of concurrency level (supported benchmarks only).
+To use default values in JS benchmarks, provide `default` instead of numerical value for the parameter.
+To use default values in Rust code, do not provide given environment value.
+
 Before running the benchmarks in Rust, remember to build them with flags `--bin -r`.
 
 Concurrent benchmarks in JS first create an array of queries, which are then executed by calling the `executeConcurrent` function.
@@ -92,12 +96,12 @@ Rust benchmarks do not create an array to execute the queries. Queries are gener
 
 - **concurrent insert**
 
-This benchmark uses `executeConcurrent` endpoint to insert `n` rows containing `uuid` and `int` into the database. Afterwards, it checks that the number of rows inserted is correct.
+This benchmark uses `executeConcurrent` endpoint to insert `n` rows containing `uuid` and `int` into the database. Afterwards, it checks that the number of rows inserted is correct. This benchmark can be parametrized by concurrency level.
 
 JS:
 
 ```bash
-node benchmark.js concurrent_insert <driver> <Number of queries>
+node benchmark.js concurrent_insert <driver> <Number of queries> <Concurrency level>
 ```
 
 Rust:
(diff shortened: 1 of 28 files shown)


benchmark/runner-config/config.yml
```diff
@@ -2,7 +2,7 @@ defaults:
   no-steps: 4
   step-progress: 4
   progress-type: multiplicative
-  num-runs: 3
+  num-runs: 2
```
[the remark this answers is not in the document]
adespawn: > Was reducing num-runs from 3 to 2 intentional

No

adespawn:
> The default stepCount values vary widely across modules

Those values were taken from the old python runner, and were determined to be the sweet spot, where the benchmark takes around 1-2 minutes to execute

benchmark/logic/benchmark.js
```diff
@@ -0,0 +1,29 @@
...
+
+const driverName = process.argv[2];
+const benchmarkName = process.argv[3];
+// Each individual benchmark has a default step count defined.
+// This value is chosen in a way that makes the benchmark run for about a 30-90 seconds.
+// This should be the same value as the last step count for the benchmark parametrized by step count (defined in benchmark config).
+// If, at some point (ex. due to driver optimizations), the benchmark start to run faster,
+// it's best to increase the default step count.
+const stepCount = process.argv[4] !== "default" ? parseInt(process.argv[4], 10) : undefined;
+const concurrencyLevel =process.argv[5] !== "default" ? parseInt(process.argv[5], 10) : 100;
```
wprzytula: ♻️ Let's have separate constants: `defaultConcurrencyLevel`, `defaultStepCount` and advise to adjust them as needed. Then, formulas in `concurrencyLevel` will not have to be ever touched.

wprzytula (changes requested):

benchmark/README.md
nikagra: Here and below. Benchmarks in this `README.md` still show `node insert.js <driver> <N>` style. All those scripts are now `module.exports = ...` only, so running them directly either crashes into a missing module error or silently does nothing. The `README` needs updating to the new node `benchmark/logic/benchmark.js <driver> <benchmark-name> [step-count]` form.

benchmark/logic/batch.js
```diff
@@ -1,50 +1,41 @@
 "use strict";
 const async = require("async");
-// Possible values of argv[2] (driver) are scylladb-driver-alpha and cassandra-driver.
-const cassandra = require(process.argv[2]);
 const utils = require("./utils");
```
nikagra, line 3: Running `node benchmark.js` without arguments may  produce:
```
TypeError [ERR_INVALID_ARG_TYPE]: The "id" argument must be of type string. Received undefined
```
`utils.js:7` has `const cassandra = require(process.argv[2])` at module level. `benchmark.js:7` does `require('./utils')` before the validation block. Please check

benchmark/README.md
adespawn: I re-worked documentation, as I totaly forgot about it when updating benchmarks

benchmark/logic/benchmark.js
```diff
@@ -0,0 +1,30 @@
...
+if (!driverName || !benchmarkName) {
+    console.error("Usage: node benchmark.js <driver> <benchmark-name> [step-count] [concurrency]");
+    exit(1);
+}
+
+const cassandra = require(driverName);
+const client = new cassandra.Client(utils.getClientArgs());
+
+const benchmark = require(`./${benchmarkName}`);
+benchmark(cassandra, client, stepCount, concurrencyLevel);
```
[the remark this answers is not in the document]
adespawn: Internal use only, no need to wory

benchmark/logic_rust/common.rs
```diff
@@ -106,13 +106,30 @@ pub(crate) fn get_deser_data() -> (
...
+        .ok()
+        .and_then(|s: String| s.parse::<i32>().ok())
+        .unwrap_or(default)
+}
+
+#[allow(dead_code)]
+pub(crate) fn get_concurrency(default: usize) -> usize {
+    env::var("CONCURRENCY")
+        .ok()
+        .and_then(|s: String| s.parse::<usize>().ok())
```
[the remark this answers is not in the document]
adespawn: Internal use only, not worth spending time on such edgecases

wprzytula (approved):

adespawn:
@nikagra Do you want me to wait for you with merging this PR?

Issue #350, opened 2025-12-10
Transition to TypeScript


adespawn:
The type safety that comes with TypeScript would be a great improvement to the maintainability of this project, but considering the current size, switching all at once is less than feasible. We can still consider the following:

Can we write individual files in TypeScript and mix them with the JS code? 
If so, we can start updating the code from places like the types implementation (which do not depend on other parts of our code) and then slowly move on with updating the other parts of the driver.

adespawn:
https://github.com/scylladb/nodejs-rs-driver/pull/447#discussion_r3309020246

This is related the improving TypeSafety in the transition to TypeScript. ClientOptions (and all other options) can operate in 2 states. One is provided from user, where every field is optional (and as a result every field can be null), and the other is internal state which operates with all options set (so no nulls, only default values). There is some logic in the DSx code that relies on the fact that those fields [are set to something](https://github.com/adespawn/nodejs/blob/2fd529786da88a83d4933d10edcd35575fafb61e/lib/execution-profile.ts#L153), but those places are not annotated by a type that says so, but they use the same type that allows for nullable values

PR #431, opened 2026-04-09
Add support for TS migration with early conversion

adespawn:
This PR adds the configuration necessary for using TS inside the codebase alongside some early conversions from JS to TS.

Created with heavy use of LLMs (especially commit converting JS to TS)

Refs: #350


wprzytula:
Had a quick look over the commits - looks great! :rocket:

tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+        "module": "node16",
+        "moduleResolution": "node16",
+        "allowJs": true,
+        "checkJs": true,
+        "strict": true,
+        "noEmit": true,
+        "esModuleInterop": true,
+        "resolveJsonModule": true
+    },
+    "include": ["lib/**/*.js", "lib/**/*.ts"]
```
nikagra, line 18: Consider adding `exclude` to ensure typecheck never checks emitted JS files (say in case developer runs commands manually)

tsconfig.build.json
```diff
@@ -0,0 +1,15 @@
+{
+    "compilerOptions": {
+        "target": "es2022",
+        "module": "node16",
+        "moduleResolution": "node16",
+        "strict": true,
+        "noImplicitAny": false,
```
nikagra, line 10: Setting `noImplicitAny` to `false` is inconsistent with `tsconfig.json` and defeats the point of using `strict`. Is there a way to get rid of the setting by ensuring TS files are well typed?

src/custom-types.d.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
nikagra, line 1: Making changes to this file contradicts comment below, it may also become overriden by NAPI-RS. Consider addressing issues in some other way

lib/datastax/graph/result-set.js
```diff
@@ -6,7 +6,7 @@ const { throwNotSupported } = require("../../new-utils");
  * @deprecated Not supported by the driver. Usage will throw an error.
  */
 class GraphResultSet {
-    constructor(result, rowParser) {
```
nikagra, line 9: Here and below: why remove parameters in deprecated constructors and functions? This may be an API breaking change. Is it to fix unused param warning?

lib/datastax/index.js
```diff
@@ -1,9 +1,8 @@
 "use strict";
 
 /**
- * DataStax module.
- * 
- * Contains modules and classes to represent functionality that is specific to DataStax products.
+ * This module contains stubs for deprecated types designed for JS users of the dirver.
```
nikagra:
```suggestion
 * This module contains stubs for deprecated types designed for JS users of the driver.
```

lib/geometry/index.js
```diff
@@ -1,8 +1,10 @@
 "use strict";
 
 /**
+ * This module contains stubs for deprecated types designed for JS users of the dirver.
```
nikagra:
```suggestion
 * This module contains stubs for deprecated types designed for JS users of the driver.
```

src/custom-types.d.ts
adespawn: This is the correct place; maybe the comment isn't written perfectly.

This file is a header that is not overwritten by napi-rs, but is taken by napi and placed at the top of index.d.ts, but I see no other way to have different comments at the beginning of this header and the autogenerated file (index.d.ts, not included in the repo)

lib/datastax/graph/result-set.js
adespawn: In JS, having unused trailing parameters vs not having them is indistinguishable from the caller side (maybe excluding the documentation)

So I removed them to fix the unused param warning

tsconfig.build.json
adespawn: This is necessary during the transition period. Otherwise, TS fails to import JS files:

```
lib/types/big-decimal.ts:1:26 - error TS7016: Could not find a declaration file for module './integer.js'. 'wd/scylladb-javascript-driver/lib/types/integer.js' implicitly has an 'any' type.

1 import Integer = require("./integer.js");
```

tsconfig.json
adespawn: Does it fail for you? For me, ts checks work in the current state, with the compiled js files.

tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+        "moduleResolution": "node16",
+        "allowJs": true,
+        "checkJs": true,
+        "strict": true,
+        "noEmit": true,
+        "esModuleInterop": true,
+        "resolveJsonModule": true
+    },
+    "include": ["lib/**/*.js"],
+}
```
wprzytula, line 19: 🔧 Please write comments explaining and justifying each option choice,

lib/datastax/index.js
wprzytula: 📝 Why haven't DSx-specific files been removed yet?

.github/workflows/code-quality.yml
```diff
@@ -40,5 +40,7 @@ jobs:
         run: npx prettier "{lib,examples,test}/**/*.js" --check
       - name: Eslint
         run: npm run eslint
```
wprzytula, line 42: commit `Add tsconfig.build.json and build:ts for in-place TS compilation`:
commit's message doesn't describe the motivations and what's done clearly enough. I don't get it.

tsconfig.build.json
```diff
@@ -0,0 +1,15 @@
...
+        "esModuleInterop": true,
+        "resolveJsonModule": true,
+        "declaration": true,
+        "declarationMap": false,
+        "sourceMap": false
+    },
+    "include": ["lib/**/*.ts"]
+}
```
wprzytula, line 18: ditto: explain and justify choices.

package.json
```diff
@@ -49,7 +49,8 @@
     "typescript": "^5.9.3"
   },
   "scripts": {
-    "build": "napi build --platform --release",
+    "build:ts": "tsc -p tsconfig.build.json",
```
wprzytula, line 52: 🔧 Comment that this needed for the release process.

wprzytula (changes requested):
In general, looks good!

tsconfig.build.json
nikagra: Add comment then to keep track of this providing criteria for enabling the option

lib/datastax/index.js
adespawn: IRC we discussed it last week in person:

TLDR is that we want to loudly inform JS users of the driver of lack of support for DSx features

.github/workflows/code-quality.yml
adespawn: Is it better now?

package.json
adespawn: Since package.json does not allow for comments (despite tsconfig allowing it...) I've added explanation in the commit message

tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+        /* We check both TS and JS with TS compiler */
+        "allowJs": true,
+        "checkJs": true,
+        "strict": true,
+        /* Just checking */
+        "noEmit": true,
+        "esModuleInterop": true,
+        /* Necessary for retrieving driver version from package.json */
+        "resolveJsonModule": true
+    },
```
wprzytula, line 17: 🔧 Not all fields are documented.

tsconfig.build.json
```diff
@@ -0,0 +1,18 @@
...
+        "noImplicitAny": false,
+        "esModuleInterop": true,
+        /* Necessary for retrieving driver version from package.json */
+        "resolveJsonModule": true,
+        /* Useful for using TS code from JS files */
+        "declaration": true,
+        "declarationMap": false,
+    },
+    "include": ["lib/**/*.ts"]
+}
```
wprzytula, line 18: 🔧 Not all dependencies are documented.

wprzytula (changes requested):

tsconfig.json
adespawn: With the self-explanatory strict, if the comments applyes to all fields before the next comment

tsconfig.build.json
adespawn: With the self-explanatory strict, if the comments applyes to all fields before the next comment

tsconfig.json
wprzytula: `declarationMap` require a separate comment.

tsconfig.build.json
wprzytula: `declarationMap` require a separate comment.

tsconfig.json
wprzytula: @nikagra ping.

wprzytula (approved):

PR #432, opened 2026-04-13
New node api utils

adespawn:
This PR adds two new utilities designed to help converting Rust objects into JS:

- HashMaps into named objects
- Enums with values into JS objects

Those types will simplify the code when creating Metadata implementation

### DIFF
--- a/src/tests/to_napi_obj_tests.rs
+++ b/src/tests/to_napi_obj_tests.rs
@@ -1,0 +1,76 @@
+use std::collections::HashMap;
+
+use crate::utils::to_napi_obj::{NamedMap, define_rust_to_js_convertible_object};
+
+#[rustfmt::skip] // fmt splits each field definition into multiple lines
+define_rust_to_js_convertible_object!(
+pub struct Coord {
+    x, x: i32,
+    y, y: i32,
+});
+
+define_rust_to_js_convertible_object!(
+pub enum Color where
+    VariantName: kind,
+{
+    Red = 0,
+    Green = 1,
+    Custom {
+        r, r: i32,
+        g, g: i32,
+        b, b: i32,
+    } = 3,
+});
+
+/// Returns a plain struct produced by the macro.
+/// Expected JS shape: `{ x: 10, y: 20 }`
+#[napi]
+pub fn tests_named_map_struct() -> Coord {
+    Coord { x: 10, y: 20 }
+}
+
+/// Returns a `Color` enum variant produced by the macro.
+/// Cases:
+/// - `0` -> `{ kind: 0 }` (Red)
+/// - `1` -> `{ kind: 1 }` (Green)
+/// - `3` -> `{ kind: 3, r: 128, g: 0, b: 255 }` (Custom)
+#[napi]
+pub fn tests_named_map_enum(case: i32) -> Color {
+    match case {
+        0 => Color::Red,
+        1 => Color::Green,
+        3 => Color::Custom {
+            r: 128,
+            g: 0,
+            b: 255,
+        },
+        _ => unimplemented!("Unknown test case"),
+    }
+}
+
+/// Returns a `NamedMap` of i32 values (V == W, identity `Into`).
+/// Expected JS shape: `{ a: 1, b: 2, c: 3 }`
+#[napi]
+pub fn tests_named_map_i32() -> NamedMap<String, i32> {
+    let mut m = HashMap::new();
+    m.insert("a".to_string(), 1i32);
+    m.insert("b".to_string(), 2i32);
+    m.insert("c".to_string(), 3i32);
+    NamedMap::new(m)
+}
+
+/// Returns a `NamedMap` whose values are converted via `From` (`i32 -> Coord`).
+/// Expected JS shape: `{ origin: { x: 0, y: 0 }, point: { x: 5, y: -3 } }`
+#[napi]
+pub fn tests_named_map_with_conversion() -> NamedMap<String, (i32, i32), Coord> {
+    let mut m = HashMap::new();
+    m.insert("origin".to_string(), (0i32, 0i32));
+    m.insert("point".to_string(), (5i32, -3i32));
+    NamedMap::new(m)
+}
+
+impl From<(i32, i32)> for Coord {
+    fn from((x, y): (i32, i32)) -> Self {
+        Coord { x, y }
+    }
+}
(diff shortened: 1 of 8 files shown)


src/utils/to_napi_obj.rs
```diff
@@ -56,4 +63,63 @@ macro_rules! define_rust_to_js_convertible_object {
...
+    unsafe fn to_napi_value(
+        env: napi::sys::napi_env,
+        val: Self,
+    ) -> napi::Result<napi::sys::napi_value> {
+        let env = Env::from_raw(env);
+        let mut obj = Object::new(&env)?;
+
+        for (key, val) in val.map.into_iter() {
+            obj.set_named_property(&key.to_string(), val.into())?;
+        }
```
wprzytula, line 193: 🤔 You only borrow the `String`. Can we require `K: AsRef<str>` instead of `K: ToString` to elide allocations?

src/utils/to_napi_obj.rs
```diff
@@ -56,4 +63,63 @@ macro_rules! define_rust_to_js_convertible_object {
...
+where
+    // Note, that JS allows objects keyed by something other than strings
+    // For now we do not need such functionality
+    K: ToString,
+    W: ToNapiValue,
+    V: Into<W>,
+{
+    pub(crate) map: HashMap<K, V>,
+    _p: PhantomData<W>,
+}
```
wprzytula, line 159: ❓ Have you tried to use this in your code? Is it convenient?

src/utils/to_napi_obj.rs
```diff
@@ -37,8 +39,34 @@ use napi::{
...
+///     Case2{some_data, someData: bool} = 2
+/// }
+/// ```
+/// Will create the following type:
+/// ```rust
+/// enum ExampleEnum{
+///     Case1,
+///     Case2{some_data: bool}
+/// }
+/// ```
```
wprzytula, line 58: ⛏️ Mising whitespace on the outer side of `{`.

src/utils/to_napi_obj.rs
```diff
@@ -53,6 +81,7 @@ macro_rules! define_rust_to_js_convertible_object {
                 env: ::napi::sys::napi_env,
                 val: Self,
             ) -> ::napi::Result<napi::sys::napi_value> {
+                use ::napi::{JsValue, bindgen_prelude::JsObjectValue};
```
wprzytula, line 84: ❓ Why is this added now? It's only relevant to the struct case, which you haven't modified in this PR.

src/utils/to_napi_obj.rs
```diff
@@ -61,6 +90,51 @@ macro_rules! define_rust_to_js_convertible_object {
...
     };
+    (enum $enum_name:ident
+        where VariantName: $variant_val_name: ident$(,)?
+        {
+        $($variant_name:ident
+            $( { $(
+                $field_name:ident, $js_name:ident: $field_type:ty),*,
+            } )?
+        = $val: literal),*,}) => {
+        pub enum $enum_name{
```
wprzytula: ⛏️ Missing whitespace before `{`.

wprzytula: 💭 Maybe let's require specifying `pub enum`, to not be surprised by implicit visibility promotion to `pub`?

src/tests/to_napi_obj_tests.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::utils::to_napi_obj::{NamedMap, define_rust_to_js_convertible_object};
+
+#[rustfmt::skip] // fmt splits each field definition into multiple lines
+define_rust_to_js_convertible_object!(Coord {
+    x, x: i32,
+    y, y: i32,
+});
+
+define_rust_to_js_convertible_object!(
+    enum Color
```
wprzytula: 🔧 After you start supporting enums with `enum $name` syntax, let's make struct support analogous: `struct $name` instead of plain `$name`.

wprzytula (changes requested):

src/utils/to_napi_obj.rs
adespawn: I have a draft for metadata implementation, where it's used and provides quite a significant reduction in the code:

```rust
define_rust_to_js_convertible_object!(
    TableWrapper {
        columns, columns: NamedMap<String, Column, ColumnWrapper>,
        partition_key, partitionKey: Vec<String>,
        clustering_key, clusteringKey: Vec<String>,
        partitioner, partitioner: Option<String>,
    }
);

impl From<Table> for TableWrapper {
    fn from(t: Table) -> Self {
        TableWrapper {
            columns: NamedMap::new(t.columns),
            partition_key: t.partition_key,
            clustering_key: t.clustering_key,
            partitioner: t.partitioner,
        }
    }
}
```

In that single draft, it's used for 6 different fields, and without this macro we would need to manually implement ToNapiValue trait for the whole object (`TableWrapper` in this case)

src/utils/to_napi_obj.rs
adespawn: All napi stuff is public, including the structures from the same macro. I can add this keyword, for better code readability, but for me, this visibility promotion is obvious

wprzytula: Let's add the `pub` keyword then.

src/utils/to_napi_obj.rs
adespawn: I moved it to more appropriate commit

src/utils/to_napi_obj.rs
```diff
@@ -30,8 +32,34 @@
...
 /// ```
+///
+/// Similarly when creating an enum in a following way:
+/// ```rust
+/// define_rust_to_js_convertible_object!(
+/// enum ExampleEnum
+/// where VariantName: kind {
+///     Case1 = 1,
+///     Case2{some_data, someData: bool} = 2
+/// }
```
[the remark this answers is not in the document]
nikagra: @adespawn I think you need into this comment again, even if it is only doc related:
- Missing closing `);` for the macro call 
- `Case2{...} = 2` is missing a trailing comma (the macro pattern requires `),*,`)
-  Inside the variant, `someData: bool` needs a trailing comma (pattern is `$field_type:ty),*,`)

nikagra (commented):
Please, revisit Copilot copilot comment I've un-resolved. Looks good otherwise

wprzytula:
Please ask me for review after you resolve @nikagra's comments.

wprzytula (approved):

PR #437, opened 2026-04-16
docs: update theme 1.9.2

dgarcia360:
Updates docs theme to 1.9.2.

CHANGELOG: https://github.com/scylladb/sphinx-scylladb-theme/blob/master/docs/source/upgrade/CHANGELOG.md#192---14-april-2026


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "uv"
+    directory: "/docs"
+    schedule:
+      interval: "daily"
+    allow:
+      - dependency-name: "sphinx-scylladb-theme"
+      - dependency-name: "sphinx-multiversion-scylla"
```
adespawn: What is this change...?

dgarcia360: Configures dependabot so that we don't have to send these PRs manually next time. 

Most likely we missed this file when installing the docs project: https://sphinx-theme.scylladb.com/stable/getting-started/installation.html#step-1-download-the-sample-project

adespawn: With the recent internal push towards fixing dependency version (like #430, DRIVER-515), I'm not sure we want to have this kind of automatic updates.

Any thoughts @wprzytula / @dkropachev / @nikagra ?

wprzytula: I'm not sure. @dkropachev may know more about the dependency version fixing effort.

dkropachev: There are two problem:
1. We need to start using commit hashes for any deps that uses git tags, including GitHub actions 
2. New dep version should be 3-6 months old

If we can make dependabot do that - great, if not, then we need to figure out something else.

adespawn: Here we are talking about pip versions, which have their own caveats:
https://news.ycombinator.com/item?id=42418299 ...

> New dep version should be 3-6 months old

Since those dependencies are released by (I think?) someone at Scylla, I'm not sure we want to have such strict requirements here.

adespawn: Ok, since I misunderstood how this depenbot works (it opens PR with an update, rather than automatically pushing to main), my concerns are a bit smaller. However, to comply with the general idea of pinning, @dgarcia360 could you pin dependencies to a specific version (so no ">="), and ensure dependpot does the same in it's updates?

dgarcia360: The >= specifiers in `pyproject.toml` are intentional. They define the minimum compatible version, while [uv.lock](https://github.com/scylladb/nodejs-rs-driver/blob/main/docs/uv.lock) is committed to the repo and handles the actual pinning to exact versions. That's the standard pattern with lock-file-based workflows: the project file declares compatibility, the lock file guarantees reproducibility.

Dependabot will open PRs that update `uv.lock` to newer versions within the allowed range. Nothing lands without review. The alternative is opening those same PRs by hand whenever a new release of a scylladb dependency comes out.

adespawn: @dgarcia360 Could you only fix the missing new line. Once that's done I will merge this PR

dgarcia360: @adespawn done

adespawn (approved):

PR #441, opened 2026-04-23
Update milestone-sync workflow

dkropachev:
Update it to version that pins all the dependencies by commit hash


adespawn:
I don't think this solves the problem of using separate organization we discussed internally

dkropachev:
> I don't think this solves the problem of using separate organization we discussed internally

not sure what you are talking about

adespawn (approved):

PR #444, opened 2026-05-11
fix: broken links

dgarcia360:
Fixes broken links detected by our new link checker: https://github.com/scylladb/scylladb-docs-homepage/pull/169

- Adds missing 404 page configuration
- Adds `redirects.yml` configuration
- Fixes an issue with JSDoc 404 which mis-parses private methods and emits an empty <a href=""></a> after every type token.


dgarcia360:
cc/ @annastuchlik

docs/source/conf.py
```diff
@@ -66,9 +66,23 @@
     ("globals", "Global functions and constants."),
 ]
 
+# -- Options for not found extension
+
+# Template used to render the 404.html generated by this extension.
+notfound_template =  '404.html'
+
+# Prefix added to all the URLs generated in the 404 page.
+notfound_urls_prefix = ''
```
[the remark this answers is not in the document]
dgarcia360: Handled by the theme already.

docs/source/conf.py
```diff
@@ -66,9 +66,23 @@
...
+
+
 # -- Options for myst parser ----------------------------------------
 myst_enable_extensions = ["colon_fence"]
 
+# -- Options for redirect extension
+# Read a YAML dictionary of redirections and generate an HTML file for each
+redirects_file = "_utils/redirects.yaml"
```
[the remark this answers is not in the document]
dgarcia360: It is ok.

docs/_extensions/jsdoc_content.py
```diff
@@ -369,6 +369,17 @@ def _process_dl(m):
...
 def _strip_noise(html, github_source_url=None):
     """Clean up JSDoc HTML noise."""
+    # JSDoc 4.0.4 mis-parses ES2022 ``#private`` methods and emits an
+    # extra ``:<a href=""></a>`` after every type token in a generic
+    # (e.g. ``Array:<a></a>.<number:<a></a>>``). Strip the pair.
+    html = re.sub(
+        r':<a\s+href="[^"]*"[^>]*>\s*</a>',
+        '',
+        html,
+    )
```
adespawn: Can you link to example of that in the current documentation?

dgarcia360: For example, see ``bits`` type in https://nodejs-rs-driver.docs.scylladb.com/stable/api/types/Integer.html#constructor

image

adespawn: But won't this regex also strip valid hrefs: 
```html
<span class="type-signature"> :<a href="../globals.html#Integer">Integer</a></span>
```
https://nodejs-rs-driver.docs.scylladb.com/stable/api/types/Integer.html#members
?

dgarcia360: Good catch worth checking, but the regex is safe here. :

image

The pattern only matches the broken empty-anchor pairs JSDoc emits like `:<a href=""></a>`, the trailing`\s*</a>` requires the anchor to be empty. 

The example `:<a href="../globals.html#Integer">Integer</a>` has `Integer` as content, so it won't be stripped.

adespawn: Okey. I got misled by the `"[^"]*"` part of the regex, while missing that we match for withspace characters inside the link tag

adespawn: > e.g. ``Array:<a></a>.<number:<a></a>>``

But I'm not sure this example makes sense to me, since we do not href in it

dgarcia360: See `bits` type for instance in https://nodejs-rs-driver.docs.scylladb.com/stable/api/types/Integer.html#constructor

JSDoc generates empty links:

image

adespawn: I understand that. I just don't understand the in code comment you wrote

dgarcia360: Edited the comment, let me know if it's clearer now: https://github.com/scylladb/nodejs-rs-driver/pull/444/changes/ac796fdac0f64d5aafecf440fae8662b0d579195#diff-d69c8e25865a410add30be36751fbb74b9ca0b274dbd80d3407c293cad4242adR372

adespawn: That's much better. Thanks!

annastuchlik:
@adespawn Could you review and merge this PR now? We need these updates in place to be able to automatically run link checks on the documentation for this project.

adespawn (approved):

PR #447, opened 2026-05-14
Convert another file to TS

adespawn:
In this pr execution profiles get TSed

### DIFF
--- a/lib/execution-profile.js
+++ b/lib/execution-profile.js
@@ -49,7 +49,7 @@ class ExecutionProfile {
         this.name = name;
         /**
          * Consistency level.
-         * @type {Number}
+         * @type {Number | undefined}
          */
         this.consistency = options.consistency;
         /**
@@ -59,7 +59,7 @@ class ExecutionProfile {
         this.loadBalancing = options.loadBalancing;
         /**
          * Client read timeout.
-         * @type {Number}
+         * @type {Number| undefined}
          */
         this.readTimeout = options.readTimeout;
         /**
@@ -69,10 +69,11 @@ class ExecutionProfile {
         this.retry = options.retry;
         /**
          * Serial consistency level.
-         * @type {Number}
+         * @type {Number | undefined}
          */
         this.serialConsistency = options.serialConsistency;
 
+        // @ts-ignore - This is an explicit check for the use of deprecated option
         if (options.graphOptions !== undefined) {
             // This option was present in the DSx driver, but is no longer relevant.
             // We explicitly check for it to inform users using this options,
@@ -95,12 +96,18 @@ class ProfileManager {
         this._defaultConfiguredRetryPolicy = undefined;
         this._setDefault(options);
         // A array of unique load balancing policies
+        /**
+         * @type {any[]}
+         */
         this._loadBalancingPolicies = [];
         // A dictionary of name keys and profile values
+        /**
+         * @type {Record<string | number, any>}
+         */
         this._profilesMap = {};
         // A dictionary of name keys and custom payload dictionaries as values
         this._customPayloadCache = {};
-        this._profiles.forEach(function (p) {
+        this._profiles.forEach( (/** @type {{ name: string | number; loadBalancing: any; }} */ p) => {
             this._profilesMap[p.name] = p;
             // Set required properties
             p.loadBalancing =
@@ -110,8 +117,7 @@ class ProfileManager {
             if (this._loadBalancingPolicies.indexOf(p.loadBalancing) === -1) {
                 this._loadBalancingPolicies.push(p.loadBalancing);
             }
-            return p;
-        }, this);
+        });
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lib/execution-profile.ts
```diff
@@ -0,0 +1,174 @@
...
+        this.retry = opts.retry;
+        this.serialConsistency = opts.serialConsistency;
+
+        // @ts-ignore - This is an explicit check for the use of deprecated option
+        if (opts.graphOptions !== undefined) {
+            // This option was present in the DSx driver, but is no longer relevant.
+            // We explicitly check for it to inform users using this options,
+            // to avoid any confusion if the user code depends on this configuration.
+            newUtils.throwNotSupported("Graph options");
+        }
```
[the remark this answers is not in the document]
adespawn: Avoiding documentation for graphOptions removes noise for new users

lib/execution-profile.js
wprzytula: ❓ Could this attribute be removed already in the first commit?

lib/execution-profile.ts
```diff
@@ -88,56 +80,49 @@ class ExecutionProfile {
  * @ignore
  */
 class ProfileManager {
-    /**
-     * @param {ClientOptions} options
-     */
-    constructor(options) {
+    _profiles: ExecutionProfile[];
+    _defaultConfiguredRetryPolicy: RetryPolicy | undefined;
+    _defaultProfile!: ExecutionProfile;
```
wprzytula: ❓ Why ` _defaultProfile!` - with exclamation mark?

lib/execution-profile.js
adespawn: No, since we get import errors on JS docs

> The remaining issues arise from
lacks of imports for js docs

lib/execution-profile.ts
adespawn: It marks that the value [cannot be null](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator). Since DSx code sets this value in a convoluted way in setDefault, this is necessary with the current code state, but it should be possible to refactor the code to not need this (ex. setDefault returns a value to set for the default, rather than setting it directly).

wprzytula: I understand what it means when `!` is used a postfix for a variable used in an expression.
But what does it mean when _a field definition_ is decorated with the `!`?

adespawn: Means that the field will not be null

wprzytula: Why then not use a non-nullable field type instead?

adespawn: As it will never be null, since it's always set in setDefault

wprzytula: I don't understand.
You're saying that a field will never be null.
Then why not make it have a non nullable type? TS has `NonNullable<T>` generic or sth like this unless I'm mistaken. Currently, it has just `ExecutionProfile` type. Why not `NonNullable<ExecutionProfile>`?

adespawn: The refactor I mentioned managed to eliminate that `!`

lib/execution-profile.ts
```diff
@@ -146,47 +123,49 @@ class ProfileManager {
...
-            this._defaultProfile.loadBalancing ||
+        defaultProfile.loadBalancing =
+            defaultProfile.loadBalancing ||
             options.policies?.loadBalancing;
-        this._defaultProfile.retry =
-            this._defaultProfile.retry || options.policies?.retry;
+        defaultProfile.retry =
+            defaultProfile.retry || options.policies?.retry;
+
+        return defaultProfile;
```
wprzytula: 🤔 The name `get*` suggests idempotence. This method actually sets a number of things, so it has serious side effects. Maybe `set_and_get`?

wprzytula (changes requested):

lib/execution-profile.ts
```diff
@@ -0,0 +1,175 @@
...
+        this.#profilesMap = {};
+        this.#profiles.forEach((p) => {
+            this.#profilesMap[p.name] = p;
+            // Set required properties
+            p.loadBalancing =
+                p.loadBalancing || this.#defaultProfile.loadBalancing;
+            // Using array indexOf is not very efficient (O(n)) but the amount of profiles should be limited
+            // and a handful of load-balancing policies (no hashcode for load-Balancing policies)
+            if (this.#loadBalancingPolicies.indexOf(p.loadBalancing!) === -1) {
+                this.#loadBalancingPolicies.push(p.loadBalancing!);
```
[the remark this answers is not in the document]
adespawn: Aghr. This is both valid and invalid. While you take just the type of the value into account, then yes, it may be undefined. HOWEVER, the client options that will be passed to ProfileManager will be extended on the default options, meaning they will have default LB policy set.

The best way to fix that would be to somehow extend the type to allow for such distinction.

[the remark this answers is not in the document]
adespawn: Outiside of scope for this PR. Probably need to keep it in mind in some issue

wprzytula (approved):

PR #449, opened 2026-05-25
Fix release process

adespawn:
There was a bug in the release process of 0.5.0, due to recent TS changes. This PR fixes those problems.

Additionaly I bump the version from 0.5.0 to 0.5.1, since I had to remove the 0.5.0 release, due to it's incorrect state.
0.5.1 was succesfully release and is working correctly now

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,12 +93,12 @@ jobs:
           node-version: 24
       - name: Install dependencies
         run: npm i
-      - name: Build TypeScript
-        run: npm run build:ts
       - name: Download artifact
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           merge-multiple: true
+      - name: Build TypeScript
+        run: npm run build:ts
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -146,6 +144,12 @@ jobs:
         with:
           path: artifacts
           merge-multiple: true
+      - name: Download all artifacts v2
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          merge-multiple: true
+      - name: Build TypeScript
+        run: npm run build:ts
```
wprzytula: ❓ Why `v2`?

wprzytula: Can you write comments that explain what is happening?
1. On each step.
2. Big Picture view before the job.

wprzytula (changes requested):

.github/workflows/release.yml
adespawn: > Why v2?

Well, this was a hacky way of fixing it. It's done more properly now

wprzytula (approved):

PR #454, opened 2026-05-28
Improve (Cql) type type safety

adespawn:
This PR replaces the uses of the magic `@type {type}` with a proper type information.
This includes some additional TypeScript code

Fixes #426 
Progress to #350


lib/types/column-info.ts
```diff
@@ -0,0 +1,66 @@
+import { CqlType } from "../../index";
+
+export interface ColumnInfoOptions {
+  frozen?: boolean;
+  reversed?: boolean;
+}
```
wprzytula: ❓ Why are these optional fields?

lib/types/column-info.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export interface UdtField {
+  name: string;
+  type: ColumnInfo;
+}
+
+export interface UdtInfo {
+  name: string;
+  fields: UdtField[];
+}
```
wprzytula: 🔧 Missing docstrings.

wprzytula: 💭 Isn't more convenient and efficient to store UDT fields as an object instead of an array?

lib/types/column-info.ts
```diff
@@ -0,0 +1,66 @@
...
+
+// ColumnInfo as a class is based on this monstrosity of a type
+// Current type does not restrict the info type based on the code.
+// For now this is not a problem, but this may come up when converting encoder to TS.
+// /**
+//  * @typedef {(singleTypeNames<keyof singleTypeNames>| types.dataTypes.duration | types.dataTypes.text)} SingleTypeCodes
+//  * @typedef {{code : SingleTypeCodes, info?: null, options? : {frozen?:boolean, reversed?:boolean} }} SingleColumnInfo
+//  * @typedef {{code : (types.dataTypes.map), info : [ColumnInfo, ColumnInfo], options?: {frozen?: Boolean, reversed?: Boolean}}} MapColumnInfo
+//  * @typedef {{code : (types.dataTypes.tuple), info : Array<ColumnInfo>, options?: {frozen?: Boolean, reversed?: Boolean}}} TupleColumnInfo
+//  * @typedef {{code : (types.dataTypes.tuple | types.dataTypes.list)}} TupleListColumnInfoWithoutSubtype TODO: guessDataType can return null on tuple/list info, why?
```
wprzytula: ❓ Is the TODO pre-existing?
🔧 Can you address it?

lib/types/column-info.ts
```diff
@@ -0,0 +1,66 @@
...
+  options?: ColumnInfoOptions;
+  /** Only present for vector custom types */
+  customTypeName?: string;
+
+  constructor(
+    code: CqlType,
+  ) {
+    this.code = code;
+    this.info = null;
+  }
```
wprzytula: ❓ Why does the constructor always set `info` to `null`?

lib/encoder.js
wprzytula, line 2289: ❓ Were it we who marked this API as experimental, or was it so before?

lib/types/cql-utils.ts
```diff
@@ -0,0 +1,77 @@
...
+): Array<unknown> {
+    if (expectedTypes.length == 0 && !params) return [];
+    const res: Array<unknown> = [];
+    for (let i = 0; i < params.length; i++) {
+        const tmp = encoder.encode(params[i], expectedTypes[i]);
+        res.push(tmp);
+    }
+    return res;
+}
+
```
wprzytula, line 108: ❓ What does this do? Missing docstring.

wprzytula (changes requested):
I appreciate the efforts to split commits in a way that the history is clearer and readable! 🚀

lib/types/column-info.ts
adespawn: > Is the TODO pre-existing?

Yes

> Can you address it?

I can take a look, but I would say it's something minor

lib/types/column-info.ts
adespawn: Backward compatibility: DSx had it like this and with this PR I try to avoid breaking changes

lib/types/column-info.ts
adespawn: Probably yes, but again: Backward compatibility

lib/encoder.js
adespawn: It was here before. We just dropped it from the public API, never chaning the comment

wprzytula: If it's not part of the public API, then please remove this misleading comment.

lib/types/column-info.ts
wprzytula: Ah, it's public API. I never grasped how to tell it from private API in this driver. What's the way to do it?

lib/types/column-info.ts
wprzytula: OK.

lib/types/column-info.ts
adespawn: It's minor part of public API, but quite important part of encoder internal API. 

When it comes to distinguishing the API... Some way would be to look at a docs. In general it *should* expose all and only public API. But this is based on a few assumptions.

lib/types/column-info.ts
adespawn: > Why does (it) always set info

To satisfy [TS compilator](https://www.typescriptlang.org/tsconfig/#strictPropertyInitialization): 

> Why does (it) set info to null

To match maybe a bit awkward way the ColumnInfo is constructed

wprzytula: > > Why does (it) set info to null
> 
> To match maybe a bit awkward way the ColumnInfo is constructed

Can we construct `ColumnInfo` fully at once? So that `constructor` is truly a _constructor_.

lib/types/cql-utils.ts
```diff
@@ -0,0 +1,74 @@
+import { CqlType } from "../../index";
+
+export interface ColumnInfoOptions {
+    /** This flag is only used in schema metadata. 
+     * For prepared statement bind markers and query result
+     * types those fields will always be set to `false` (even if the DB column
+     * corresponding to given marker / result type is frozen).  */
+    frozen?: boolean;
+    reversed?: boolean;
+}
```
wprzytula, line 17: 1. Missing docstring for the whole interface.
2. Missing docstring for `reversed` field.

wprzytula (changes requested):

lib/types/column-info.ts
```diff
@@ -0,0 +1,77 @@
+import { CqlType } from "../../index";
+
+/** Wrapper over frozen field. This wrapper is present to make this interface backward compatible. */
+export interface ColumnInfoOptions {
```
wprzytula: 🤔
1. _Wrapper over frozen field_ ?? So why is there also the `reversed` field?
2. In what way would absence of this wrapper break backward compatibility?

lib/types/cql-utils.ts
```diff
@@ -0,0 +1,77 @@
...
+/** Wrapper over frozen field. This wrapper is present to make this interface backward compatible. */
+export interface ColumnInfoOptions {
+    /** This flag is only used in schema metadata. 
+     * For prepared statement bind markers and query result
+     * types those fields will always be set to `false` (even if the DB column
+     * corresponding to given marker / result type is frozen).  */
+    frozen?: boolean;
+    // This is a legacy field that was present in the DSx code.
+    // It's likely we will want to remove this field at some point
+    reversed?: boolean;
```
wprzytula, line 16: 🔧 Missing explanation:
1. What's the semantics of the field in the DSx code?
2. Is this field used by current implementation?

lib/types/column-info.ts
adespawn: Currently when someone wants to access frozen field, they write: `smth.options.frozen`. This wrapper keeps this semantic. Without it we would write: `smth.frozen`

wprzytula: The notion _wrapper_ is IMO misleading in this case. This is a grouping of options. Now that we removed support for `reversed` option, the group is a singleton. It's still not a _wrapper_ in how I understand this.

wprzytula (approved):

PR #459, opened 2026-06-03
Enable already supported tests

adespawn:


test/integration/supported/client-execute-tests.js
```diff
@@ -2170,7 +2168,7 @@ describe("Client @SERVER_API", function () {
         numericTests(keyspace, false);
         pagingTests(keyspace, false);
 
-        // No support for keyspace
+        // No support for local table
```
wprzytula, line 2171: ❓ What's _local table_?

adespawn: `select * from local`

wprzytula: This doesn't explain anything to me. What's such table?

adespawn: No clue. This is a DSx test that fails with: `Database returned an error: The query is syntactically correct but invalid, Error message: unconfigured table local` on a ccm + cassandra...

I also cannot say that I fully understand this test...

wprzytula (approved):

PR #463, opened 2026-06-08
Logging infrastructure

adespawn:
This PR extracts the changes from #423. More specificly, this PR constains changes that introduce logging infrastructure (1st commit) split into 3.

Why?
The initial PR has grown a bit to much, and I believe it would be just faster to split it into smaller chunks to get it done faster.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -62,6 +62,7 @@ class Client extends events.EventEmitter {
      */
     rustClient;
     #encoder;
+    #loggingId;
     /**
      * Creates a new instance of {@link Client}.
      * @param {clientOptions.ClientOptions} options The options for this instance.
@@ -236,6 +237,15 @@ class Client extends events.EventEmitter {
         );
 
         try {
+            {
+                const emitter = this.emit.bind(this, "log");
+                this.#loggingId = rust.setupLogging(
+                    emitter,
+                    // This is only a temporary change. In the following PRs we will configuration for logging.
+                    types.logLevels.warning,
+                );
+            }
+
             this.rustClient = await rust.SessionWrapper.createSession(
                 this.rustOptions,
             );
@@ -244,6 +254,7 @@ class Client extends events.EventEmitter {
             this.connected = false;
             this.connecting = false;
             this.emit("connected", err);
+            this.#closeLogging();
             throw err;
         }
 
@@ -880,6 +891,15 @@ class Client extends events.EventEmitter {
 
         this.connected = false;
         this.isShuttingDown = true;
+
+        this.#closeLogging();
+    }
+
+    #closeLogging() {
+        if (this.#loggingId !== undefined) {
+            rust.removeLogging(this.#loggingId);
+            this.#loggingId = undefined;
+        }
     }
 
     /**
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -1,0 +1,49 @@
+use tracing::Level;
+
+/// Maps a Rust `tracing::Level` to the JS log-level strings.
+pub(crate) fn rust_level_to_js(level: &Level) -> &'static str {
+    match *level {
+        Level::TRACE => "trace",
+        Level::DEBUG => "debug",
+        Level::INFO => "info",
+        Level::WARN => "warning",
+        Level::ERROR => "error",
+    }
+}
+
+/// Parses a JS-side log-level name.
+/// Note: "off" should be filtered at the JS side and should not be passed to Rust.
+pub(crate) fn parse_js_level_to_rust(level: &str) -> Option<Level> {
+    match level {
+        "trace" => Some(Level::TRACE),
+        "debug" => Some(Level::DEBUG),
+        "info" => Some(Level::INFO),
+        "warning" => Some(Level::WARN),
+        "error" => Some(Level::ERROR),
+        _ => None,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn rust_level_to_js_mapping() {
+        assert_eq!(rust_level_to_js(&Level::TRACE), "trace");
+        assert_eq!(rust_level_to_js(&Level::DEBUG), "debug");
+        assert_eq!(rust_level_to_js(&Level::INFO), "info");
+        assert_eq!(rust_level_to_js(&Level::WARN), "warning");
+        assert_eq!(rust_level_to_js(&Level::ERROR), "error");
+    }
+
+    #[test]
+    fn parse_js_level_to_rust_valid() {
+        assert_eq!(parse_js_level_to_rust("trace"), Some(Level::TRACE));
+        assert_eq!(parse_js_level_to_rust("debug"), Some(Level::DEBUG));
+        assert_eq!(parse_js_level_to_rust("info"), Some(Level::INFO));
+        assert_eq!(parse_js_level_to_rust("warning"), Some(Level::WARN));
+        assert_eq!(parse_js_level_to_rust("error"), Some(Level::ERROR));
+        assert_eq!(parse_js_level_to_rust("some garbage"), None);
+    }
+}
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -1,9 +1,12 @@
+use std::fmt;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{OnceLock, RwLock};
 
 use napi::bindgen_prelude::FnArgs;
 use napi::threadsafe_function::{ThreadsafeFunction, ThreadsafeFunctionCallMode};
+use std::fmt::Write;
 use tracing::Level;
+use tracing::field::{Field, Visit};
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::{Layer, Registry};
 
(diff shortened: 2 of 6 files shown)


src/logging.rs
```diff
@@ -0,0 +1,226 @@
...
+/// subscriber the very first time this is called.
+fn get_or_init_callbacks() -> &'static RwLock<Vec<RegisteredCallback>> {
+    CALLBACKS.get_or_init(|| {
+        let subscriber = Registry::default().with(JsForwardingLayer);
+        tracing::subscriber::set_global_default(subscriber).unwrap_or_else(|e| {
+            // This will fail only if we try to call `set_global_default` in other place in the code.
+            // Since we are under OnceLock here, we will not call this function here twice.
+            // For this reason we can safely panic here.
+            panic!("This is likely due to a bug in the driver: failed to initialized logger. {e}");
+        });
```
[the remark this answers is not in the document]
adespawn: For some reason I explained why we can panic here..

lib/client.js
```diff
@@ -235,6 +236,17 @@ class Client extends events.EventEmitter {
...
 
+        {
+            const emitter = this.emit.bind(this);
+            this.#loggingId = rust.setupLogging(
+                (level, target, message, furtherInfo) => {
+                    emitter("log", level, target, message, furtherInfo);
+                },
+                // This is only a temporary change. In the following PRs we will configuration for logging.
+                types.logLevels.warning,
+            );
```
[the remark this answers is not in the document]
adespawn: ??? This is not introduced in this PR. This is just some temp value...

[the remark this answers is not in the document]
adespawn: Ah, nvm. Vscode hid this file

src/logging.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    #[test]
+    fn parse_js_level_to_rust_valid() {
+        assert_eq!(parse_js_level_to_rust("trace"), Some(Level::TRACE));
+        assert_eq!(parse_js_level_to_rust("debug"), Some(Level::DEBUG));
+        assert_eq!(parse_js_level_to_rust("info"), Some(Level::INFO));
+        assert_eq!(parse_js_level_to_rust("warning"), Some(Level::WARN));
+        assert_eq!(parse_js_level_to_rust("error"), Some(Level::ERROR));
+    }
+}
```
wprzytula, line 237: 🔧 The test should also make sure that invalid input is handled gracefully (not with a panic).

lib/client.js
```diff
@@ -235,6 +236,17 @@ class Client extends events.EventEmitter {
             undefined,
         );
 
+        {
+            const emitter = this.emit.bind(this);
+            this.#loggingId = rust.setupLogging(
+                (level, target, message, furtherInfo) => {
+                    emitter("log", level, target, message, furtherInfo);
+                },
```
wprzytula: 💭 I see you bind `this` to `this.emit`. Why not bind `"log"` too, so that `emitter` can be passed directly as the closure argument to `setupLogging`?

src/logging.rs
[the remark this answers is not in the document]
wprzytula: Copilot is right - it's possible that another Rust library has already initialized the subscriber. In such case we should probably error out, not panic.

src/logging.rs
```diff
@@ -1,4 +1,59 @@
...
+use crate::errors::{
+    ConvertedError, ConvertedResult, JsResult, make_js_error, with_custom_error_sync,
+};
+
+/// Shorthand for the NAPI callback type.
+///
+/// Generic parameters: `CalleeHandled = false`, `Weak = true` so that the
+/// callback does not prevent the Node.js event-loop from exiting.
+type LogCallback = ThreadsafeFunction<
+    /* T: */ FnArgs<(String, String, String, String)>,
```
wprzytula, line 22: 🔧 Please write semantics of these `String`s in a comment.

src/logging.rs
```diff
@@ -24,6 +79,101 @@ pub(crate) fn parse_js_level_to_rust(level: &str) -> Option<Level> {
...
+        let meta = event.metadata();
+        let event_level = meta.level();
+
+        // Quick pre-check: is *any* callback interested in this level?
+        // Level ordering in tracing: TRACE > DEBUG > INFO > WARN > ERROR
+        // (more verbose = greater).  An event passes if it is at most as
+        // verbose as the callback's threshold, i.e. event_level <= min_level.
+        if !callbacks.iter().any(|cb| *event_level <= cb.min_level) {
+            return;
+        }
```
wprzytula: ⛏️ This part is redundant:
```rs
if callbacks.is_empty() {
    return;
}
```
because of this part:
```rs
if !callbacks.iter().any(|cb| *event_level <= cb.min_level) {
    return;
}
```

src/logging.rs
```diff
@@ -24,6 +79,101 @@ pub(crate) fn parse_js_level_to_rust(level: &str) -> Option<Level> {
...
+        for cb in callbacks.iter() {
+            if *event_level <= cb.min_level {
+                cb.callback.call(
+                    FnArgs {
+                        data: (
+                            level_str.to_owned(),
+                            target.to_owned(),
+                            message.to_owned(),
+                            "".to_owned(),
+                        ),
```
wprzytula, line 164: ❓ What's the empty string here?

src/logging.rs
```diff
@@ -24,6 +79,101 @@ pub(crate) fn parse_js_level_to_rust(level: &str) -> Option<Level> {
...
+/// After this call the callback associated with `id` will no longer receive
+/// any tracing events.  If `id` does not match any registered callback the
+/// call is a silent no-op.
+#[napi]
+pub fn remove_logging(id: i64) -> Option<bool> {
+    if let Some(lock) = CALLBACKS.get() {
+        lock.write().ok()?.retain(|cb| cb.id != id as u64);
+    }
+    Some(true)
+}
```
wprzytula, line 213: 🤔 This always returns either `None` or `Some(true)`.

wprzytula, line 213: ❓ Why does `setup_logging` return `JsResult` and `remove_logging` return `Option`?

src/logging.rs
wprzytula: 🔧 Aha, it's a message that is going to be populated in the next commit. In such case:
1. add a TODO.
2. write about this clearly in the commit message.

Otherwise, I'd expect the commits to be purely incremental, and here they're not: you introduce something and then change it without prior notice.

src/logging.rs
```diff
@@ -79,6 +82,29 @@ pub(crate) fn parse_js_level_to_rust(level: &str) -> Option<Level> {
...
+
+impl MessageVisitor {
+    fn new() -> Self {
+        Self {
+            log_message: String::new(),
+        }
+    }
+}
+
+// Collects all fields and values in a single log event into a single String, similarly to C++-rs driver
```
wprzytula: ⛏️ CPP RS Driver

src/logging.rs
```diff
@@ -79,6 +82,29 @@ pub(crate) fn parse_js_level_to_rust(level: &str) -> Option<Level> {
...
+impl Visit for MessageVisitor {
+    fn record_debug(&mut self, field: &Field, value: &dyn fmt::Debug) {
+        if self.log_message.is_empty() {
+            write!(self.log_message, "{field}: {value:?}").unwrap();
+        } else {
+            write!(self.log_message, ", {field}: {value:?}").unwrap();
+        }
+    }
+}
+
```
wprzytula: Please look at C# RS implementation of logging, as it improved compared to CPP RS approach: https://github.com/scylladb-zpp-2025-csharp-rs-driver/csharp-driver/blob/master/rust/src/logging.rs

src/logging.rs
wprzytula: No, it's not a message. Message is `String::new()`; what's the `""`?

wprzytula (changes requested):

src/logging.rs
adespawn: While any failure in the setup of the logger justifies throwing a JS error. Compared to this in the cleanup I believe it's better to give the cleanup process to finish, and just inform user through a log msg that something wrong had happen (lower severity than an error
)

adespawn: To makie it easier:

 -   to use ? semantics in Rust
 -   distinguish vales on the JS side

src/logging.rs
adespawn: It's currently unused `furtherInfo` field

src/logging.rs
wprzytula: I believe it's not a convincing justification to make the return type broader than it actually needs to be.

src/logging.rs
wprzytula: 1. WDYM by _currently_? In this PR? In our intended logging bridge implementation?
2. What is it used for in the original driver?
3. Why isn't it used in our implementation?

adespawn:
Part of the comment

lib/client.js
```diff
@@ -235,6 +236,17 @@ class Client extends events.EventEmitter {
             undefined,
         );
 
+        {
+            const emitter = this.emit.bind(this);
+            this.#loggingId = rust.setupLogging(
+                (level, target, message, furtherInfo) => {
+                    emitter("log", level, target, message, furtherInfo);
```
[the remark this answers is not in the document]
wprzytula: @adespawn Is this relevant?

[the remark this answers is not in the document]
adespawn: Yes, but I already fixed it

src/logging.rs
adespawn: In our intended logging bridge implementation.

> What is it used for in the original driver?

Sometimes additional information gets passed to the logger. See `(?<!console)\.log\((.*),(.*),(.*)\)` regex search on the code

> Why isn't it used in our implementation?

I was recreating how c++ behaves in logging, and there is no matching field there

wprzytula: Please put this explanation in the comment there.

src/logging.rs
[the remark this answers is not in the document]
adespawn: Improved comment, as discussed

src/logging.rs
```diff
@@ -24,6 +85,104 @@ pub(crate) fn parse_js_level_to_rust(level: &str) -> Option<Level> {
...
+/// Unregister a previously-registered logging callback.
+///
+/// After this call the callback associated with `id` will no longer receive
+/// any tracing events.  If `id` does not match any registered callback the
+/// call is a silent no-op.
+///
+/// Return false only when the removal fails (RW lock is poisoned)
+#[napi]
+pub fn remove_logging(id: i64) {
```
wprzytula: 🔧 Outdated comment.

src/logging.rs
```diff
@@ -126,7 +157,7 @@ impl<S: tracing::Subscriber> Layer<S> for JsForwardingLayer {
                         data: (
                             level_str.to_owned(),
                             target.to_owned(),
-                            message.to_owned(),
+                            visitor.log_message.to_owned(),
```
wprzytula: 🔧 `visitor.log_message` is already an owned String.

wprzytula (changes requested):

wprzytula (approved):

PR #464, opened 2026-06-08
Update the release process

### LINKED ISSUE #450: Add package verification after publishing to npm
The last release published a broken package to npm. Let's make sure such problems are immediately discovered. I suggest adding a new step to the release workflow that will, for example, build and run driver's code examples with the package fetched from npm.

adespawn:
This PR introduces 3 changes to the release process:
- Testing with --dry-run: Closes #416
- Post release tests: Closes #450 
- Ability to manually test the release process

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -189,3 +189,26 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+  verify-published-package:
+    needs:
+      - publish
+    name: Verify published package
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version: 24
+      - uses: ./.github/common/example
+        with:
+          build-command: |
+            cd examples
+            VERSION=$(node -p "require('../package.json').version")
+            npm pkg set "dependencies.scylladb-driver-alpha=$VERSION"
+            (for i in 1 2 3 4 5; do
+              npm install && exit 0
+              echo "Attempt $i failed, retrying in 15s (waiting for npm propagation)..."
+              sleep 15
+            done
+            echo "Giving up. Couldn't download newly released package."
+            exit 1)
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,11 @@ on:
   release:
     types: [published]
   workflow_dispatch:
+    inputs:
+      publish:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -162,6 +162,13 @@ jobs:
       # We do not want them in the main package, where the napi-rs has placed them.
       - name: Clean up binaries 
         run: rm *.node
+      - name: Dry-run publish
+        run: |
+          cd npm/linux-x64-gnu && npm publish --dry-run --access public && cd -
+          cd npm/linux-arm64-gnu && npm publish --dry-run --access public && cd -
+          npm publish --dry-run --access public --ignore-scripts
+        env: 
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
adespawn: `NODE_AUTH_TOKEN` Is not used, since we use trusted publishing

.github/workflows/release.yml
```diff
@@ -162,6 +162,13 @@ jobs:
       # We do not want them in the main package, where the napi-rs has placed them.
       - name: Clean up binaries 
         run: rm *.node
+      - name: Dry-run publish
+        run: |
+          cd npm/linux-x64-gnu && npm publish --dry-run --access public && cd -
+          cd npm/linux-arm64-gnu && npm publish --dry-run --access public && cd -
+          npm publish --dry-run --access public --ignore-scripts
```
wprzytula, line 187: The commit message would be great as a comment here.

> Add dry-run publish step before actual release
> We publish multiple packages atomically: 2 platform-specific native
> addon packages (linux-x64-gnu, linux-arm64-gnu) and 1 main driver
> package. If any package fails mid-publish, we end up with a partial
> release that is difficult to recover from.
>
> Why do we need separate dry runs, but only a single real publish commands?
>
> Napi-rs manages releasing sub-packages through prepublish configured command.
> However `prepublish --dry-run` flag only skips side-effects
> without actually validating against the npm registry. To get real
> validation, we call `npm publish --dry-run` explicitly on every sub-package.
>
> The main package uses --ignore-scripts to prevent the prepublishOnly
> hook (napi prepublish) from running during the dry-run, since that
> hook would actually publish the sub-packages.

.github/workflows/release.yml
```diff
@@ -176,3 +176,24 @@ jobs:
...
+        with:
+          build-command: |
+            cd examples
+            VERSION=$(node -p "require('../package.json').version")
+            npm pkg set "dependencies.scylladb-driver-alpha=$VERSION"
+            for i in 1 2 3 4 5; do
+              npm install && break
+              echo "Attempt $i failed, retrying in 15s (waiting for npm propagation)..."
+              sleep 15
+            done
```
wprzytula, line 219: 🔧 This silently succeeds (false-positive) in case npm propagation takes long time. Make this fail in such case.

wprzytula (changes requested):

wprzytula (approved):

PR #474, opened 2026-06-16
docs: review makefile edits

dgarcia360:
Removes npx dependency (not necessary) and moves the `-r` fix to Makefile.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -43,7 +43,7 @@ clean:
 # Generate output commands
 .PHONY: jsdoc
 jsdoc:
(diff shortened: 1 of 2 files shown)


dgarcia360:
@annastuchlik could you please check if you get any error with these changes while building the docs?

annastuchlik (approved):
Looks good, thanks!

annastuchlik:
@wprzytula This is an update to the fix I introduced when preparing the docs for publication. It's an improvement that should be merged.

adespawn (approved):

Issue #475, opened 2026-06-16
Inconsistency in JS vs TS options


adespawn:
Client and execution otpions are double defined in `client-options.js` and `query-options.js` and `main.d.ts`. Those are 2 conflicting definitions. Improve the consistency of those 2 places.

For example in JS definition of client options we have:
- `@property {SslOptions} [sslOptions]`, while in TS definition we have: `interface SslOptions`. JS SSL options have pfx and passphrase fields missing from TS deffinition (ok, this one is my mistake from #388, but still this is a bug...) 
- `id` as per JS can be `string | UUID`, while TS still allows only for `UUID` (again, my mistake. I was extending JS interface but missed the TS definition)
- `idempotent` is named like this in JS but it's named `isIdempotent` in TS (#92 👀 )
- missing `metrics` field in JS code. This one is actually DSx but not mine

PR #476, opened 2026-06-16
Improve client file

adespawn:
This includes TS conversion, improving type precision for query arguments, and discovery of #475

### DIFF
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -90,18 +90,21 @@ const loggingFinalizationRegistry = new FinalizationRegistry(
  * console.log(row['key']);
  */
 class Client extends events.EventEmitter {
-    /**
-     * @package
-     */
+    /** @package */
     rustClient: rust.SessionWrapper | undefined;
     #encoder: Encoder;
     #loggingId: number | undefined;
 
+    /** @package */
     options: clientOptions.ClientOptions;
+    /** @package */
     rustOptions: ReturnType<typeof clientOptions.setRustOptions>;
     readonly profileManager!: InstanceType<typeof ProfileManager>;
+    /** @package */
     connected: boolean;
+    /** @package */
     connecting?: boolean;
+    /** @package */
     isShuttingDown: boolean;
     /**
      * Gets the schema and cluster metadata information.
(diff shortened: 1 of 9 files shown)


adespawn:
Tests fail due to #477

adespawn:
> Narrow #loggingId to number before FinalizationRegistry.register().

> this.#loggingId is typed as number | undefined, but FinalizationRegistry.register() requires a number heldValue. The current code at line 304 passes this.#loggingId directly, causing a type error in strict mode. Use a local narrowed value before registering.

Invalid: TS recognizes it's valid since we assign number just before the mentioned line

adespawn:
Note, currently we still have some small divergence between client.ts api and client in main.d.ts: The main difference is that main.d.ts does not allow for overloads and concurrent support for callbacks / promises

lib/client.ts
```diff
@@ -106,8 +113,9 @@ class Client extends events.EventEmitter {
      * behavior and of the server as seen from the driver side.
      *
      * By default, a [DefaultMetrics]{@link module:metrics~DefaultMetrics} instance is used.
+     * TODO: This filed is currently not supported
```
wprzytula: filed -> field

lib/client.ts
```diff
@@ -437,6 +454,8 @@ class Client extends events.EventEmitter {
                 pageState,
             );
         } else {
+            // This is checked in isNamedParameters. The assert here is to satisfy complicator
```
wprzytula: complicator 😆 
(compiler)

lib/client.ts
```diff
@@ -538,15 +557,18 @@ class Client extends events.EventEmitter {
             throw new Error("Expected to obtain a string query");
         }
 
+        let hints = execOptions.getHints();
+        let hints2: ColumnInfo[] | undefined;
+        if (hints && hints.length != 0 && Array.isArray(hints[0])) {
+            throw TypeError("For single statements expected 1D array of hins");
+        }
```
wprzytula, line 567: hins -> hints

wprzytula (changes requested):

wprzytula (dismissed):

wprzytula (approved):

PR #479, opened 2026-06-17
Update the package name

adespawn:
This changes the name for the GA to the `@scylladb/driver` as decided internally


adespawn:
The NPM should already be configured for those packages

adespawn:
Dry run release: https://github.com/scylladb/nodejs-rs-driver/actions/runs/27676931749

wprzytula (approved):

PR #482, opened 2026-06-17
docs: compile TypeScript before running JSDoc

### LINKED ISSUE #477: TS files missing from API documentation
Files written in TS are missing from public API documentation (like `uuid.ts` - we only have `https://nodejs-rs-driver.docs.scylladb.com/stable/api/types/TimeUuid.html` that is still written in JS)

annastuchlik:
## Problem

Fixes #477.

TypeScript source files (`uuid.ts`, `local-time.ts`, `tuple.ts`, `big-decimal.ts`, `version-number.ts`, `execution-profile.ts`, etc.) were silently excluded from the generated API documentation.

JSDoc cannot parse TypeScript syntax — it errors out immediately on type annotations, private class fields, and `import =` statements. As a result, the `jsdoc ./lib -r` invocation scanned `lib/` recursively but skipped every `.ts` file.

## Fix

Run `tsc` before `jsdoc` in the `jsdoc` Makefile target:

```makefile
cd .. && ./node_modules/.bin/tsc -p tsconfig.build.json --noEmitOnError false; npm run js-doc
```

`tsc` compiles `.ts` files in-place to `.js` alongside the sources (these are gitignored, per `lib/types/.gitignore`). JSDoc then picks them up normally via its default `.js` pattern. JSDoc comments are preserved in the compiled output because `noEmitOnError` is not set in `tsconfig.build.json`.

`--noEmitOnError false` is needed because `tsc` reports type errors in a clean checkout or in the docs CI environment — `index.d.ts` and `index.js` are generated artifacts of the Napi/Rust build step (`napi build`) which does not run there. With `--noEmitOnError false`, `tsc` still emits the `.js` files and exits with code 0, so `make` always continues to the JSDoc step.

## Known limitation

Parameter type information (the Type column in JSDoc parameter tables) is not populated for parameters typed only via TypeScript annotations. JSDoc can only read types from explicit `@param {Type}` tags in the comment, not from TypeScript type syntax stripped during compilation. This is a known limitation of the compile-first approach and is out of scope for this fix.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -43,7 +43,7 @@ clean:
 # Generate output commands
 .PHONY: jsdoc
 jsdoc:
-	cd .. && npm run js-doc
+	cd .. && ./node_modules/.bin/tsc -p tsconfig.build.json --noEmitOnError false; npm run js-doc
 	@echo
 	@echo "JSDoc build finished. The API docs are in ../public/docs/."
 


adespawn:
> This would not fix the issue because the compiled .js\ output does not contain the JSDoc comments — they live in the .ts\ source files. Documenting compiled output would produce empty or stripped API pages.

The only part that would be striped was type information, while any written docs still remains after TypeScript compimlation:

```ts
    /**
     * Creates a new instance of Uuid based on a Buffer
     * @param buffer The 16-length buffer.
     */
    constructor(buffer: Buffer) {
        if (!buffer || buffer.length !== 16) {
            throw new TypeError(
                "You must provide a buffer containing 16 bytes",
            );
        }
        this.#raw = buffer;
    }
```

->
```js
    /**
     * Creates a new instance of Uuid based on a Buffer
     * @param buffer The 16-length buffer.
     */
    constructor(buffer) {
        if (!buffer || buffer.length !== 16) {
            throw new TypeError("You must provide a buffer containing 16 bytes");
        }
        this.#raw = buffer;
    }
```

annastuchlik:
@adespawn Yes, I told AI it was wrong :) Give me a second. I've converted this PR to Draft. Once I'm happy with the result, I'll ping you and Stanislaw here for review.

annastuchlik:
@adespawn @wprzytula You can review it now. 
Note: https://github.com/scylladb/nodejs-rs-driver/issues/477 will be fixed by publishing .ts files info, but there still will be the bug reported by @adespawn: https://github.com/scylladb/nodejs-rs-driver/issues/477#issuecomment-4727375837 - the type information is missing:
image

This can only be fixed by after we completing JS to TS transition and applying a solution appropriate for TS.

adespawn (approved):

PR #485, opened 2026-06-18
Add the docs fix to muitiversion

adespawn:
This commit applies the same fix from #482 to the docs release-build process


adespawn:
@annastuchlik / @dgarcia360 This should fix #477

Lorak-mmk (approved):

PR #486, opened 2026-06-18
docs: complie ts inside js-doc script

dgarcia360:
## Problem

Multiversion is not publishing reference docs for .ts files.

## Fix

Compile ts too in multiversion, not only when building the preview locally.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -43,7 +43,7 @@ clean:
 # Generate output commands
 .PHONY: jsdoc
 jsdoc:
-	cd .. && ./node_modules/.bin/tsc -p tsconfig.build.json --noEmitOnError false; npm run js-doc
+	cd .. && npm run js-doc
 	@echo
 	@echo "JSDoc build finished. The API docs are in ../public/docs/."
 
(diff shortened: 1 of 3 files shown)


dgarcia360:
Closes https://github.com/scylladb/nodejs-rs-driver/issues/477

dgarcia360:
https://github.com/scylladb/nodejs-rs-driver/commit/c3b84a806c12f49f65839d0fd4daf3c4b0427591 potentially fixed the issue too. I'll rebase since now the fix would be redundant.

dgarcia360:
@annastuchlik @adespawn Consider merging these changes so that it only defines how to build js reference docs in one place.

Also, there's no need to include `./node_modules/.bin/` before `tsc`. `tsc` is already available via the `typescript` dependency.

annastuchlik (approved):

annastuchlik:
> Consider merging these changes so that it only defines how to build js reference docs in one place.
> 
> Also, there's no need to include `./node_modules/.bin/` before `tsc`. `tsc` is already available via the `typescript` dependency.


@adespawn @wprzytula Please review and merge this PR. While the current solution is working, it may get us in trouble in the future when we want to update the logic (after full migration to .ts) - and we'll have to remember to update it in two places.

adespawn (approved):

wprzytula (approved):

PR #499, opened 2026-07-20
ClusterSnapshot and support for Host metadata

### LINKED ISSUE #500: Use net.SocketAddress instead of String to represent Host address

hmarszalek:
`JsThreadOnly`: a small wrapper type. `SessionWrapper` needs async methods, and Rust requires anything used in async code to be safe to share across threads (`Send`/`Sync`). But the JS object handles we wanted to cache are only safe to touch on the JS thread. `JsThreadOnly` requires an `Env` value, which itself can't survive being passed to another thread, to prove that we're actually on the JS thread whenever it's used.

`ClusterSnapshot`: a cache for cluster metadata, to avoid rebuilding fresh JS objects from scratch every call. Pointer comparison is used to check if the snapshot is fresh. I cache an actual `HostMap` of real `Host` objects (constructed with `js_ctor`) into `ClusterSnapshot`, so `client.hosts` is served from the cache instead of being built fresh each time.

`CopyableBuffer`: A borrowed byte slice that, when converted to a napi value, is always copied into a fresh JS-owned `Buffer` via `napi_create_buffer_copy`.

`Host` address: `Host.address` used to be a plain "ip:port" string. We changed it to Node's real `net.SocketAddress` object, so it actually behaves like an address.

`HostMap` keying: `HostMap` used to look up hosts by address. We changed it to key by the host's `Uuid` instead, because a host's id never changes, but its address can.

`HostMap` storage: `HostMap`'s internal storage (keyed by hostId) is now built directly on the Rust side as a `Record`, via `NamedMap`, and handed to the JS constructor already built, instead of Rust passing over an intermediate `Array` of `Host` for the constructor to convert into a lookup structure itself.

`host.js` to `host.ts`: converted `Host`/`HostMap` to TypeScript, following the same pattern as the other already-migrated files.

Fixes: #500

## Pre-review checklist

- [x] I have adjusted the migration documentation (removed/changed API) in `./docs/source/migration-guide`


src/metadata/state.rs
```diff
@@ -0,0 +1,22 @@
...
+    pub(crate) inner: Arc<scylla::cluster::ClusterState>,
+}
+
+impl ClusterSnapshot {
+    pub(crate) fn new(
+        inner: Arc<scylla::cluster::ClusterState>,
+    ) -> ConvertedResult<Self> {
+        Ok(ClusterSnapshot { inner })
+    }
+}
```
wprzytula, line 31: 🤔 Why does this return a `ConvertedResult`? It's always `Ok`, so no need for `Result` here.

src/session.rs
```diff
@@ -51,6 +53,9 @@ unsafe impl<T> Sync for JsThreadOnly<T> {}
 #[napi]
 pub struct SessionWrapper {
     pub(crate) inner: CachingSession,
+    /// Cache of the last `ClusterSnapshot` (JS object) that was returned to JS,
+    /// alongside the `Arc<ClusterState>` pointer it was built from.
+    cluster_snapshot: Arc<Mutex<JsThreadOnly<Option<Reference<ClusterSnapshot>>>>>,
```
wprzytula: 😲  Wow, what a deeply nested type!

src/session.rs
```diff
@@ -320,6 +330,52 @@ impl SessionWrapper {
...
+    ) -> ConvertedResult<Reference<ClusterSnapshot>> {
+        let mut cache_guard = self
+            .cluster_snapshot
+            .lock()
+            .expect("poisoning impossible due to process-aborting panics");
+        let rust_cluster_state = self.inner.get_session().get_cluster_state();
+
+        let cached_state = cache_guard.0.as_ref();
+
+        if cached_state.is_none() || !Arc::ptr_eq(&rust_cluster_state, &cached_state.unwrap().inner)
```
wprzytula: 🔧 Avoid the `unwrap`!
You can use `is_none_or`.

src/session.rs
```diff
@@ -320,6 +330,52 @@ impl SessionWrapper {
...
+        }
+
+        ConvertedResult::Ok(
+            cache_guard
+                .0
+                .as_ref()
+                .expect("cluster snapshot was just initialized above")
+                .clone(env)?,
+        )
+    }
```
wprzytula, line 358: ♻️ Avoid the `expect` at the end by refactoring this. Tip: match on `cached_state` and make it syntactically guaranteed that the returned value exists.

src/metadata/host.rs
```diff
@@ -11,23 +19,56 @@ pub struct HostWrapper {
 }
 );
 
+impl From<Arc<Node>> for HostWrapper {
+    fn from(inner: Arc<Node>) -> Self {
+        HostWrapper {
+            host_id: Buffer::from(inner.host_id.as_bytes().as_slice()),
+            address: inner.address.to_string(),
```
wprzytula: 🤔 Does JS Driver really represent Host's address as a string?
Can we move to a socket address?

src/metadata/state.rs
```diff
@@ -8,15 +12,45 @@ use std::sync::Arc;
 /// Rather than mutating the previous snapshot in place, the driver produces a brand
 /// new `Arc<ClusterState>` on every refresh. This lets us cheaply detect whether the
 /// snapshot backing a given `ClusterSnapshot` is stale, by comparing Arc pointers.
-#[napi]
+#[napi(custom_finalize)]
 pub struct ClusterSnapshot {
     pub(crate) inner: Arc<scylla::cluster::ClusterState>,
+    /// All nodes known by the Rust driver at the time this snapshot was created, as JS objects.
+    pub(crate) known_nodes: Vec<napi::bindgen_prelude::sys::napi_ref>,
```
wprzytula: ⛏️ In this file you fully qualify `napi_ref`, and in another file you import it.

src/metadata/state.rs
```diff
@@ -8,15 +12,45 @@ use std::sync::Arc;
...
 }
```
wprzytula: ❓ Why does this hold raw `napi_ref` instead of `Reference`? What's the difference between these two?

src/metadata/state.rs
```diff
@@ -8,15 +12,45 @@ use std::sync::Arc;
...
+            let mut ref_count = 0;
+            check_status!(
+                unsafe { napi_reference_unref(env.raw(), host_ref, &mut ref_count) },
+                "Failed to unref HostWrapper N-API reference while finalizing ClusterSnapshot",
+            )?;
+            check_status!(
+                unsafe { napi_delete_reference(env.raw(), host_ref) },
+                "Failed to delete HostWrapper N-API reference while finalizing ClusterSnapshot",
+            )?;
+        }
```
wprzytula: ❓ What's the difference between these two functions? Why do we need to call both here?

lib/client.ts
```diff
@@ -337,6 +338,9 @@ class Client extends events.EventEmitter {
             this.rustClient = await rust.SessionWrapper.createSession(
                 this.rustOptions,
             );
+            this.metadata = new metadataModule.Metadata(
+                this as unknown as PublicClient,
+            );
```
wprzytula: ❓ How is `this` different from `PublicClient`?

lib/metadata/index.js
```diff
@@ -145,7 +147,12 @@ class Metadata {
...
      */
     getTable(keyspaceName, name) {
-        throw new Error("TODO: Not implemented");
+        let clusterSnapshot = this.#client.getClusterSnapshot();
+        let tableWrapper = clusterSnapshot.getTable(keyspaceName, name);
+        if (tableWrapper === null) {
+            return null;
+        }
+        return TableMetadata.fromRust(tableWrapper);
     }
```
wprzytula, line 141: 💭 Now, two subsequent calls `metadata.getTable()` can return different results if the underlying Rust Driver's `ClusterState` changes inbetween.
Was this possible in the old driver?

lib/metadata/table-metadata.js
```diff
@@ -39,6 +39,20 @@ class ColumnMetadata {
...
+     * @internal
+     * @ignore
+     */
+    static fromRust(columnWrapper) {
+        let columnMetadata = new ColumnMetadata();
+        columnMetadata.type = convertComplexType(columnWrapper.typ);
+        columnMetadata.kind = columnWrapper.kind.kind;
+        return columnMetadata;
+    }
 }
```
wprzytula: 🔧 I dislike that we need to expose Rust `columnWrapper` to JS. This has major overhead. Can't we instead construct `ColumnMetadata` directly in Rust? This is the approach taken in Python RS codewide: to avoid wrapping and exposing native Rust structs to the dynamic language.

wprzytula: Using Node API, Rust should be able to call the `ColumnMetadata` constructor, converting the necessary parameters (type, kind) beforehand.

src/metadata/state.rs
```diff
@@ -54,3 +59,84 @@ impl napi::bindgen_prelude::ObjectFinalize for ClusterSnapshot {
...
+            RustColumnKind::Regular => ColumnKind::Regular,
+            RustColumnKind::Static => ColumnKind::Static,
+            RustColumnKind::Clustering => ColumnKind::ClusteringKey,
+            RustColumnKind::PartitionKey => ColumnKind::PartitionKey,
+            // ColumnKind is #[non_exhaustive] in the rust driver, so future variants
+            // added there need a fallback here. Regular is the safest default.
+            _ => ColumnKind::Regular,
+        }
+    }
+}
```
wprzytula: 🔧 We don't want a silent fallback. We want the compilation/linting to fail loudly if a new variant is used, which is exactly what `#[deny(clippy::wildcard_enum_match_arm)]` does.

src/metadata/state.rs
```diff
@@ -54,3 +59,84 @@ impl napi::bindgen_prelude::ObjectFinalize for ClusterSnapshot {
...
+                        partition_key: rust_table.partition_key.clone(),
+                        clustering_key: rust_table.clustering_key.clone(),
+                        partitioner: rust_table.partitioner.clone(),
+                    }))
+                } else {
+                    ConvertedResult::Ok(None)
+                }
+            } else {
+                ConvertedResult::Ok(None)
+            }
```
wprzytula: ♻️ `if-let`s are unreadable - they damage locality. Use `let-else`.

lib/metadata/table-metadata.js
wprzytula: Actually, `_rust.ColumnWrapper` already has the desired shape. It has the same properties as `ColumnMetadata`. It should be easy to merge them into one, by creating the latter in Rust directly.

src/metadata/state.rs
```diff
@@ -54,3 +59,84 @@ impl napi::bindgen_prelude::ObjectFinalize for ClusterSnapshot {
...
+);
+
+define_rust_to_js_convertible_object!(
+pub struct TableWrapper {
+    columns, columns: Vec<(String, Column)>,
+    partition_key, partitionKey: Vec<String>,
+    clustering_key, clusteringKey: Vec<String>,
+    partitioner, partitioner: Option<String>,
+}
+);
```
wprzytula: 💭 This type is purely ephemeral - it only exists to be instantly structurally converted to JS object.
But still this allocates a lot! Both `String`s and `Vec`s. Do you see the problem?

I have a draft branch that would address such issues by introducing borrowed wrappers, using references (`&str`) and iterators (`impl Iterator<Item = &'a str>`). This would avoid the ephemeral overhead altogether. Does this make sense?

src/custom-types.d.ts
```diff
@@ -56,6 +56,35 @@ export interface HostWrapper {
...
+  typ: ComplexType
+  kind: ColumnKind
+}
+
+export interface TableWrapper {
+  columns: Array<[string, Column]>
+  partitionKey: Array<string>
+  clusteringKey: Array<string>
+  partitioner?: string | null
+}
```
wprzytula: ❓ ❗ Why is this here? Why is it called `Wrapper` and present in the public API?

test/integration/supported/metadata-materialized-view-tests.js
wprzytula, line 1: ❓ Are these tests new or taken from `broken`?

wprzytula (commented):
Please examine expanded napi macro's code of the added items. Verify that there's no big waste of memory and computation there.

src/metadata/host.rs
hmarszalek: > Does JS Driver really represent Host's address as a string?

Yup. 
But it should be possible to move to a socket address.

src/metadata/state.rs
hmarszalek: `#[napi]` generates a real JS class (backed by a native N-API wrapped object). The Rust struct instance lives on the heap, and JS holds an opaque handle to it. Supports `Reference<T>`.

`define_rust_to_js_convertible_object!` generates a plain Rust struct without #[napi], plus a manual ToNapiValue impl that builds a plain JS object literal via Object::new + set_named_property, consuming the value by move. Can't use `Reference<T>` and requires manual napi_ref implementation.

src/metadata/state.rs
hmarszalek: `napi_reference_unref` decreases the reference count. Once it hits 0, the value stops being artificially kept alive by this reference (it can still be collected, or kept alive by something else entirely - e.g. a JS variable holding it).

`napi_delete_reference` frees the reference handle itself. This is unrelated to the count. napi_create_reference also allocates a small native-side bookkeeping struct (the opaque napi_ref) that Node-API uses internally to track this particular reference.

lib/client.ts
hmarszalek: typechecking issues...

test/integration/supported/metadata-materialized-view-tests.js
hmarszalek: New, just simple examples to prove the bridge works.

lib/metadata/table-metadata.js
hmarszalek: @wprzytula To create a JS class in Rust, we would need to register its constructor first. Is that what you want me to do?

src/metadata/host.rs
wprzytula: Let's do that in a separate PR I think. A pre-PR or a follow-up, whichever you find more reasonable.

src/metadata/state.rs
wprzytula: I know the difference between `#[napi]` and `define_rust_to_js_convertible_object!`, but thanks for confirming.

What I still miss is the difference between `Reference` and `napi_ref`. What's `Reference` internally? Why can't it be used instead of `napi_ref` in our case?

lib/client.ts
wprzytula: This explains nothing to me. Can you elaborate?

lib/metadata/table-metadata.js
wprzytula: In C# RS, we simply pass pointers to C# constructors (in case of exceptions, among others) to Rust. There, they can be static functions and have static lifetime. In Node.js, I guess functions must be registered dynamically. Is this what you mean?
If we register such a constructor once in the driver's lifetime, will it decrease overhead of all subsequent metadata constructions? I mean, we won't have to create plain JS object on Rust side and then construct the proper JS class instance in JS code from it.

src/metadata/state.rs
hmarszalek: Using `into_reference` for `Reference<T>` requires T: `JavaScriptClassExt` (i.e. #[napi] class with registered constructor + napi_wrap + finalizer). 

```rust
pub struct Reference<T: 'static> {
  raw: *mut T,
  napi_ref: crate::sys::napi_ref,
  env: *mut c_void,
  // the finalize callbacks can only be written with the `Env` passed in
  // So we can use `Cell` rather than `AtomicPtr` here
  finalize_callbacks: Arc<Cell<*mut dyn FnOnce()>>,
}
```

We are operating on 100% JS objects (here literals, but I am currently working on class constructors), so we cannot turn them into a `Reference`.

lib/metadata/table-metadata.js
hmarszalek: > If we register such a constructor once in the driver's lifetime, will it decrease overhead of all subsequent metadata constructions? I mean, we won't have to create plain JS object on Rust side and then construct the proper JS class instance in JS code from it.

Yes

src/metadata/js_ctor.rs
```diff
@@ -0,0 +1,66 @@
+use std::cell::RefCell;
+
+/// Defines a per-thread constructor registry for a single pure-JS metadata class, together with:
```
wprzytula: ❓ Does this have to be a metadata class? Can it be a general-purpose macro for our project?

src/metadata/js_ctor.rs
```diff
@@ -0,0 +1,66 @@
...
+/// - a `#[napi]` `register_*_ctor` function that JS calls once, at module load time, to hand
+///   Rust a reference to the class's constructor;
+/// - a `pub(crate)` `build_*` function that constructs a new instance of that class directly,
+///   given the constructor arguments.
+///
+/// The `Return` type parameter of the underlying `Function`/`FunctionRef` is always `()`: it is
+/// only used by `Function::call`, whereas we always construct instances via
+/// `Function::new_instance`, which returns an untyped `Unknown` value regardless of it.
```
wprzytula: 🤔 I fail to understand this.

lib/client.ts
hmarszalek: I don't exactly understand it YET. That's why this PR is still a draft...

hmarszalek: Outdated... Problem for the future

src/metadata/js_ctor.rs
wprzytula: Aha, this means that the `Return` type is actually always ignored, because it's not used by `Function::new_instance`, so it can be set to arbitrary type; for simplicity, `()` is used.

src/metadata/js_ctor.rs
```diff
@@ -0,0 +1,66 @@
...
+        static_name: $static_name:ident,
+        register_fn: $register_fn:ident,
+        build_fn: $build_fn:ident,
+        args: $args_ty:ty,
+        class_name: $class_name:literal,
+    ) => {
+        thread_local! {
+            $(#[$doc])*
+            static $static_name: RefCell<Option<FunctionRef<$args_ty, ()>>> = const { RefCell::new(None) };
+        }
```
wprzytula: 🔧 Justify why thread-local storage is the proper choice for this.

src/metadata/js_ctor.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// Registers the JS class constructor, so that Rust can later construct fully-formed JS instances directly.
+        /// This must be called exactly once per thread (i.e. once per `worker_threads` worker, and once on the
+        /// main thread), by the corresponding `lib/metadata/*.js` module, before any cluster metadata is accessed
+        /// on that thread.
```
wprzytula: ❓ How is this enforced? It's definitely our duty, not user's.

src/metadata/js_ctor.rs
```diff
@@ -0,0 +1,66 @@
...
+                        " constructor is already registered",
+                    )));
+                }
+                *slot = Some(ctor_ref);
+                Ok(())
+            })
+        }
+
+        /// Constructs a JS class instance directly, by calling its constructor, registered via register_fn.
+        pub(crate) fn $build_fn(env: &Env, args: $args_ty) -> napi::Result<Object<'static>> {
```
wprzytula: 🔧 Please add a lifetime parameter: flowing from `&'env Env` to `Object<'env>`.

src/session.rs
```diff
@@ -320,6 +330,52 @@ impl SessionWrapper {
...
+}
+
+impl SessionWrapper {
+    /// Returns the JS `ClusterSnapshot` object representing the current cluster state.
+    ///
+    /// `ClusterState` snapshots produced by the Rust driver are never mutated in place:
+    /// whenever the driver refreshes cluster topology/schema metadata in the background,
+    /// it produces a brand new `Arc<ClusterState>`, leaving any previously handed out
+    /// `Arc` untouched. This lets us detect whether our cached snapshot is stale by
+    /// comparing Arc pointers (`Arc::ptr_eq`):
```
wprzytula, line 321: 🔧 Add mention that we prevent ABA by keeping Arc alive.

src/metadata/host.rs
```diff
@@ -1,33 +1,63 @@
...
+
+/// Builds the arguments passed to the JS Host constructor for the given node.
+fn host_ctor_args(node: &Arc<Node>) -> FnArgs<(String, Option<String>, Option<String>, Buffer)> {
+    FnArgs::from((
+        node.address.to_string(),
+        node.datacenter.clone(),
+        node.rack.clone(),
+        Buffer::from(node.host_id.as_bytes().as_slice()),
+    ))
 }
```
wprzytula, line 57: 🔧 Let's try avoid this cloning and pass borrowed data to JS ctor.

src/metadata/js_ctor.rs
```diff
@@ -0,0 +1,66 @@
...
+/// `Function::new_instance`, which returns an untyped `Unknown` value regardless of it.
+#[expect(unused)]
+macro_rules! define_js_ctor {
+    (
+        $(#[$doc:meta])*
+        static_name: $static_name:ident,
+        register_fn: $register_fn:ident,
+        build_fn: $build_fn:ident,
+        args: $args_ty:ty,
+        class_name: $class_name:literal,
```
wprzytula: Let's pass it as `ident` and then `stringify!()`. At call sites this will look nicer, because e.g. `Host` looks more like JS `Host`.

src/metadata/state.rs
```diff
@@ -7,15 +12,45 @@ use std::sync::Arc;
...
+                unsafe { napi_reference_unref(env.raw(), host_ref, &mut ref_count) },
+                "Failed to unref HostWrapper N-API reference while finalizing ClusterSnapshot",
+            )?;
+            check_status!(
+                unsafe { napi_delete_reference(env.raw(), host_ref) },
+                "Failed to delete HostWrapper N-API reference while finalizing ClusterSnapshot",
+            )?;
+        }
+        Ok(())
     }
```
wprzytula: 🔧 Let's implement `ObjectFinalize` structurally:
1. for our `napi_ref` wrapper (`NapiRef`?)
2. for `Vec` - in `napi` crate (orphan rule!)
3. for `HashMap`.
4. for `ClusterSnapshot`.

lib/metadata/index.js
```diff
@@ -145,7 +149,8 @@ class Metadata {
      * @param {String} name Name of the Table.
      */
     getTable(keyspaceName, name) {
-        throw new Error("TODO: Not implemented");
+        let clusterSnapshot = this.#rustClient.getClusterSnapshot();
+        return clusterSnapshot.getTable(keyspaceName, name);
     }
```
wprzytula, line 141: ❓ Wouldn't it be cheaper here to fully omit the JS-wrapped `ClusterSnapshot` and performing the whole operation in Rust?
Pseudocode:
```rust
let cs: ClusterState = session.get_cluster_state();
let ks = cs.get_keyspace(ks_name);
let tbl = ks.tables.get(tbl_name);
js_table_ctor(tbl)
```

lib/metadata/table-metadata.js
```diff
@@ -73,8 +88,36 @@ class TableMetadata {
...
      */
     partitioner;
+
+    /**
+     * Constructs a TableMetadata instance.
+     *
+     * Instances of this class are constructed directly by the native driver when reading cluster metadata.
+     * @param {Array<Array<(String|ColumnMetadata)>>} columns
+     * @param {Array.<String>} partitionKey
+     * @param {Array.<String>} clusteringKey
```
wprzytula: https://share.gemini.google/1pSBNHJJPhFo

Tl;dr: Gemini advises to use `Array<string>` or `string[]`. Note the lowercase and lack of the dot.

src/session.rs
hmarszalek: :smiling_face_with_tear:

src/metadata/js_ctor.rs
hmarszalek: It's registered on module load.

lib/metadata/index.js
hmarszalek: You mean to delegate all metadata queries through `SessionWrapper` / `rustClient`?

hmarszalek: Should be cheaper, because we would avoid wrapping and unwrapping both RS-JS types.

src/utils/js_instance.rs
```diff
@@ -0,0 +1,47 @@
...
+impl<C> ToNapiValue for JsInstance<'_, C> {
+    unsafe fn to_napi_value(env: sys::napi_env, val: Self) -> napi::Result<sys::napi_value> {
+        unsafe { ToNapiValue::to_napi_value(env, val.object) }
+    }
+}
+
+impl<C> TypeName for JsInstance<'_, C> {
+    fn type_name() -> &'static str {
+        "Object"
+    }
```
wprzytula: 💭 Why not use `std::any::type_name::<C>()`?

wprzytula: ❓ What is `TypeName::type_name` used for in napi-rs?

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,105 @@
...
+use crate::utils::js_instance::JsInstance;
+
+/// A typed abstraction over a raw N-API owning a strong reference that pins a `JsInstance<C>`
+/// The `C` type parameter records which JS class the pinned object is an instance of, so retrieving
+/// it later yields a correctly-typed `JsInstance<'_, C>` rather than an untyped `Object`. `JsInstance`
+/// is a plain JS object not wrapped as a napi-rs `#[napi]` class instance, so `Reference<T>` cannot be used.
+///
+/// Unlike a raw `napi_ref`, which has no `Drop` semantics of its own, `NapiRef` releases the reference
+/// it owns automatically when dropped (`napi_reference_unref`, then `napi_delete_reference` once the
+/// strong count reaches zero) - mirroring what napi-rs's own `Reference<T>` does for `#[napi]` class
```
wprzytula: > `napi_reference_unref`, then `napi_delete_reference` once the strong count reaches zero

💭 Can't it happen that when `NapiRef` is dropped and `napi_reference_unref` is called, there are still some other references that keep the strong count positive?

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,105 @@
...
+            env: env.raw(),
+            _class: PhantomData,
+        })
+    }
+
+    /// Retrieves the `C` instance currently pinned by this reference, as a `JsInstance<'env, C>`
+    /// whose lifetime is tied to the borrow of `env` passed in - preventing the returned object
+    /// from outliving the `Env` it was retrieved through. `env` must refer to the same environment
+    /// this `NapiRef` was created with.
+    pub(crate) fn get<'env>(&self, env: &'env Env) -> napi::Result<JsInstance<'env, C>> {
```
wprzytula: ⚠️ Lifetime issue: I think `napi_value` is only guaranteed to live as long as `napi_ref` that the value is pinned by. Do you agree?

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,105 @@
...
+        if ref_count == 0 {
+            let delete_status = unsafe { sys::napi_delete_reference(self.env, self.napi_ref) };
+            debug_assert_eq!(
+                delete_status,
+                sys::Status::napi_ok,
+                "Failed to delete N-API reference"
+            );
+        }
+    }
+}
```
wprzytula, line 85: ❓ What if ref count is not `0` when dropping this? Will the `napi_ref` leak and will never be deleted?

src/session.rs
```diff
@@ -320,6 +330,56 @@ impl SessionWrapper {
...
+                Some(ClusterSnapshot::new(rust_cluster_state).into_reference(env)?);
+        }
+
+        match cache_guard.0.as_ref() {
+            Some(cached_state) => ConvertedResult::Ok(cached_state.clone(env)?),
+            None => ConvertedResult::Err(ConvertedError::from(make_js_error(
+                "Cluster snapshot was just initialized above, but is still None",
+            ))),
+        }
+    }
```
wprzytula: ❓ Is `None` here possible due to some external factors or only due to a bug in the driver logic?

If it's clearly a result of a bug, then we can simply panic here (using `unreachable!`).

lib/client.ts
```diff
@@ -205,9 +205,9 @@ class Client extends events.EventEmitter {
     get hosts(): HostMap {
         if (!this.rustClient) return HostMap.fromRust([]);
 
-        // For now we retrieve all the hosts per each access to this field.
-        // This may be inefficient when user works directly with this field multiple times,
-        // but with this approach we shift the responsibility of ensuring validity of the data to Rust driver
+        // rustClient.getAllHosts() attempts to read the cached hosts from the Rust driver.
+        // If no cache is available, or the previous cache is stale, it will trigger a refresh
+        // of the host list from the cluster.
         return HostMap.fromRust(this.rustClient.getAllHosts());
```
wprzytula: `Host cache`: please add proper commit message.

src/metadata/host.rs
```diff
@@ -1,33 +1,54 @@
...
```
wprzytula, line 57: ❓ Can we make this use borrowed types to avoid allocations?

test/integration/supported/metadata-host-tests.js
```diff
@@ -0,0 +1,111 @@
...
+
+        describe("when the client is connected", function () {
+            it("should return a HostMap with real Host instances", function (done) {
+                const hosts = setupInfo.client.hosts;
+                assert.instanceOf(hosts, HostMap);
+                assert.strictEqual(hosts.length, 1);
+
+                const host = hosts.values()[0];
+                assert.instanceOf(host, Host);
+                done();
```
wprzytula, line 25: 🔧 `host tests` commit: please add a comment for every test, explaining what the test verifies and how.

lib/metadata/index.js
```diff
@@ -38,6 +38,10 @@ const _traceMaxAttemps = 5;
  */
 const _traceAttemptDelay = 400;
 
+// Used for JS doc
+// eslint-disable-next-line no-unused-vars
+const { TableMetadata } = require("./table-metadata");
+
```
wprzytula: 🔧 Explain what js-doc requires this for.

lib/metadata/index.js
wprzytula, line 141: 💭 Same question as before: why not move the whole operation to Rust, not to even instantiate the JS ClusterSnapshot?

hmarszalek: I am planning to. Just a lot of changes were suggested and I am doing them one / two a the time.

src/metadata/host.rs
hmarszalek: Done.

lib/metadata/index.js
hmarszalek: TableMetadata is needed to describe the return type in the documentation, but the import is never actually used in the code. This results in eslint failure.

src/utils/js_instance.rs
hmarszalek: napi-rs requires this trait to be implemented in some code paths where conversion code names the type on error. After reviewing it, I realized that this trait is already implemented for the underlying `Object` and is not necessary here.

wprzytula: OK, so my comment is outdated?

lib/metadata/index.js
wprzytula: `getTable` operates on `client` and returns a JS table object. It means there's no need to create an intermediate `ClusterSnapshot` JS object.

lib/metadata/index.js
wprzytula: Let's put it in the comment.

src/utils/napi_ref.rs
hmarszalek: I check the `ref_count` and only call `napi_delete_reference` if it is equal to 0.

src/utils/napi_ref.rs
hmarszalek: Yes, that is the point of pinning it. I don't understand the issue.

src/utils/napi_ref.rs
hmarszalek: We assume every clone will be eventually dropped. The last one will trigger `napi_delete_reference`.

src/utils/napi_ref.rs
wprzytula: 1. Then I'd rather write: 
>, then `napi_delete_reference` if the strong count has reached zero

2. In such case (strong count remaining higher than 0), are we sure the underlying `napi_ref` is deleted at some point?

src/utils/napi_ref.rs
wprzytula: You are creating a `JsInstance` whose lifetime is only constrained by `'env`, which is the lifetime of the JS `Env`. You do not constrain `JsInstance`'s lifetime with lifetime of `self`. This means that `JsInstance` can outlive `NapiRef`, leading to UAFs.

wprzytula: "Pinning" you're referring to is a runtime concept. Lifetimes are a compile-time concept of the borrow checker. Let's leverage borrow checker to prevent misuse.

src/utils/napi_ref.rs
wprzytula: OK, won't Node ever call `napi_reference_ref` on its own?

src/utils/napi_ref.rs
hmarszalek: The only way to call `napi_reference_ref` is with `NapiRef::clone`. Every object created this way will eventually be dropped.

src/utils/napi_ref.rs
hmarszalek: Nope, this reference count is entirely separate from Node.

src/utils/napi_ref.rs
hmarszalek: Fixed in #506.

lib/metadata/index.js
hmarszalek: It's already used in many other places in our driver without additional comments.

lib/metadata/table-metadata.js
```diff
@@ -73,8 +88,36 @@ class TableMetadata {
...
+     * @ignore
+     */
+    constructor(columns, partitionKey, clusteringKey, partitioner) {
+        let columnsByName = {};
+        for (let [name, columnMetadata] of columns) {
+            columnsByName[name] = columnMetadata;
+        }
+        this.columns = columnsByName;
+        this.partitionKey = partitionKey;
+        this.clusteringKey = clusteringKey;
```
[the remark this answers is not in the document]
hmarszalek: Fututre problem :smiling_face_with_tear:

lib/metadata/index.js
hmarszalek: Previously, the DSx driver checked `isMetadataSyncEnabled` option. If it was disabled, it fetched table info from the database. Otherwise, it returned information based on the current keyspace cache

test/integration/supported/metadata-host-tests.js
```diff
@@ -0,0 +1,119 @@
...
+
+            it("should iterate hosts via forEach() consistently with values()/keys()", function (done) {
+                const hosts = setupInfo.client.hosts;
+                const seen = [];
+                hosts.forEach((host, address) => seen.push([address, host]));
+
+                assert.lengthOf(seen, hosts.length);
+                seen.forEach(([address, host]) => {
+                    assert.strictEqual(host.address, address);
+                });
```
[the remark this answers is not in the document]
hmarszalek: TODO: fix host tests when API is decided.

wprzytula (changes requested):
Code review using Opus 5 found 3 issues:

1. lib/host.js:201 — HostMap.forEach now passes value.address (a net.SocketAddress) as the callback's key argument instead of the previous string key. Consumers using that key for string comparison or as a Map key break silently.
2. lib/host.js:225 — keys(asString = true) defaults to returning strings, but its JSDoc declares @returns {Array<_nodeNet.SocketAddress>}. Callers following the documented contract get strings.
3. lib/metadata/index.js:205 — getTrace silently ignores the consistency argument (only mentioned in a comment). Callers passing a non-default consistency get the trace read at the session-level setting, with no error.

src/types/type_wrappers.rs
```diff
@@ -81,6 +81,7 @@ impl ToNapiValue for ComplexType<'_> {
             ColumnType::Native(native_type) => {
                 obj.set_named_property(
                     base_type_name,
+                    #[deny(clippy::wildcard_enum_match_arm)]
```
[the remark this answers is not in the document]
wprzytula: 🤔 Hmm, is this right?
🔧 Anyway, let's make this `_`.

lib/metadata/index.d.ts
```diff
@@ -111,7 +111,7 @@ export interface SchemaFunction {
 
 export interface UdtField {
   name: string;
-  type: DataTypeInfo;
+  type: ColumnInfo;
 }
```
wprzytula, line 115: ❓ What was `DataTypeInfo`? Why change it? WDYM by "fix"?

wprzytula: Much more context is needed in the commit message. I can see `UdtField` interface was defined at two sites. Why?

lib/types/cql-utils.ts
wprzytula, line 1: Commit `Preserve type.frozen during conversion from ComplexType to ColumnInfo.`
❓ What's the difference between `ComplexType` and `ColumnInfo`? Why is one converted into the other? Again the context is missing.

lib/types/cql-utils.ts
```diff
@@ -72,10 +72,14 @@ export class ColumnInfo {
...
         info: ColumnInfo["info"] = null,
         customTypeName?: string,
+        options?: ColumnInfoOptions,
     ) {
         this.code = code;
         this.info = info;
         this.customTypeName = customTypeName;
+        if (options !== undefined) {
+            this.options = options;
+        }
```
wprzytula: ❓ Why is such check needed? Can't we just assign undefined?

lib/types/cql-utils.ts
```diff
@@ -113,28 +117,40 @@ export function convertComplexType(type: ComplexType): ColumnInfo {
...
+                        fields: type.udt_types.map(
+                            (typ: ComplexType, index: number) => ({
+                                type: convertComplexType(typ),
+                                name: type.udt_name[index],
+                            }),
+                        ),
+                    },
+                    undefined,
+                    { frozen: type.frozen },
+                );
```
wprzytula: 💭 I think that for performance we should consider moving this logic into Rust - in spirit of our new approach: use Rust to build end JS types; avoid creating intermediate JS objects.

wprzytula: ❓ Why is this function even in JS? `ComplexType` is a wrapped Rust type, isn't it? So this introduces big overhead.

lib/metadata/index.d.ts
```diff
@@ -81,20 +81,22 @@ export interface TableMetadata {
   partitioner: string | null;
 }
 
+export interface TracingEvent {
+  id: Uuid;
+  activity: any;
+  source: any;
+  elapsed: any;
+  thread: any;
+}
```
wprzytula: ❓ Why do we use `any` here? Can't we type these fields?

lib/metadata/index.d.ts
```diff
@@ -182,8 +184,18 @@ export interface Metadata {
 
   getTrace(
     traceId: Uuid,
-    consistency?: types.consistencies,
-  ): QueryTrace | null;
+    consistency: types.consistencies,
```
wprzytula: ❓ Why do you now forbid nullability of `consistency`?

lib/metadata/index.d.ts
```diff
@@ -182,8 +184,18 @@ export interface Metadata {
...
+  ): void;
+
+  getTrace(
+    traceId: Uuid,
+    consistency: types.consistencies,
+  ): Promise<QueryTrace>;
+
+  getTrace(traceId: Uuid, callback: ValueCallback<QueryTrace>): void;
+
+  getTrace(traceId: Uuid): Promise<QueryTrace>;
```
wprzytula: 💭 Do we need two pairs of these function? I understand that we need one callback-based API and one promise-based. But why both:
```js
getTrace(traceId: Uuid): Promise<QueryTrace>;
```
```js
getTrace(
    traceId: Uuid,
    consistency: types.consistencies,
  ): Promise<QueryTrace>;
```

Can't we represent them as one function with optional `consistency` argument?

lib/metadata/index.d.ts
wprzytula: Aha, because you handle this case as a separate function. Do we need to do such split?

lib/metadata/index.d.ts
```diff
@@ -182,8 +184,18 @@ export interface Metadata {
...
-    consistency?: types.consistencies,
-  ): QueryTrace | null;
+    consistency: types.consistencies,
+    callback: ValueCallback<QueryTrace>,
+  ): void;
+
+  getTrace(
+    traceId: Uuid,
+    consistency: types.consistencies,
+  ): Promise<QueryTrace>;
```
wprzytula: ❓ Why was `getTrace` sync before? Didn't it need to execute I/O to obtain the trace?

lib/metadata/query-trace.js
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * @type {Number|null}
+     */
+    elapsed;
+
+    /**
+     * @type {String|null}
+     */
+    thread;
+}
```
wprzytula: 🤔 Here you strongly type the fields. Why not in the `.ts` file?

lib/metadata/query-trace.js
wprzytula, line 1: ❓ Why not write this file in TS from the very beginning? In general, writing new code in TS should be easier than migrating existing JS code to TS.

test/typescript/metadata-tests.ts
```diff
@@ -31,15 +31,16 @@ async function myTest(): Promise<any> {
...
-    const trace: QueryTrace | null = client.metadata.getTrace(
-        types.Uuid.random(),
-    );
-    const traceWithConsistency: QueryTrace | null = client.metadata.getTrace(
-        types.Uuid.random(),
-        types.consistencies.one,
+    let trace: QueryTrace = await client.metadata.getTrace(types.Uuid.random());
+    client.metadata.getTrace(types.Uuid.random(), (err, t) =>
+        useResult<QueryTrace>(err, t),
     );
```
wprzytula: ⛏️ I don't think we need to change this `const` into `let`.

src/session.rs
```diff
@@ -26,6 +26,25 @@ pub struct BatchWrapper {
...
+///
+/// # Safety
+/// Values of this type must only be constructed, read, or dropped from the JS thread
+/// (i.e. from within a synchronous `#[napi]` function, or a finalizer callback - both of which
+/// N-API always runs on the JS thread). Do not read or drop this from within an `async` method.
+#[expect(unused)]
+struct JsThreadOnly<T>(T);
+
+unsafe impl<T> Send for JsThreadOnly<T> {}
+unsafe impl<T> Sync for JsThreadOnly<T> {}
```
wprzytula: 💡 Maybe let's increase usage safety by putting this into a submodule, exposing only a constructor? This way, once `JsThreadOnly` is created with some value `T` inside, no Rust code will ever be able to look into this value.
I don't know if planned use of this struct doesn't contradict with this idea of mine, though.

lib/host.js
```diff
@@ -13,14 +14,24 @@ const _rust = require("../index");
 class Host extends events.EventEmitter {
     /**
      * Creates a new Host instance.
+     *
+     * Instances of this class are constructed directly by the native driver when reading cluster metadata.
```
wprzytula: ❓ What _native driver_ is?

lib/host.js
```diff
@@ -154,12 +150,25 @@ class HostMap extends events.EventEmitter {
     #items;
     #values;
 
-    constructor() {
+    /**
+     * Creates a new HostMap instance.
+     *
+     * Instances of this class are constructed directly by the native driver when reading cluster
+     * metadata, which passes the already-built {@link Host} instances to key by their address.
```
wprzytula: 🔧 ditto: _native driver_

lib/host.js
wprzytula, line 177: 🤔 Is `values` anywhere set to anything other than `null`? This seems broken.
💭 Can't we create the Map straight in Rust, instead of creating an intermediate Array in Rust only to convert it into Map here?

lib/host.js
```diff
@@ -251,25 +260,14 @@ class HostMap extends events.EventEmitter {
...
 
 module.exports = {
     Host,
     HostMap,
 };
+
+// Registers the Host and HostMap constructors with the native driver, so that Rust can
+// construct fully-formed instances directly when reading cluster metadata.
+_rust.registerHostCtor(Host);
+_rust.registerHostMapCtor(HostMap);
```
wprzytula: 🔧 ditto: _native driver_

wprzytula (changes requested):
Posting partial review - up until `Implement host caching with ClusterSnapshot integration` (without that commit).

lib/metadata/index.d.ts
hmarszalek: > What was DataTypeInfo? Why change it?

```js
export interface DataTypeInfo {
  code: dataTypes;
  info: string | DataTypeInfo | DataTypeInfo[];
  options: {
    frozen: boolean;
    reversed: boolean;
  };
}

export interface ColumnInfo {
  name: string;
  type: DataTypeInfo;
}
```

In #454 `type` field type is changed to `ColumnInfo` in user-defined-type.js, but the change isn't applied to index.d.ts

hmarszalek: > I can see UdtField interface was defined at two sites. Why?

Probably an oversight. Originally, there was no Udt class (it was constructed as an untyped object), but UdtField was needed for ColumnInfo. When the Udt class was added, the duplication wasn't resolved.

lib/types/cql-utils.ts
hmarszalek: `ComplexType` is the wrapper over a `ColumnType` constructed on the Rust side to allow for `ToNapiValue` implementation. `ColumnInfo` is part of the Cassandra driver API we need to keep for backward compatibility. `convertComplexType` is the bridge between the two.

lib/types/cql-utils.ts
hmarszalek: > ComplexType is a wrapped Rust type, isn't it? So this introduces big overhead.

It used to be, but it was rewritten to avoid that overhead. Hence all the complicated conversion logic.

hmarszalek: > use Rust to build end JS types; avoid creating intermediate JS objects.

That is a good idea, and it is now possible with js_ctor.

lib/metadata/index.d.ts
hmarszalek: It used to be `any`, but we should type those fields.

lib/metadata/index.d.ts
hmarszalek: This is the pattern that was used before in the cassandra nodejs driver.  

But when it comes to Trace in particular... From what I understand, the Rust driver does not allow to specify consistency when fetching trace info. If that is the case, the consistency parameter should probably be deprecated.

lib/metadata/index.d.ts
hmarszalek: It was my oversight.

src/session.rs
hmarszalek: We can put it in submodule, but the `with_cluster_snapshot` will need to access it both normally and mutably to reach / update the snapshot.

lib/host.js
hmarszalek: native code  - the Rust side

lib/host.js
hmarszalek: > Is values anywhere set to anything other than null?

In the cassandra driver, the `#values` is lazily initialized and frozen on the first call to `values()`.

hmarszalek: > Can't we create the Map straight in Rust

Now that the Map is keyed by uuid that would require registering the uuid constructor instead of constructing it on the JS side from a buffer.  Rewriting the Array to the Map should be cheap, since we are reusing already constructed Host objects.

lib/host.js
```diff
@@ -152,14 +164,30 @@ class Host extends events.EventEmitter {
...
+    constructor(hosts) {
         super();
 
         this.#items = new Map();
+        this.#itemsByIp = new Map();
         this.#values = null;
 
+        for (const host of hosts) {
+            this.#items.set(host.hostId, host);
+            this.#itemsByIp.set(socketAddressKey(host.address), host.hostId);
```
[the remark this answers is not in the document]
hmarszalek: This is a real issue. We could use a buffer as the `#items` map key or write a simple collection like `DeepMap` that will use deep comparison for lookup.

lib/types/cql-utils.ts
wprzytula: Will you do this change in a separate PR?
Does it make more sense as a pre-PR or a follow-up?

src/session.rs
```diff
@@ -29,6 +32,9 @@ pub struct BatchWrapper {
 #[napi]
 pub struct SessionWrapper {
     pub(crate) inner: CachingSession,
+    /// Cache of the last `ClusterSnapshot` that was computed, alongside the `Arc<ClusterState>`
+    /// pointer it was built from.
+    cluster_snapshot: Mutex<JsThreadOnly<Option<ClusterSnapshot>>>,
 }
```
wprzytula, line 38: ❓ What's the exact semantics of the `JsThreadOnly` here? I fail to understand.

src/metadata/host.rs
```diff
@@ -1,33 +1,65 @@
...
+    cluster_state: &ClusterState,
+    env: &Env,
+) -> napi::Result<NapiRef<js_constructible_class::HostMap>> {
+    let hosts = cluster_state
+        .get_nodes_info()
+        .iter()
+        .map(|node| build_host(env, host_ctor_args(node, env)?))
+        .collect::<napi::Result<Vec<_>>>()?;
+
+    let host_map = build_host_map(env, FnArgs::from((hosts,)))?;
```
wprzytula: 💭 Here we are building a Rust Vec of JS Hosts, then we convert it to JS Array, then build a JS HostMap from that Array. This means two intermediate allocations (Vec and Array). Can we instead leverage `ffi_callback_for_each` idea from C# RS and build the HostMap using such callback straight from the iterator that yields JS Hosts?

src/utils/js_ctor.rs
```diff
@@ -12,11 +12,23 @@ use crate::utils::js_instance::JsInstance;
...
 
+/// Arguments passed to `Host(address, datacenter, rack, hostId)`.
+pub(crate) type HostCtorArgs<'a> = FnArgs<(String, Option<&'a str>, Option<&'a str>, Buffer)>;
+
+/// Arguments passed to `HostMap(hosts)`.
+///
+/// `hosts` is an array of already-built `Host` instances. The map keys each one by its own
+/// `host.address`, rather than us passing `[address, Host]` pairs, so that no per-host pair array
+/// has to be allocated on the JS heap, and the address is not sent across the boundary twice.
+type HostMapCtorArgs<'a> = FnArgs<(Vec<JsInstance<'a, js_constructible_class::Host>>,)>;
```
wprzytula: 🤔 This looks like an LLM leftover after a refactoring session. I guess LLM first suggested to send `[address, Host]` pairs and you asked it not to. This is OK - an obvious optimisation - but this comment is not needed to stay. It adds no value for the reader.

lib/host.js
```diff
@@ -2,11 +2,18 @@
...
 const { throwNotSupported } = require("./new-utils");
 const Uuid = require("./types/uuid");
 const _rust = require("../index");
 
+function socketAddressKey(address) {
+    return address.family === "ipv6"
+        ? `[${address.address}]:${address.port}`
+        : `${address.address}:${address.port}`;
+}
+
```
wprzytula: ♻️ Commit `Key HostMap by host Uuid and change Host address from string to SocketAddress`:
This obviously contains two unrelated changes. Please split into two commits.

lib/client.ts
```diff
@@ -32,7 +32,7 @@ import { HostMap } from "./host";
 
 // Imports for the purpose of type hints.
 import type { QueryOptions } from "./query-options";
-import type {
+import {
     ArrayOrObject,
     CqlValue,
     Host,
```
wprzytula, line 38: ❓ Justify the change.

src/utils/cache.rs
wprzytula, line 1: commit `Reference and NapiRef cache`:
🤔 I have no idea what this is needed and useful for after reading this commit message.
Please provide rationale and usage examples.

lib/host.js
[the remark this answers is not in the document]
wprzytula: What's your preference? Elaborate your suggested solution.

lib/metadata/index.d.ts
wprzytula: Rust Driver supports specifying consistency for fetching trace info, but only on the Session level:
https://docs.rs/scylla/latest/scylla/client/session_builder/type.SessionBuilder.html#method.tracing_info_fetch_consistency

src/session.rs
wprzytula: Then what's the semantics of `JsThreadOnly`? Are all accesses and mutations to that struct really limited to the JS thread?

lib/host.js
wprzytula: > Rewriting the Array to the Map should be cheap, since we are reusing already constructed Host objects.

We still allocate a temporary Array for no benefit.

> Now that the Map is keyed by uuid that would require registering the uuid constructor instead of constructing it on the JS side from a buffer.

Can you explain more?

wprzytula (changes requested):
This PR mixes up 3 distinct things:
1. Topology metadata.
2. Schema metadata.
3. Query traces.

Each of them should be a separate PR. This will make the review feedback loop much easier and make the required context window smaller.

IMO, query traces are of lowest importance and should come last. Also topology, being already present in the driver, makes sense to come first. That said, it's up to you what the order of those PRs will be.

lib/types/cql-utils.ts
hmarszalek: Definitely a follow-up. It's not really related to this PR, and this could possibly be a very big change.

src/session.rs
hmarszalek: We are responsible for making sure it is never accessed outside of the JS thread. There is a safety comment explaining it.

lib/host.js
hmarszalek: We have three ways to create a JS object from Rust:
1. `#[napi]` class wrapper,
2. object literal with `define_rust_to_js_convertible_object`,
3. JS class with `js_ctor`.

When keying by Uuid object, we want the key to be an instance of this class, instead of a random object literal, so the user can look up hosts by the id. To construct an actual Uuid object, we can either register the class or do it in Rust.

hmarszalek: This issue will disappear if we key the `#items` map by uuid buffer instead of uuid class. This would also solve the lookup problem (any uuid object or buffer would result in correct access).

lib/host.js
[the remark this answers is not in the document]
hmarszalek: Keying by buffer is simpler and less error-prone - it would allow the user to look up hosts with both any Uuid class and a simple id buffer. It would also allow us to avoid unnecessary array allocation when constructing HostMap. I am not sure how proper/refined this solution is. Previously, the map was keyed by a plain address string.

src/session.rs
hmarszalek: `SessionWrapper` exposes `async` methods, and for those to compile every field held by the wrapper must also be `Sync`. `JsThreadOnly` asserts `Send`/`Sync` to satisfy the compiler and allow us to carry fields like cached JS value references that are NOT `Send`/'Sync'.

lib/client.ts
hmarszalek: That block used to be a pure `import type { ... }` because everything in it was only ever used in type positions. Now we want `metadataModule` to be a value referenced at runtime, not just a type. The change should only remove the `type` for `metadataModule` and keep it for the rest of the imports.

lib/metadata/index.d.ts
hmarszalek: `tracing_info_fetch_consistency` is only settable on `SessionBuilder`. It cannot be specified or changed for each query as nodejs API requires.

lib/host.js
wprzytula: > This issue will disappear if we key the `#items` map by uuid buffer instead of uuid class. This would also solve the lookup problem (any uuid object or buffer would result in correct access).

I'm not opposed to storing Uuid as Buffer internally, if it's going to simplify correct implementation.
Just make sure that:
1. We do support correct getter semantics (and add unit testing for them).
2. The resulting implementation is free of unnecessary overhead.

lib/host.js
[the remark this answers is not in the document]
wprzytula: Will users still be able to pass instances of Uuid class for lookup if we store Buffers internally?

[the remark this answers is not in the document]
hmarszalek: Yes

src/session.rs
wprzytula: I know how we can make `JsThreadOnly` safe! Or at least safer.
You say that all accesses to `JsThreadOnly` must occur on the JS thread.
Q: How do we identify a JS thread?
A: By presence of napi env!
We can expose a single `get(&'env self, _env: &'env Env) -> &'env T` method on `JsThreadOnly` that will restrict accesses to places where we do have access to napi env. WDYT?
We can also have a mutable counterpart: `get_mut(&'env mut self, _env: &'env Env) -> & 'env mut T`.

WDYT?

hmarszalek: This is a really good idea.

lib/host.js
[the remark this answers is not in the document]
wprzytula: Why isn't it the same the other way round?
_Will users still be able to pass instances of Buffer class for lookup if we store instances of Uuid class internally?_
I understand your answer is _no_.

src/metadata/host.rs
wprzytula: I'm experimenting with this idea with Claude. Will let you know what the results are.

lib/metadata/index.d.ts
hmarszalek: I analyzed this further, and `DataTypeInfo` isn't used anywhere in the driver except for outdated declarations. The `ColumnInfo` in metadata index.d.ts is also outdated - it is declared in `types/cql-utils.ts`, and that is the implementation that is actually used. 

```ts
export class ColumnInfo {
    code: CqlType;
    info:
        | null
        | ColumnInfo
        | [ColumnInfo, ColumnInfo]
        | ColumnInfo[]
        | UdtInfo
        | [ColumnInfo, number]
        | string;
    options?: ColumnInfoOptions;
    /** Only present for vector custom types */
    customTypeName?: string;
}
```

hmarszalek: index.d.ts needs to be updated to reflect that.

hmarszalek: Moved to #513.

lib/metadata/index.d.ts
hmarszalek: Moved to #514.

lib/host.ts
```diff
@@ -2,11 +2,23 @@
...
+/**
+ * Returns the string representation of a given socket address.
+ * @param {_nodeNet.SocketAddress} address
+ * @returns {String}
+ */
+function socketAddressKey(address) {
+    return address.family === "ipv6"
+        ? `[${address.address}]:${address.port}`
+        : `${address.address}:${address.port}`;
+}
```
wprzytula, line 17: 🤔 Naming is strange. Why `socketAddressKey` to return a string repr?

wprzytula: Why not `socketAddressToString`?

wprzytula (dismissed):

lib/host.ts
```diff
@@ -33,91 +29,92 @@ function hostIdToString(buffer) {
...
-    get workloads() {
+    get workloads(): string[] | null {
         throwNotSupported("Host.workloads");
         return null;
     }
 
-    set workloads(_) {
+    set workloads(_: string[] | null) {
         throwNotSupported("Host.workloads");
     }
```
wprzytula: 🔧 In JS, we left such deprecated APIs instead of removing them to have a clearer runtime error:
"not supported" instead of "no such property".
In TS, we can achieve compile-time errors, so let's leverage them. For that, just delete all deprecated methods.

lib/host.ts
```diff
@@ -296,29 +291,29 @@ class HostMap extends events.EventEmitter {
...
      */
-    clear() {
+    clear(): void {
         throwNotSupported("HostMap.clear");
     }
 
-    inspect() {
+    inspect(): Record<string, Host> {
         return this.#hostsById;
     }
```
wprzytula, line 240: 💭 I see `readonly Host[]`. Can we write `readonly Record<string, Host>` too?

wprzytula (changes requested):

src/utils/to_napi_obj.rs
```diff
@@ -197,3 +197,40 @@ where
 }
 
 pub(crate) use define_rust_to_js_convertible_object;
+
+/// A borrowed byte slice that, when converted to a napi value, is always copied into a fresh,
+/// JS-owned `Buffer` via `napi_create_buffer_copy`.
+pub struct CopyableBuffer<'a> {
+    data: &'a [u8],
+}
```
wprzytula: 🔧 Missing explanation why use this instead of `Buffer` from napi-rs.

wprzytula (changes requested):

wprzytula (approved):

PR #504, opened 2026-07-25
Redesign of Metadata API

hmarszalek:
- Removed the no longer applicable `clearPrepared` endpoint.
- Exposed `checkSchemaAgreement` as public API
- Made Metadata class methods declarations sync (as introduced in #418), and rewrote `test/typescript/metadata-tests.ts` to match it.
- Renamed the keyspace API from misleading `refreshKeyspace(s)` to `getKeyspace(s)`.
- Removed `isMetadataSyncEnabled` option
- Fixed formatting and ordering.

PR #506, opened 2026-07-28
Constructing and pinning plain JS class instances from Rust

hmarszalek:
This PR lays the groundwork for Rust to construct instances of plain JS classes directly by calling their real JS constructor and keeping them alive across multiple native calls, without wrapping them as `#[napi]` classes. It introduces three abstractions, each building on the last:

**`JsInstance<'env, C>`** - a zero-overhead `Object<'env>` type tagged with a phantom class marker `C`, so the system can tell our JS objects apart at compile time, even though they are untyped `Object`s at the N-API level.

**`NapiRef<C>`** - an owning wrapper around a raw `napi_ref` that pins a `JsInstance<C>` against GC across calls. napi-rs's `Reference<T>` only works for `#[napi]`-wrapped class instances (it relies on their finalizer to eventually delete the reference). Plain JS objects have no such finalizer, so `NapiRef`'s `Drop` takes on that responsibility itself.

**`define_js_ctor!`** - a macro generating a `register_*_ctor` (JS hands Rust a reference to the class's constructor once per env, at module load) and a `build_*` function that calls it directly, returning a `JsInstance<'env, C>`.

## Tests

NapiRef GC pinning and multi-environment registry isolation are thoroughly tested.
Introduces GC pinning tests that verify NapiRef keeps objects alive across forced garbage collection and releases them after clearing, plus new multi-environment tests that confirm `define_js_ctor! `registry keying is truly per-environment (not just per-OS-thread) by loading the addon binary from two separate paths on the same thread.
Integrates `--expose-gc` into the test workflow for full verification without breaking existing test runs.


src/utils/js_ctor.rs
```diff
@@ -0,0 +1,87 @@
...
+        register_fn: $register_fn:ident,
+        build_fn: $build_fn:ident,
+        args: $args_ty:ty,
+        class_name: $class_name:ident,
+    ) => {
+        thread_local! {
+            /// When multiple threads are used (e.g. via `worker_threads`), each thread has its own environment
+            /// and therefore its own copy of the JS constructor reference. `thread_local!` ensures that each
+            /// thread has its own `RefCell<Option<FunctionRef<(), ()>>>`.
+            $(#[$doc])*
```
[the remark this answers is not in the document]
wprzytula: This seems valid and serious.

src/utils/js_ctor.rs
```diff
@@ -0,0 +1,87 @@
...
+/// the exact same `Args` type the `Function` handle was typed with. To solve this, we cache the
+/// constructor reference with a fully erased, always-`'static` `Args = ()` (`register_fn`'s
+/// parameter is `Function<(), ()>`), and re-tag it, at each `build_fn` call, to the call's
+/// real argument type via `Function::from_napi_value`.
+///
+/// The `Return` type parameter of the underlying `Function`/`FunctionRef` is always ignored: it is
+/// only used by `Function::call`, but we always construct instances with `Function::new_instance`,
+/// so we set it to arbitrary `()`.
+#[expect(unused)]
+macro_rules! define_js_ctor {
```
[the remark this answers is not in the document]
wprzytula: +1, please unit test our abstractions.

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,132 @@
...
+    napi_ref: sys::napi_ref,
+    env: sys::napi_env,
+    _class: PhantomData<C>,
+}
+
+impl<C> NapiRef<C> {
+    /// Pins `value` (a `C` instance) against garbage collection with a fresh, strong (ref count 1)
+    /// `napi_ref`, taking ownership of the reference's lifetime.
+    #[allow(dead_code)]
+    pub(crate) fn new(env: &Env, value: JsInstance<'_, C>) -> napi::Result<Self> {
```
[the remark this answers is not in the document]
wprzytula: +1

wprzytula (commented):
This PR should start with commit [JsInstance](https://github.com/scylladb/nodejs-rs-driver/pull/506/changes/f02b2f5845cb749a68abca1b7af1871f8cacdbff). All previous commits are not relevant, so stacking this PR on #504 brings no benefit.

src/utils/mod.rs
```diff
@@ -1,5 +1,7 @@
 pub mod from_napi_obj;
+pub mod js_ctor;
 pub mod js_instance;
+pub mod napi_ref;
```
wprzytula, line 4: `js_ctor` mod should be added in a relevant commit. This commit does not compile, because there's no source file for the module.

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,132 @@
...
+/// is driven exclusively by the calls this type makes. Consequently the last `NapiRef` to be dropped always
+/// observes strong count 0 and deletes the reference, so `napi_delete_reference` is guaranteed to run for
+/// every `napi_ref` this type creates, exactly once.
+///
+/// # Safety
+/// Every `NapiRef` must only be created, read, cloned, or dropped on the JS thread that owns the `Env` it was created with.
+pub struct NapiRef<C> {
+    napi_ref: sys::napi_ref,
+    env: sys::napi_env,
+    _class: PhantomData<C>,
```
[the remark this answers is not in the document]
hmarszalek: We will not be using a `thread_local`, but a designed `NapiRefCache` that implements finalization. This does not apply.

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,132 @@
...
+/// Unlike a raw `napi_ref`, which has no `Drop` semantics of its own, `NapiRef` releases the reference
+/// it owns automatically when dropped (`napi_reference_unref`, then napi_delete_reference if the strong
+/// count has reached zero).
+///
+/// The count of the underlying `napi_ref` always equals the number of currently live `NapiRef`s pinning it:
+/// `new` starts it at 1, `try_clone` increments it by handing out one more `NapiRef`, and `Drop` decrements it
+/// by exactly one. Node/V8 never calls `napi_reference_ref`/`napi_reference_unref` on our behalf - the count
+/// is driven exclusively by the calls this type makes. Consequently the last `NapiRef` to be dropped always
+/// observes strong count 0 and deletes the reference, so `napi_delete_reference` is guaranteed to run for
+/// every `napi_ref` this type creates, exactly once.
```
wprzytula: 👍 Thanks for this explanation!

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,132 @@
...
+    napi_ref: sys::napi_ref,
+    env: sys::napi_env,
+    _class: PhantomData<C>,
+}
+
+impl<C> NapiRef<C> {
+    /// Pins `value` (a `C` instance) against garbage collection with a fresh, strong (ref count 1)
+    /// `napi_ref`, taking ownership of the reference's lifetime.
+    #[allow(dead_code)]
+    pub(crate) fn new(env: &Env, value: JsInstance<'_, C>) -> napi::Result<Self> {
```
wprzytula, line 32: ⛏️ Use `expect` instead of `allow`.

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,132 @@
...
+            env: env.raw(),
+            _class: PhantomData,
+        })
+    }
+
+    /// Retrieves the `C` instance currently pinned by this reference, as a `JsInstance<'a, C>` whose
+    /// lifetime is tied to both the borrow of `self` and the borrow of `env` passed in - preventing
+    /// the returned object from outliving either this `NapiRef` or the `Env` it was retrieved
+    /// through. `env` must refer to the same environment this `NapiRef` was created with.
+    #[allow(dead_code)]
```
wprzytula: ⛏️ Ditto: `expect`

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,132 @@
...
+            env.raw(),
+            "NapiRef::get called with an Env different from the one it was created with",
+        );
+        let mut result = ptr::null_mut();
+        check_status!(
+            unsafe { sys::napi_get_reference_value(env.raw(), self.napi_ref, &mut result) },
+            "Failed to get N-API reference value",
+        )?;
+        Ok(JsInstance::from_object(Object::from_raw(env.raw(), result)))
+    }
```
wprzytula, line 67: ❓ Why do we need to pass `env` to `get()`? We could use the one in `self` instead.
I see you're doing an equality assertion. Is it really needed? (I don't know, please think about this).

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,132 @@
...
+            _class: PhantomData,
+        })
+    }
+
+    /// Retrieves the `C` instance currently pinned by this reference, as a `JsInstance<'a, C>` whose
+    /// lifetime is tied to both the borrow of `self` and the borrow of `env` passed in - preventing
+    /// the returned object from outliving either this `NapiRef` or the `Env` it was retrieved
+    /// through. `env` must refer to the same environment this `NapiRef` was created with.
+    #[allow(dead_code)]
+    pub(crate) fn get<'a>(&'a self, env: &'a Env) -> napi::Result<JsInstance<'a, C>> {
```
wprzytula: 💭 Why do we constrain `JsInstance` with the lifetime of `&Env`, while we don't do the same for `NapiRef` (storing `sys::napi_env` instead that does not have a lifetime parameter)?

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,132 @@
...
+            "NapiRef converted to a napi_value in an Env different from the one it was created with",
+        );
+        let mut result = ptr::null_mut();
+        check_status!(
+            unsafe { sys::napi_get_reference_value(env, val.napi_ref, &mut result) },
+            "Failed to get N-API reference value",
+        )?;
+        Ok(result)
+    }
+}
```
wprzytula: 🤔 Is this correct? If we drop the `NapiRef` when this returns, then `napi_ref` is dropped too. Then what keeps the returned `napi_value` alive?

src/utils/js_ctor.rs
```diff
@@ -0,0 +1,95 @@
+/// Zero-sized marker types naming each JS class that Rust constructs directly.
+/// They exist only to parametrize `JsInstance` and, in turn, `NapiRef`.
+pub mod js_constructible_class {}
```
wprzytula: ❓ What's this? What an empty module is useful for?

src/utils/js_ctor.rs
```diff
@@ -0,0 +1,95 @@
...
+        static_name: $static_name:ident,
+        register_fn: $register_fn:ident,
+        build_fn: $build_fn:ident,
+        args: $args_ty:ty,
+        class_name: $class_name:ident,
+    ) => {
+        $(#[$doc])*
+        static $static_name: LazyLock<
+            Mutex<HashMap<usize, FunctionRef<(), ()>>>,
+        > = LazyLock::new(Default::default);
```
wprzytula: ❓ Why do we need both `LazyLock` and `Mutex`? This is double synchronization.
💭 I guess that `Mutex` isn't enough on its own because `HashMap::new` is not a `const` function, so it can't be called in `static`. Perhaps `Mutex<Option<HashMap<...>>>` would be acceptable though?

wprzytula: 🤔 Haven't we been using thread-local storage before? Why store the registry in a thread-shared `static` storage instead?

src/utils/js_ctor.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// Registers the JS class constructor, so that Rust can later construct fully-formed JS instances
+        /// directly. It is called exactly once per N-API environment (i.e. once per `worker_threads` worker,
+        /// and once on the main thread), by the corresponding module on load, before any cluster metadata
+        /// is accessed in that environment.
+        #[napi]
+        pub fn $register_fn(ctor: Function<(), ()>, env: Env) -> napi::Result<()> {
```
wprzytula, line 66: ❓ How do we ensure that this is indeed called on the module load, as the docstring of `define_js_ctor!` says?

src/utils/js_ctor.rs
```diff
@@ -0,0 +1,95 @@
...
+                }
+            }
+            // Register the cleanup hook - which removes this environment's entry, dropping its `FunctionRef`,
+            // exactly when this specific environment is torn down, rather than leaving it to linger past its
+            // environment's lifetime or be silently reused by a later environment that happens to be allocated
+            // at the same address.
+            env.add_env_cleanup_hook(key, |key| {
+                $static_name.lock().unwrap().remove(&key);
+            })?;
+            $static_name.lock().unwrap().insert(key, ctor_ref);
```
wprzytula: 🤔 Can't we perform the insertion while still holding the lock that was taken the first time? This way we won't take it twice.
Or maybe is it done this way on purpose?

src/utils/js_ctor.rs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        /// Constructs a JS class instance directly, by calling its constructor, registered via register_fn.
+        /// The returned object's lifetime is tied to the `&'env Env` borrow passed in: the `napi_value` is
+        /// only guaranteed to remain a valid, live GC root for as long as that handle scope is on the stack.
+        /// This prevents the returned object from silently outliving the native call.
+        pub(crate) fn $build_fn<'env>(
+            env: &'env Env,
+            args: $args_ty,
+        ) -> napi::Result<JsInstance<'env, js_constructible_class::$class_name>> {
```
wprzytula, line 104: 💭 This hints that `js_constructible_class` module will contain some entities. Not yet though.

src/utils/js_ctor.rs
```diff
@@ -0,0 +1,95 @@
...
+                        " constructor is not registered yet; ensure lib/metadata/*.js has been ",
+                        "loaded before accessing cluster metadata",
+                    ))
+                })?;
+                ctor_ref.borrow_back(env)?
+            };
+            // Re-tag the constructor's phantom `Args` type parameter to this call's real
+            // (possibly borrowed) argument type.
+            let ctor: Function<'_, $args_ty, ()> =
+                unsafe { Function::from_napi_value(env.raw(), ctor.raw())? };
```
wprzytula, line 120: 💭 I don't understand why we need this cast. This looks suspicious.

src/utils/js_ctor.rs
```diff
@@ -0,0 +1,95 @@
...
+            let instance: Unknown = ctor.new_instance(args)?;
+            Ok(JsInstance::from_object(Object::from_raw(env.raw(), instance.raw())))
```
wprzytula, line 122: ❓ What is `Unknown`? How does it relate to `napi_value` and `Object`?
❓ Why doesn't `ctor.new_instance()` return `Object`?

src/tests/napi_ref_tests.rs
wprzytula, line 1: 🔧 Please add a more exhaustive module docstring that explain what this tests and how. What does `TestPin` represent?

src/utils/js_ctor.rs
```diff
@@ -1,6 +1,21 @@
...
+
+use crate::utils::js_instance::JsInstance;
+
 /// Zero-sized marker types naming each JS class that Rust constructs directly.
 /// They exist only to parametrize `JsInstance` and, in turn, `NapiRef`.
-pub mod js_constructible_class {}
+pub mod js_constructible_class {
+    /// Test-only marker for `TestPin(name, value)`, used by `crate::tests::napi_ref_tests`.
+    pub enum TestPin {}
+}
```
wprzytula, line 15: 💭 I dislike that we need to add markers here apart from just invoking `define_js_ctor!`.

src/tests/napi_ref_tests.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use crate::errors::{ConvertedResult, JsResult, with_custom_error_sync};
+use crate::utils::js_ctor::{build_test_pin, js_constructible_class};
+use crate::utils::js_instance::JsInstance;
+use crate::utils::napi_ref::NapiRef;
+
+thread_local! {
+    static PINNED: RefCell<Option<NapiRef<js_constructible_class::TestPin>>> =
+        const { RefCell::new(None) };
+}
```
wprzytula, line 18: 🤔 I don't understand why we use thread local storage for the test at all. It complicates understanding of the test, and I don't see why we need it.

src/tests/napi_ref_tests.rs
```diff
@@ -0,0 +1,86 @@
...
+pub fn tests_build_pinned<'env>(
+    name: String,
+    value: i32,
+    env: &'env Env,
+) -> JsResult<JsInstance<'env, js_constructible_class::TestPin>> {
+    with_custom_error_sync(|| {
+        let instance = build_test_pin(env, FnArgs::from((name.as_str(), value)))?;
+        ConvertedResult::Ok(instance)
+    })
+}
```
wprzytula: 😕 I'm lost in this naming. We build a `TestPin` but we don't pin it? We don't pin a Pin?

src/utils/js_ctor.rs
```diff
@@ -0,0 +1,116 @@
...
+                    " constructor is not registered yet; ensure lib/metadata/*.js has been ",
+                    "loaded before accessing cluster metadata",
+                ))
+            })?;
+            let ctor: Function<'_, (), ()> = ctor_ref.borrow_back(env)?;
+            // Re-tag the constructor's phantom `Args` type parameter to this call's real
+            // (possibly borrowed) argument type.
+            let ctor: Function<'_, $args_ty, ()> =
+                unsafe { Function::from_napi_value(env.raw(), ctor.raw())? };
+            let instance: Unknown = ctor.new_instance(args)?;
```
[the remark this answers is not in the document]
wprzytula: Makes sense, no?

test/unit/napi-ref-tests.js
```diff
@@ -0,0 +1,101 @@
...
+        });
+
+        it("try_clone() should resolve to the same underlying object as get()", function () {
+            rust.testsPinAndStore("charlie", 30);
+            const viaGet = rust.testsGetPinned();
+            const viaClone = rust.testsClonePinned();
+            assert.strictEqual(viaGet, viaClone);
+            assert.deepEqual(viaClone, { name: "charlie", value: 30 });
+        });
+    });
```
wprzytula: 💭 By mixing `NapiRef` testing with `define_js_ctor!` testing, it's hard to grasp what's needed for what. Could you please introduce testing of one part first (e.g., no `NapiRef` yet) in one commit, then add the other part in another commit? Now I fail to see how this all interconnects.

wprzytula (changes requested):

src/utils/js_ctor.rs
hmarszalek: Add comment explaining why

src/utils/napi_ref.rs
hmarszalek: Because they wrap two different N-API concepts. `JsInstance` wraps an `Object<'env>` - a `napi_value`, which is a local V8 handle. It is only valid until the current native call returns. `NapiRef` wraps a `napi_ref`, a persistent handle whose entire purpose is to survive across calls and scopes. Tying `NapiRef` to `&Env` would defeat the purpose of its existence.

src/utils/napi_ref.rs
hmarszalek: self.env is a bare pointer with no lifetime to borrow from.

src/utils/js_ctor.rs
hmarszalek: `new_instance` returns `Unknown` because napi-rs has no type describing a constructor's output. `Unknown` and `Object` are structurally identical.

src/utils/napi_ref.rs
```diff
@@ -0,0 +1,129 @@
...
+    napi_ref: sys::napi_ref,
+    env: sys::napi_env,
+    _class: PhantomData<C>,
+}
+
+impl<C> NapiRef<C> {
+    /// Pins `value` (a `C` instance) against garbage collection with a fresh, strong (ref count 1)
+    /// `napi_ref`, taking ownership of the reference's lifetime.
+    pub(crate) fn new(env: &Env, value: JsInstance<'_, C>) -> napi::Result<Self> {
+        let napi_val = unsafe { ToNapiValue::to_napi_value(env.raw(), value) }?;
```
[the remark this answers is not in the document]
wprzytula: Makes sense. If `JsInstance` carries a lifetime anyway, it can hold `&Env`.

src/utils/napi_ref.rs
wprzytula: > I see you're doing an equality assertion. Is it really needed? (I don't know, please think about this).

hmarszalek: `NapiRef` holds a raw pointer with `napi_env`. If `get` reconstructed into its own `Env` internally (`Env::from_raw(self.env)`), that would be a fresh local value.

hmarszalek: The assertion is not necessary, but it guards against UB if somewhere we would accidentally pass the wrong environment.

src/utils/js_instance.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub(crate) fn env_raw(&self) -> sys::napi_env {
+        self.object.value().env
+    }
+}
+
+impl<C> ToNapiValue for JsInstance<'_, C> {
+    unsafe fn to_napi_value(env: sys::napi_env, val: Self) -> napi::Result<sys::napi_value> {
+        unsafe { ToNapiValue::to_napi_value(env, val.object) }
+    }
+}
```
wprzytula, line 50: 💭 Why not panic here if `env` is different from `self.object.value().env`?

wprzytula: Or perhaps napi-rs does so itself?

wprzytula (changes requested):
Please add missing commit messages.

src/utils/js_instance.rs
hmarszalek: Added assertion

wprzytula (approved):

PR #513, opened 2026-08-09
Schema metadata

### LINKED ISSUE #518: Support for Schema Metadata
Rewrite in TypeScript and implement these classes (#268):
- [x] KeyspaceMetadata
- [x] TableMetadata
- [x] MaterializedView
- [x] UserDefinedTypes

Implement following methods from Metadata class (#261):
- [x] getKeyspace
- [x] getKeyspaces
- [x] getTable
- [x] getMaterializedView
- [x] getUdt

hmarszalek:
Implements `client.metadata` for schema: keyspaces, tables, materialized views and user-defined types, backed by the Rust driver's in-memory cluster state. All lookups are synchronous.

## Old public API

Everything was asynchronous, because each lookup read the system_schema tables on demand. getTable, getMaterializedView and getUdt all took a callback or returned a promise.

### Keyspaces

`metadata.keyspaces` was a plain object keyed by keyspace name. It was kept up to date with `refreshKeyspaces()` – nothing was populated until you called it, and it queried `system_schema` to do so.

`KeyspaceMetadata` carried its own name, `durableWrites`, and replication as two loosely related properties:

```js
ks.strategy;         // "org.apache.cassandra.locator.NetworkTopologyStrategy"
ks.strategyOptions;  // { dc1: 3, dc2: 1 }
```

strategy was the server's Java class name as a string, and `strategyOptions` an untyped bag whose meaningful keys depended on which class the string named – `replication_factor` for `SimpleStrategy`, datacenter names for `NetworkTopologyStrategy`.

### Tables and columns

Shaped for iteration, with a parallel index for lookup:

```js
table.columns;          // [ { name: "pk1", type: { code, info } }, … ]
table.columnsByName;    // { pk1: { name, type }, … }
table.partitionKeys;    // [ { name: "pk1", … }, { name: "apk2", … } ]
table.clusteringKeys;   // [ { name: "ck", … } ]
table.clusteringOrder;  // [ "ASC" ]
```

So a column object appeared three times over: in the array, in the by-name index, and again inside the key arrays. Alongside came the table's full storage configuration – caching, defaultTtl, bloom filter false-positive chance, comment, compaction class and options, compression, gcGraceSeconds, id, extensions, indexes, and more.

### Materialized views

`getMaterializedView(keyspace, name, cb)`, extending the table metadata with the base tableName. DataStax also exposed whereClause and includeAllColumns.

### User-defined types

`getUdt(keyspace, name, cb)`, returning a plain object:

```js
udtInfo.name;    // "udt1"
udtInfo.fields;  // [ { name, type }, … ]
```

## New public API

`Client`
- `client.metadata` – a `Metadata`, initialized during `connect()`.

`Metadata`
- constructor now accepts `SessionWrapper` as a parameter
- Implemented `Metadata` methods (`getKeyspace`, `getKeyspaces`, `getTable`, `getMaterializedView`, `getUdt`)

`KeyspaceMetadata`
- `strategy: Strategy, durableWrites: boolean`
- `tables`, `views`, `udts` – `Readonly<Record<string, T>>`
- `getTable(name)`, `getMaterializedView(name)`, `getUdt(name)` – `T | null`

Replication strategy – `Strategy` is a discriminated union over `SimpleStrategy` (`replicationFactor`), `NetworkTopologyStrategy` (`datacenterRepfactors`), `LocalStrategy`, and `OtherStrategy` (`name`, `data`), discriminated by `kind: StrategyKind`. `StrategyKind` is exported as a runtime value; the four variants are exported as types only, so kind is the way to tell them apart.

Schema objects – `TableMetadata` (`columns`, `partitionKey`, `clusteringKey`, `partitioner`), `ColumnMetadata` (`type`, `kind`), `ColumnKind` (`Regular`/`Static`/`ClusteringKey`/`PartitionKey`), `MaterializedView` (extends `TableMetadata`, adds `tableName`), `Udt` (`name`, `keyspace`, `fields`), `UdtField` (`name`, `type`). `partitionKey`/`clusteringKey` are arrays of column names, and columns is keyed by name.

Everything above is read-only to full depth: fields cannot be reassigned, the records cannot gain entries, and the arrays are readonly. This is enforced at compile time only – readonly is erased at runtime.

## How it is implemented

`#[napi]` class – `KeyspaceMetadata`. `KeyspaceWrapper` is a real napi class (`js_name = "KeyspaceMetadata"`) holding the Rust Keyspace plus three lazy caches. Its properties are `#[napi(getter)]`, which is what makes them read-only at runtime as well as in the type. It is handed back as a cached `Reference`, so repeated lookups return the same JS object.

Registered JS constructors – everything else. `TableMetadata`, `ColumnMetadata`, `MaterializedView`, `Udt`, `UdtField` and the four strategy classes are ordinary TypeScript classes in `lib/metadata/`, each registers its constructor through `define_js_ctor!`.

`NamedMap` – the record-shaped getters. `tables`, `views`, `udts` and `getAllKeyspaces` build a plain JS object keyed by name.

## Rust-side caching

- Introduces `JsCache<E>`, a lazily-populated cache storing converted JS objects per key, so every lookup path returns the same underlying object. A converted value can be kept alive in one of two ways: `#[napi]` class instances are stored as a `Reference<V>`, while the plain JS objects built through `js_ctor` have to be pinned behind a raw `napi_ref` (`NapiRef<C>`). That is the only difference between the two, so it is abstracted behind the `CacheEntry` trait - `store()` pins a value, `load()` produces a fresh handle to it - leaving one cache implementation shared by both. The `ReferenceCache<V>` and `NapiRefCache<C>` aliases name the two instantiations.
- `ClusterSnapshot` now caches `KeyspaceWrappers` by name and each `KeyspaceWrapper` caches its own tables/views/udts the same way.

## Tests, examples, and documentation

Integration tests for all implemented methods, including caching verification - same object across repeated calls, same object whether reached directly or through the keyspace, cache invalidated on ALTER KEYSPACE.

Adds examples showing how to fetch schema metadata and documentation explaining the exact new API and how it differs from the previous one.

Fixes: #518 

## Pre-review checklist


src/utils/cache.rs
wprzytula, line 1: ♻️  Both caches look extremely similar. Can we deduplicate them into one generic structure and instantiate it for both kind of references (`Reference` and `NapiRef`)?

lib/metadata/index.d.ts
```diff
@@ -152,11 +152,17 @@ export enum StrategyKind {
...
+  kind: StrategyKind;
+  // Only set when `kind` is `StrategyKind.SimpleStrategy`, null otherwise.
+  replicationFactor: number | null;
+  // Only set when `kind` is `StrategyKind.NetworkTopologyStrategy`, null otherwise.
+  datacenterRepfactors: { [datacenter: string]: number } | null;
+  // Only set when `kind` is `StrategyKind.Other`, null otherwise.
+  name: string | null;
+  // Only set when `kind` is `StrategyKind.Other`, null otherwise.
+  data: { [key: string]: string } | null;
+}
```
wprzytula: 🤔 This is relaxation of the type definition, which loses guarantees that a `Strategy` instance gives. Before, we could pattern match and we would clearly know which variant we are. Now, all fields are nullable and no information is carried on the type level (just written in the comments). I see this as a regression.

src/utils/js_ctor.rs
```diff
@@ -54,6 +44,16 @@ pub(crate) type HostCtorArgs<'a> = FnArgs<(
...
+/// Arguments passed to `Strategy(kind, replicationFactor, datacenterRepfactors, name, data)`.
+/// Only the field(s) relevant to `kind` are set (`Some`); the rest are `None`.
+type StrategyCtorArgs<'a> = FnArgs<(
+    u32,
+    Option<u32>,
+    Option<HashMap<&'a str, u32>>,
+    Option<&'a str>,
+    Option<HashMap<&'a str, &'a str>>,
+)>;
+
```
wprzytula, line 89: commit `Keyspace Strategy declaration`: why does this commit move this chunk of code?

lib/metadata/keyspace-metadata.ts
```diff
@@ -121,16 +122,75 @@ class KeyspaceMetadata {
...
+    }
+
+    /**
+     * Creates a KeyspaceMetadata instance backed by a native keyspace wrapper.
+     * @internal
+     * @ignore
+     */
+    static fromRust(keyspaceWrapper: rust.KeyspaceWrapper): KeyspaceMetadata {
+        return new KeyspaceMetadata(keyspaceWrapper);
     }
```
wprzytula: 💭 If `KeyspaceMetadata` is now a wrapper over Rust struct, then why even add a new wrapping boilerplate in JS? napi-rs already provides required functionality with the `#[napi]` macro. Why reimplement `udts()` getter in JS instead of using the `#[napi]`-exposed one? Why reassign `udts` from `this.#wrapper.udts` to `this.#udts`?

src/metadata/state.rs
```diff
@@ -312,4 +359,30 @@ impl KeyspaceWrapper {
...
+                    .views
+                    .iter()
+                    .map(|(name, view)| {
+                        Ok((name.clone(), convert_rust_materialized_view(env, view)?))
+                    })
+                    .collect::<ConvertedResult<
+                        HashMap<String, JsInstance<'env, js_constructible_class::MaterializedView>>,
+                    >>()
+            })?;
+            ConvertedResult::Ok(NamedMap::new(views.into_iter().collect()))
```
wprzytula: ❓ Why do you need to `collect()`, then `into_iter()`, then `collect()` again, just to pass to `NamedMap::new()`, which again does not need a materialized collection (at least conceptually, I believe)?

examples/README.md
```diff
@@ -43,8 +43,10 @@ Those examples are present in this directory and have the `.broken` in the name
   - [Insert and retrieve using the Mapper](mapper/mapper-insert-retrieve.js) -->
 - Metadata
   - [Get hosts information](metadata/metadata-hosts.js)
-<!--   - [Get keyspaces information](metadata/metadata-keyspaces.js)
-  - [Get table information](metadata/metadata-table.js) -->
+  - [Get keyspaces information](metadata/metadata-keyspaces.js)
+  - [Get table information](metadata/metadata-table.js)
+  - [Get user-defined type information](metadata/metadata-udt.js)
+  - [Get materialized view information](metadata/metadata-materialized-view.js)
```
wprzytula: ♻️ I think all schema naturally belongs to one example. This separation is artificial IMO.

wprzytula (changes requested):

src/metadata/state.rs
hmarszalek: It is a mistake, just views should be passed to NamedMap. Same for the rest.

src/utils/cache.rs
hmarszalek: Should be possible.

lib/metadata/index.d.ts
hmarszalek: @wprzytula, in the DSx driver, the strategy was a string plus a separate strategyOption. We could stop routing it through js_ctor and create a plain literal with the relevant keys. Or we could create separate constructors for each strategy.

lib/metadata/keyspace-metadata.ts
hmarszalek: From what I understand, cross-language calls are expensive. This way, udts need to be fetched from Rust once, and on subsequent calls, there is no need to call into Rust at all.

wprzytula: OK.

lib/metadata/keyspace-metadata.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    /**
+     * Creates a KeyspaceMetadata instance backed by a native keyspace wrapper.
+     * @internal
+     * @ignore
+     */
+    static fromRust(keyspaceWrapper: rust.KeyspaceWrapper): KeyspaceMetadata {
+        return new KeyspaceMetadata(keyspaceWrapper);
+    }
```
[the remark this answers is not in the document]
wprzytula: @hmarszalek Is this still true?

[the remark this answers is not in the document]
hmarszalek: Partly... The conclusion still holds – `getKeyspace()` returns different objects on each call.

lib/metadata/keyspace-metadata.ts
hmarszalek: OTOH, using `#[napi]` keyspace instead of a wrapper would help preserve object identity. Subsequent calls to getKeyspace / getKeyspaces would return the same underlying object.

lib/metadata/keyspace-metadata.ts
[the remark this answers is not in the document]
wprzytula: So this not only defeats our goal of _identity_, but also means that we are having unnecessary work (overhead) on each lookup. Unacceptable.

lib/metadata/keyspace-metadata.ts
wprzytula: This likely is the way to go.

src/utils/cache.rs
wprzytula, line 1: 🔧 A lot of items here are `pub`. Won't `pub(crate)` suffice?

src/utils/cache.rs
```diff
@@ -0,0 +1,162 @@
...
+///
+/// # Concurrency
+///
+/// These caches only ever live inside N-API class instances that are exclusively accessed on the
+/// JS thread. A plain `RefCell` is therefore sufficient and no locking is required.
+pub struct JsCache<E: CacheEntry> {
+    map: RefCell<HashMap<String, E>>,
+    /// Set once `get_or_init_all`'s builder has run to completion.
+    complete: Cell<bool>,
+}
```
wprzytula: 💭  Shouldn't both `map` and `complete` be under the same `RefCell` to prevent logical bugs / panics?

src/utils/cache.rs
```diff
@@ -0,0 +1,162 @@
...
+            let built = f()?
+                .into_iter()
+                .map(|(key, value)| Ok((key, E::store(env, value)?)))
+                .collect::<ConvertedResult<Vec<_>>>()?;
+            let mut map = self.map.borrow_mut();
+            for (key, entry) in built {
+                map.entry(key).or_insert(entry);
+            }
+            self.complete.set(true);
+        }
```
wprzytula: 🔧 I think we don't need to `collect()` here.

src/metadata/state.rs
```diff
@@ -29,3 +32,78 @@ impl ClusterSnapshot {
...
+#[napi(object)]
+pub struct OtherStrategy {
+    /// `kind` is `StrategyKind.Other`.
+    #[napi(ts_type = "3")]
+    pub kind: u32,
+    /// Name of the strategy, as reported by the server.
+    pub name: String,
+    /// Additional parameters of the strategy, which the driver does not interpret.
+    pub data: HashMap<String, String>,
+}
```
wprzytula: ❓ What is the performance overhead that use of `napi(object)` involves?

src/metadata/state.rs
```diff
@@ -171,6 +226,33 @@ fn convert_rust_strategy(
...
+                };
+                ks.tables.get_or_init(env, &table, || {
+                    let Some(rust_table) = ks.inner.tables.get(&table) else {
+                        return ConvertedResult::Ok(None);
+                    };
+                    ConvertedResult::Ok(Some(convert_rust_table(env, rust_table)?))
+                })
+            })
+        })
+    }
```
wprzytula, line 411: 💭  I can't judge, based on just `pub` here, whether this is part of the driver's public API or not. This is crucial (one must be extra careful not to break the API). Same goes for `KeyspaceWrapper::tables`.
Consider annotating each `#[napi]`-exposed `pub` function with marker whether it's public in JS or not.

src/metadata/state.rs
```diff
@@ -224,4 +306,27 @@ impl KeyspaceWrapper {
...
+                    .tables
+                    .iter()
+                    .map(|(name, table)| Ok((name.clone(), convert_rust_table(env, table)?)))
+                    .collect::<ConvertedResult<
+                        HashMap<String, JsInstance<'env, js_constructible_class::TableMetadata>>,
+                    >>()
+            })?;
+            ConvertedResult::Ok(NamedMap::new(tables))
+        })
+    }
```
wprzytula, line 511: 🤔  It's a bit weird to me that on `SessionWrapper` we only have single-table getter, whereas on `KeyspaceWrapper` we only have all-tables getter. Why not both at both sites, or both on `KeyspaceWrapper` only?

lib/metadata/index.d.ts
```diff
@@ -143,17 +146,6 @@ export interface SchemaFunction {
...
-  type: ColumnInfo;
-}
-
-export interface Udt {
-  name: string;
-  keyspace: string;
-  fields: UdtField[];
-}
-
 export class Metadata {
```
wprzytula: 🤔 I got confused why you delete these definitions as part of implementing UDT support. Only then I realised that instead you use the definitions from `user-defined-type.ts`. Does this mean that you could do it as soon as you TS-fied that source file? If so, let's do this type stub switch in a separate commit.

wprzytula (changes requested):
🔧 Please list the full new introduced API in the PR cover letter, including information on how those APIs are implemented (`#[napi]`, `#[napi(object)]`, exposed constructors) + performance considerations of the implementation.

src/metadata/state.rs
hmarszalek: `SessionWrapper` used by JS `Metadata` class allows fetching only a single `Table`/`View`/`Udt`. `KeyspaceWrapper` used by JS `KeyspaceMetadata` class, only allows fetching all `table` at once.

hmarszalek: We could easily allow for both Metadata and KeyspaceMetadata to allow both singular and plural fetch, if you think that is a good idea.

wprzytula: First please explain why the discrepancy even exists. Why only singular in Metadata and only plural in KeyspaceMetadata.

hmarszalek: `Metadata` class API (only singular) is taken from DSx driver, which offered only singular endpoints.
`KeyspaceMetadata` was entirely redesigned, user could only access the `tables[]` array field (like in Rust driver). We could, and it is probably a good idea, add a way to fetch singular table from `KeyspaceMetadata`. I don't know about adding this option to `Metadata class`.

hmarszalek:
```js
export class Metadata {
  getKeyspace(name: string): KeyspaceMetadata | null; // <- new, used to be a `keyspaces[]` field

  getKeyspaces(): Map<string, KeyspaceMetadata>; // <-new

  getTable( // <- old
    keyspaceName: string,
    name: string,
  ): TableMetadata | null;

  getMaterializedView( // <- old
    keyspaceName: string,
    name: string,
  ): MaterializedView | null;

  getUdt(keyspaceName: string, name: string): Udt | null; // <- old
}
```

wprzytula: So my suggestion is:
- keep the singular endpoint in `Metadata` for backwards compat;
- add both singular and plural endpoints in `KeyspaceMetadata`.

hmarszalek:
```js
export interface KeyspaceMetadata {
  strategy: Strategy;
  durableWrites: boolean;
  tables: { [name: string]: TableMetadata };
  views: { [name: string]: MaterializedView };
  udts: { [name: string]: Udt };
}
```

wprzytula: Shall we mark the whole metadata API as readonly? `Metadata`, `KeyspaceMetadata`, `TableMetadata`, etc.

src/metadata/state.rs
hmarszalek: I thought it would simplify the code, but after checking the performance, it is much slower than using `js_ctor`. I am currently updating the code to use custom constructors instead.

wprzytula (changes requested):
commit `Make metadata API readonly`: this is a fix to previous commits. It should be introduced in the first place, not come later as a fix.

wprzytula:
_Pre-review checklist_ misses 3 points checked. Make sure you did cover all of them, and if not, justify that in the cover letter.

src/metadata/state.rs
```diff
@@ -73,3 +156,61 @@ fn convert_rust_strategy<'env>(env: &'env Env, strategy: &Strategy) -> napi::Res
...
+        &self,
+        env: &Env,
+    ) -> JsResult<NamedMap<String, Reference<KeyspaceWrapper>>> {
+        with_custom_error_sync(|| {
+            self.with_cluster_snapshot(env, |snapshot: &ClusterSnapshot| {
+                let keyspaces = snapshot.all_keyspace_wrappers(env)?;
+                ConvertedResult::Ok(NamedMap::new(keyspaces))
+            })
+        })
+    }
```
wprzytula, line 444: 💭 `KeyspaceWrapper`s are cached, but the `NamedMap` of `KeyspaceWrapper`s is not - it's constructed eagerly on each call. Do you think it's OK?

lib/metadata/index.ts
```diff
@@ -112,15 +112,15 @@ class Metadata {
...
 
     /**
      * Gets all keyspace metadata.
      * @returns A map of all keyspaces indexed by name.
      */
-    getKeyspaces(): Map<string, KeyspaceMetadata> {
-        throw new Error("TODO: Not implemented");
+    getKeyspaces(): ReadonlyMap<string, KeyspaceMetadata> {
+        return new Map(Object.entries(this.#rustClient.getAllKeyspaces()));
     }
```
wprzytula, line 124: ❓ What's the model of `Metadata` here?
In C# RS Driver, we modelled `Metadata` as a snapshot, am I correct?
Here, `Metadata` always gives the newest available metadata from `Session`.
OTOH, once we have a `KeyspaceMetadata` at hand, it's a snapshot (it's linked to a fixed `ClusterState`).

Is this in line with your mental model? Isn't this confusing? WDYT?

wprzytula, line 124: 🔧 Not only does Rust construct a new `NamedMap` (a plain JS object) on each call to `getKeyspaces()`, but also JS converts that object to another (`new Map`). This seems expensive.
1. Let's delete the middle-man (`object` on the Rust side), or another way remove that allocation.
2. Optionally (please assess whether it makes sense): cache the map/object on the Rust side. But JS user likely will call `getKeyspaces()` only once and then keep the returned value, so maybe this is not important and over-optimisation.

wprzytula: Or maybe let's ignore both. Idk, please decide.

src/metadata/state.rs
```diff
@@ -120,6 +124,70 @@ impl KeyspaceWrapper {
...
+    fn cached_table<'env>(
+        &self,
+        env: &'env Env,
+        name: &str,
+    ) -> ConvertedResult<Option<JsInstance<'env, js_constructible_class::TableMetadata>>> {
+        self.tables
+            .get_or_init(env, name, || match self.keyspace().tables.get(name) {
+                Some(table) => ConvertedResult::Ok(Some(convert_rust_table(env, table)?)),
+                None => ConvertedResult::Ok(None),
+            })
```
wprzytula: ♻️ Functional style:
```suggestion
           .get_or_init(env, name, || {
               let table = self.keyspace()
                   .tables.get(name)
                   .map(|table| convert_rust_table(env, table))
                   .transpose()?;
               ConvertedResult::Ok(table)
            })
```

docs/source/metadata/metadata.md
```diff
@@ -0,0 +1,220 @@
...
+
+```javascript
+keyspace.getTable("users");
+keyspace.getMaterializedView("users_by_email");
+keyspace.getUdt("address");
+```
+
+Prefer these when you need a single object. Reading `keyspace.tables` converts and caches *every*
+table in the keyspace in order to hand you one of them, while `getTable()` converts only the one
+you asked for. Either way you get the same cached instance.
```
wprzytula: 🤔 I think this is too low-level, an implementation detail. Reader doesn't have to know:
1. What _conversion_ you mean - why does driver have to _convert_ tables? From what into what?
2. _the same cached instance_ - here perhaps it makes more sense, because at the beginning you said:
> the driver keeps a snapshot of the cluster's schema in memory

...so a reader may guess this snapshot is what is cached.

docs/source/metadata/metadata.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+A metadata object you are holding is therefore a snapshot: it will not change under you, but it
+also will not pick up schema changes. Re-read it from `client.metadata` when you need current
+information.
+
+:::{note}
+Refreshes happen in the background, so immediately after a DDL statement the driver may still
+report the previous schema for a short while.
+:::
```
wprzytula: 💭 OK, so here you do describe the semantics (equality of keyspaces, but inequality of keyspace containers).
I don't see why this should be the case. If we cache keyspace _wrappers_, why not cache _collections of keyspace wrappers_? That's just a second layer of caching, justified just as much as the first one IMO. Or am I wrong?

docs/source/migration-guide/migration-guide.md
```diff
@@ -238,6 +238,119 @@ raw `Buffer`.
...
+
+```javascript
+// cassandra-driver
+ks.strategy; // "org.apache.cassandra.locator.NetworkTopologyStrategy"
+ks.strategyOptions; // { dc1: "3" }
+
+// ScyllaDB Node.js RS Driver
+ks.strategy.kind; // StrategyKind.NetworkTopologyStrategy
+ks.strategy.datacenterRepfactors; // { dc1: 3 }
+```
```
wprzytula, line 297: ❓ Shouldn't this be `"dc1"` (in quotes)?

docs/source/migration-guide/migration-guide.md
```diff
@@ -238,6 +238,119 @@ raw `Buffer`.
...
+const pkTypes = table.partitionKeys.map((column) => column.type);
+
+// ScyllaDB Node.js RS Driver
+const pkTypes = table.partitionKey.map((name) => table.columns[name].type);
+```
+
+Table storage options - `caching`, `comment`, `compactionClass`, `compactionOptions`,
+`compression`, `defaultTtl`, `gcGraceSeconds`, `id`, `bloomFilterFalsePositive`,
+`speculativeRetry`, `extensions`, `indexes` and the rest - are not exposed. Only `partitioner`
+survives, and it is `null` unless the table overrides the cluster default. Query `system_schema`
```
wprzytula: > Only `partitioner` survives, and it is `null` unless the table overrides the cluster default.

🤔 Are you sure this is the case?

docs/source/migration-guide/migration-guide.md
```diff
@@ -238,6 +238,119 @@ raw `Buffer`.
...
+`MaterializedView` still extends the table metadata and still has `tableName`, but
+`whereClause` and `includeAllColumns` are not exposed.
+
+`Udt` keeps `name` and `fields` (an array of `{ name, type }`) and gains `keyspace`.
+
+#### Keyspaces no longer carry their own name
+
+`KeyspaceMetadata` has `strategy`, `durableWrites`, `tables`, `views` and `udts`, and nothing else.
+There is no `name` property - you either passed the name to `getKeyspace()` or read it as the key
+from `getKeyspaces()`. `virtual` and `graphEngine` are not exposed either.
```
wprzytula: 🤔 I don't see how `virtual` and `graphEngine` fit this section.

docs/source/migration-guide/migration-guide.md
```diff
@@ -238,6 +238,119 @@ raw `Buffer`.
...
+
+#### Metadata objects are cached, shared and read-only
+
+Repeated lookups of the same object return the same instance, and every schema change replaces
+them wholesale:
+
+```javascript
+client.metadata.getTable("ks", "tbl") === client.metadata.getKeyspace("ks").tables["tbl"]; // true
+```
```
wprzytula, line 349: ⚠️ I think this is false! `client.metadata` does not have a stable `ClusterState` underneath - it may be swapped for another by the Rust Driver's tokio worker at any time. So in particular, a switch can happen in the mid of this line's evaluation (between `metadata.getTable()` and `metadata.getKeyspace()` are called).

wprzytula (changes requested):
My confusion about the semantics of the new `Metadata` (_snapshot_ vs _newest available_) convinces me once again that we should start with a thorough design written in an issue, and update that design if during implementation some changes come up as needed. WDYT @hmarszalek?

docs/source/migration-guide/migration-guide.md
hmarszalek: `{ dc1: "3" }` is valid JS

src/metadata/state.rs
```diff
@@ -21,11 +51,543 @@ pub(crate) struct ClusterSnapshot {
...
+                        let reference =
+                            wrapper.into_reference(*env).map_err(ConvertedError::from)?;
+                        ConvertedResult::Ok((name.to_owned(), reference))
+                    })
+                    .collect::<ConvertedResult<HashMap<_, _>>>()
+            })?;
+            let record: NamedMap<String, Reference<KeyspaceWrapper>> = NamedMap::new(wrappers);
+            record.into_jsinstance(env).map_err(ConvertedError::from)
+        })
+    }
```
wprzytula, line 113: 📌 Please open an issue to delete this intermediate allocation of HashMap.

wprzytula (approved):

PR #514, opened 2026-08-09
Query trace

### LINKED ISSUE #523: Support for tracing info
Introduce `query-trace.ts` and implement `getTrace()` endpoint in Metadata class.

hmarszalek:
Executing a statement with `traceQuery: true` reports a trace id on `result.info.traceId`, and passing that id to `getTrace()` now returns the trace the server recorded – request type, coordinator, timing, bound parameters, and the per-step event timeline.

**Rust side.** `SessionWrapper.getTracingInfo()` fetches the `TracingInfo` from the underlying Rust driver and converts it into fully-formed JS `QueryTrace`/`TracingEvent` instances through the constructor registration. Since `JsInstance` values are JS-thread-only, the async method returns a plain Send carrier (`TracingInfoResult`) and the actual JS objects get built in its `ToNapiValue` impl, back on the JS thread.

**JS side.** `QueryTrace` and `TracingEvent` are converted to (written in) TypeScript and construct their own `TimeUuid`/`InetAddress` fields from the raw buffers Rust hands over. `getTrace()` is async, supporting both the promise and callback styles used elsewhere in the driver. The optional consistency argument is accepted for compatibility but not wired, since the Rust driver uses session-level tracing consistency instead.

Added integration tests covering a real `QueryTrace` and its populated events array. The tracing example was unbroken and rewritten against the real API. A new Query Tracing docs page was added to the toctree.

Fixes: #523 

## Pre-review checklist


eslint.config.mjs
```diff
@@ -12,6 +12,7 @@ export default [
   includeIgnoreFile(path.join(__dirname, "lib", ".gitignore")),
   includeIgnoreFile(path.join(__dirname, "lib", "metadata", ".gitignore")),
   includeIgnoreFile(path.join(__dirname, "lib", "types", ".gitignore")),
+  includeIgnoreFile(path.join(__dirname, "lib", "metadata", ".gitignore")),
   pluginJs.configs.recommended,
   {
     languageOptions: {
```
wprzytula, line 17: ❓ Shouldn't this have been done earlier, in an earlier commit or a separate pre-commit? This fixes a pre-existing bug AFAIU.

lib/metadata/index.js
```diff
@@ -27,16 +15,6 @@ const _selectSchemaVersionPeers = "SELECT schema_version FROM system.peers";
...
-/**
- * @const
- * @private
- */
-const _traceMaxAttemps = 5;
-/**
- * @const
- * @private
- */
-const _traceAttemptDelay = 400;
```
wprzytula: ♻️ Deletion of these is separate from implementing support for tracing info - it's merely deletion of old helpers that will no longer be needed.

lib/metadata/index.js
```diff
@@ -218,11 +201,24 @@ class Metadata {
...
+     * @returns {Promise<QueryTrace>}
+     */
+    getTrace(traceId, consistency, callback) {
+        if (!callback && typeof consistency === 'function') {
+          // Both callback and consistency are optional parameters
+          // In this case, the second parameter is the callback
+          callback = consistency;
+          // eslint-disable-next-line no-useless-assignment
+          consistency = null;
+        }
```
wprzytula: 😖 I can't say how much I hate this lack of respect towards parameter ordering that permeats the old DSx driver. I hope this won't block us from typescriptifying metadata.js.

src/tracing_info.rs
```diff
@@ -0,0 +1,101 @@
...
+use crate::utils::js_instance::JsInstance;
+
+/// Converts an `IpAddr` to its raw bytes (4 for IPv4, 16 for IPv6), for handing to the JS
+/// `InetAddress` constructor, which itself takes a raw address buffer.
+fn ip_addr_to_buffer(ip: IpAddr) -> Buffer {
+    match ip {
+        IpAddr::V4(v4) => Buffer::from(v4.octets().as_slice()),
+        IpAddr::V6(v6) => Buffer::from(v6.octets().as_slice()),
+    }
+}
```
wprzytula: 📌  I see you're still using `Buffer` for small allocations, such as Uuid or IpAddr. The problem is, this creates a napi external buffer, which has giant overhead. Can you remember that I suggested introducing a `CopyBuffer` or something like this, to cope with that? Let's have a separate PR that does that and base this one on top of that one.

docs/source/tracing/tracing.md
```diff
@@ -0,0 +1,42 @@
+# Query Tracing
+
+Tracing is a feature provided by Scylla. When sending a query we can set a flag that signifies that we would like it to be traced. After completing the query ScyllaDB provides a tracing_id which can be used to fetch information about it - which nodes it was sent to, what operations were performed etc.
```
wprzytula: ⛏️ Scylla -> ScyllaDB

wprzytula (changes requested):

lib/metadata/query-trace.ts
```diff
@@ -0,0 +1,88 @@
...
+ * @alias module:metadata~QueryTrace
+ */
+class QueryTrace {
+    readonly requestType: string | null;
+    readonly coordinator: InetAddress | null;
+    readonly parameters: Readonly<Record<string, string>>;
+    readonly startedAt: Date | null;
+    readonly duration: number | null;
+    readonly clientAddress: InetAddress | null;
+    readonly events: readonly TracingEvent[];
```
wprzytula, line 54: ❓ What's the difference between `Readonly<...>` and `readonly TracingEvent[]`? One is a type, another is a keyword.

lib/metadata/query-trace.ts
```diff
@@ -56,33 +60,44 @@ class QueryTrace {
...
+        this.coordinator = coordinator ? new InetAddress(coordinator) : null;
         this.parameters = parameters || {};
-        this.startedAt = startedAt;
+        this.startedAt =
+            typeof startedAt === "number" ? new Date(startedAt) : null;
         this.duration = duration;
-        this.clientAddress = clientAddress;
+        this.clientAddress = clientAddress
+            ? new InetAddress(clientAddress)
+            : null;
```
wprzytula, line 93: 📝 Passing addresses as `Buffer` is optimal, because `InetAddress` holds a `Buffer` inside, so this allocation is anyway needed.

examples/tracing/retrieve-query-trace.js
```diff
@@ -0,0 +1,59 @@
+"use strict";
+const cassandra = require("@scylladb/driver");
```
wprzytula, line 2: 📌 Let's rename all `cassandra` variables into `scylladb` as a separate PR.

docs/source/migration-guide/migration-guide.md
```diff
@@ -359,6 +359,39 @@ Because those instances are shared, they are read-only to full depth: fields can
...
+
+`getTrace()` still accepts an optional consistency level between the trace id and the callback, so
+existing call sites keep compiling and running:
+
+```javascript
+const trace = await client.metadata.getTrace(traceId, types.consistencies.all);
+```
+
+It is accepted for compatibility only and is not plumbed through to the request, so the trace is
+read at the driver's own consistency level regardless of what you pass.
```
wprzytula, line 393: 📌 In a follow-up, you should add a per-client setting of consistency level for tracing queries, and bridge it with Rust Driver's corresponding setting on SessionBuilder.

wprzytula (approved):

lib/metadata/query-trace.ts
hmarszalek: `Readonly<T>` is a mapped type that applies the `readonly` modifier to every property of `T`. `readonly T[]` is `ReadonlyArray<T>`'s type shorthand.

I could apply `ReadonlyArray<T>` for symmetry.

wprzytula: Understood. No need for the symmetry.

PR #515, opened 2026-08-11
Client Routes support

hmarszalek:
Under PrivateLink nodes addresses must be resolved through their connection ids. The rust driver already does all of this behind its `unstable-client-routes` feature (reads `system.client_routes`, subscribes to `CLIENT_ROUTES_CHANGE`, installs a Host ID based translator), so this PR just enables and exposes it:

- Enabled the `unstable-client-routes` feature on the `scylla` dependency.
- Added a `clientRoutes` client option taking the Scylla Cloud connection ids to read routes for, with an optional per-endpoint `hostnameOverride`.
- Introduced a `ConfiguredSessionBuilder` enum, containing both default `SessionBuilder` and `ClientRoutesSessionBuilder`, since Rust driver defines them separately.
- Rejected `clientRoutes` combined with `policies.addressResolution` or `sslOptions` explicitly rather than dropping them silently.

## Pre-review checklist

### DIFF
--- a/main.d.ts
+++ b/main.d.ts
@@ -227,8 +227,20 @@ export interface ExecutionOptions {
   setHints(hints: string[]): void;
 }
 
+/** A single client routes proxy the driver should read `system.client_routes` for. */
+export interface ClientRoutesProxy {
+  /** The ScyllaDB Cloud connection id used to filter `system.client_routes`. */
+  connectionId: string;
+  /**
+   * Overrides the hostname read from `system.client_routes` for this connection id.
+   * Useful for testing and for some cloud architectures.
+   */
+  hostnameOverride?: string;
+}
+
 export interface ClientOptions {
   contactPoints?: string[];
+  clientRoutes?: { proxies: ClientRoutesProxy[] };
   localDataCenter?: string;
   keyspace?: string;
   authProvider?: auth.AuthProvider;
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
   "scripts": {
     "build:ts": "tsc -p tsconfig.build.json",
     "build": "sh -c 'napi build --platform --release \"$@\" && npm run build:ts' --",
+    "build:test": "sh -c 'napi build --platform --features client-routes-proxy-tests \"$@\" && npm run build:ts' --",
     "all-tests": "npm run unit && npm run integration && npm run unit-not-supported",
     "unit": "./node_modules/.bin/mocha test/unit -t 5000 --recursive",
     "unit-gc": "./node_modules/.bin/mocha --node-option=expose-gc test/unit -t 5000 --recursive --grep 'expose-gc'",
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -12,9 +12,14 @@ use openssl::ssl::{
 use openssl::x509::X509;
 use openssl::x509::store::X509StoreBuilder;
 use scylla::client::SelfIdentity;
+use scylla::client::client_routes::{
+    ClientRoutesConfig as ScyllaClientRoutesConfig, ClientRoutesProxy,
+};
 use scylla::client::execution_profile::ExecutionProfileBuilder;
+use scylla::client::session::Session;
 use scylla::client::session_builder::{
-    GenericSessionBuilder, SessionBuilder, SessionBuilderKindSupportsKnownNodes,
+    ClientRoutesSessionBuilder, GenericSessionBuilder, SessionBuilder,
+    SessionBuilderKindSupportsKnownNodes,
 };
 use scylla::policies::host_filter::AllowListHostFilter;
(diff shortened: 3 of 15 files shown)


wprzytula:
:wrench: General naming comment: please do not use _PrivateLink support_ naming. Say: _Client Routes support_. This is because Client Routes is an implemented mechanism in ScyllaDB + drivers to support many kinds of proxied setups, with AWS PrivateLink and GCP PSC among others.

src/session/config.rs
```diff
@@ -349,6 +393,44 @@ pub(crate) fn configure_session_builder(
     Ok(builder)
 }
 
+/// Configures a builder for a session that reaches nodes through PrivateLink
+/// endpoints, resolving their addresses by Host ID.
+fn configure_client_routes_builder(
+    client_routes_config: ClientRoutesConfig,
+    options: &SessionOptions,
+) -> ConvertedResult<ClientRoutesSessionBuilder> {
```
wprzytula, line 411: ❓ Why does one function take `&SessionOptions` and the other take `SessionOptions`?

src/session/config.rs
```diff
@@ -414,21 +423,43 @@ fn configure_client_routes_builder(
...
+        .enumerate()
+        .map(|(index, endpoint)| {
+            let connection_id = endpoint
+                .connection_id
+                .filter(|id| !id.is_empty())
+                .ok_or_else(|| {
+                    ConvertedError::from(make_js_error(format!(
+                        "clientRoutes endpoint at index {index} must have a non-empty connectionId"
+                    )))
+                })?;
```
wprzytula, line 448: ♻️ I don't understand why commit that introduces hostname overrides changes empty connection id handling.

main.d.ts
```diff
@@ -227,9 +227,20 @@ export interface ExecutionOptions {
...
+/** A single PrivateLink endpoint the driver should read `system.client_routes` for. */
+export interface ClientRoutesEndpoint {
+  /** The ScyllaDB Cloud connection id used to filter `system.client_routes`. */
+  connectionId: string;
+  /**
+   * Overrides the hostname read from `system.client_routes` for this connection id.
+   * Useful for testing and for some cloud architectures.
+   */
+  hostnameOverride?: string;
+}
```
wprzytula: 🔧 _Endpoint_ naming was purposely dropped in Rust Driver because it was confusing. Rust Driver has `ClientRoutesProxy` instead - let's call this struct the same.

package.json
```diff
@@ -52,6 +52,7 @@
   "scripts": {
     "build:ts": "tsc -p tsconfig.build.json",
     "build": "sh -c 'napi build --platform --release \"$@\" && npm run build:ts' --",
+    "build:test": "sh -c 'napi build --platform --features client-routes-proxy-tests \"$@\" && npm run build:ts' --",
```
wprzytula, line 55: 📌 The native add-on exposes many more test-only API endpoints. See #66. In a follow-up, you can gate all such test-only code behind one feature flag (e.g., `"testing"`).

test/integration/supported/client-routes-proxy-tests.js
```diff
@@ -0,0 +1,247 @@
...
+const rust = require("../../../index");
+
+const NODE_COUNT = 3;
+const CONNECTION_ID = "nodejs-driver-client-routes-test";
+const REST_API_PORT = 10000;
+// Returns the coordinator's host id, which is how this test identifies which node
+// actually served a query.
+const WHOAMI_QUERY = "SELECT host_id FROM system.local WHERE key='local'";
+// An address nothing listens on, used to make a route unusable on purpose.
+const DEAD_ADDRESS = "127.0.0.254";
```
wprzytula: ❓ How can you know that no one listens on this address?

test/integration/supported/client-routes-proxy-tests.js
```diff
@@ -0,0 +1,247 @@
...
+
+describe("Client routes (PrivateLink)", function () {
+    this.timeout(180000);
```
wprzytula: Oh wow, what a long timeout - 3 minutes? Do we need such long one?

wprzytula (changes requested):

src/session/config.rs
hmarszalek: The default path needs to move `address_translator_config` out. This isn't necessary for `ClientRoutesConfig`.

test/integration/supported/client-routes-proxy-tests.js
hmarszalek: It can be decreased to 1 minute (still generous, but should be enough)

wprzytula: Please do that.

hmarszalek: Already done.

wprzytula (approved):

Issue #380, opened 2026-01-13
Ensure TypeHints validity for custom types


adespawn:
Reading the code logic, the text representation of the type hints (see [parsing code](https://github.com/scylladb/nodejs-rs-driver/blob/88380ecf14b235c6919685ee838ab5976e3be6f3/lib/types/index.js#L126-L168)) does not handle properly custom types (mostly vector).

PR #520, opened 2026-08-19
Add vector search example

hmarszalek:
Vector search needs nothing new from this driver: the CQL vector type is already supported, and `ORDER BY ... ANN OF ?` is plain CQL sent through the normal `execute()` path. This PR proves it with a minimal working example.

`examples/vector/vector-search-ann.js` stores comment embeddings in a `vector<float, 3>` column, creates a vector index, and runs an ANN similarity query.
`examples/vector/vector-insert-select.js` covers plain vector column insert/select.

Enabled tests for map, set, ascii, and varint variants inside a vector.

## Pre-review checklist

- ~[ ] I added relevant tests for new features and bug fixes.~
- ~[ ] I have made sure that the type stubs / TS definitions match the JS public API.~

### DIFF
--- a/examples/vector/vector-insert-select.js
+++ b/examples/vector/vector-insert-select.js
@@ -1,0 +1,70 @@
+"use strict";
+const cassandra = require("@scylladb/driver");
+const { getClientArgs } = require("../util");
+
+const client = new cassandra.Client(getClientArgs());
+
+/**
+ * Creates a table with a vector column, inserts a row and selects a row.
+ */
+async function example() {
+    await client.connect();
+
+    await client.execute(
+        "CREATE KEYSPACE IF NOT EXISTS examples WITH replication =" +
+            "{'class': 'NetworkTopologyStrategy', 'replication_factor': '1' }",
+    );
+
+    await client.execute(
+        "CREATE TABLE IF NOT EXISTS examples.vector_comments " +
+            "(id uuid PRIMARY KEY, comment text, comment_vector vector<float, 5>)",
+    );
+
+    // A Vector can be built from a plain Array (with an explicit subtype)...
+    const commentVector = new cassandra.types.Vector(
+        [0.12, 0.34, 0.56, 0.78, 0.91],
+        "float",
+    );
+    const id1 = cassandra.types.Uuid.random();
+
+    console.log(`Inserting comment for id ${id1} from plain Array`);
+    await client.execute(
+        "INSERT INTO examples.vector_comments (id, comment, comment_vector) VALUES (?, ?, ?)",
+        [id1, "Great product, fast shipping!", commentVector],
+        { prepare: true },
+    );
+
+    const result = await client.execute(
+        "SELECT comment, comment_vector FROM examples.vector_comments WHERE id = ?",
+        [id1],
+        { prepare: true },
+    );
+    const row = result.first();
+    console.log(`Retrieved comment for id ${id1}: "${row["comment"]}"`);
+    console.log(`Retrieved vector for id ${id1}: ${row["comment_vector"]}`);
+
+    // ...or from a Float32Array.
+    const commentVectorArray = new Float32Array([0.11, 0.22, 0.33, 0.44, 0.55]);
+    const id2 = cassandra.types.Uuid.random();
+
+    console.log(`Inserting comment for id ${id2} from Float32Array`);
+    await client.execute(
+        "INSERT INTO examples.vector_comments (id, comment, comment_vector) VALUES (?, ?, ?)",
+        [id2, "Exactly as described", commentVectorArray],
+        { prepare: true },
+    );
+
+    const result2 = await client.execute(
+        "SELECT comment, comment_vector FROM examples.vector_comments WHERE id = ?",
+        [id2],
+        { prepare: true },
+    );
+    const row2 = result2.first();
+    console.log(`Retrieved comment for id ${id2}: "${row2["comment"]}"`);
+    console.log(`Retrieved vector for id ${id2}: ${row2["comment_vector"]}`);
+}
+
+example().catch(function (err) {
+    console.error("There was an error", err);
+    process.exitCode = 1;
+});
--- a/examples/vector/vector-search-ann.js
+++ b/examples/vector/vector-search-ann.js
@@ -1,0 +1,102 @@
+"use strict";
+const cassandra = require("@scylladb/driver");
+const { getClientArgs } = require("../util");
+
+const client = new cassandra.Client(getClientArgs());
+
+const comments = [
+    { text: "Great product, fast shipping!", vector: [0.9, 0.1, 0.0] },
+    { text: "Arrived quickly, very happy", vector: [0.8, 0.2, 0.0] },
+    { text: "The colour is completely wrong", vector: [0.1, 0.9, 0.0] },
+    { text: "Broke after a single use", vector: [0.0, 0.2, 0.9] },
+];
+
+// The embedding we search with. Closest in meaning to the "fast shipping"
+// comments, so those are the ones the ANN query should return first.
+const queryVector = [0.95, 0.05, 0.0];
+
+/**
+ * Stores comment embeddings in a `vector` column and finds the ones most
+ * similar to a query embedding with an `ORDER BY ... ANN OF` query.
+ */
+async function example() {
+    await client.connect();
+
+    await client.execute(
+        "CREATE KEYSPACE IF NOT EXISTS examples WITH replication =" +
+            "{'class': 'NetworkTopologyStrategy', 'replication_factor': '1' }",
+    );
+
+    await client.execute(
+        "CREATE TABLE IF NOT EXISTS examples.vector_search_comments " +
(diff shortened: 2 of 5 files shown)


examples/vector/vector-search-ann.js
```diff
@@ -0,0 +1,106 @@
...
+                prepare: true,
+            });
+            if (result.rows.length > 0) {
+                return result.rows;
+            }
+        } catch (_) {
+            // Vector search not yet ready; wait and retry.
+        }
+        if (Date.now() > deadline) {
+            return [];
```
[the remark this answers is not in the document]
hmarszalek: I am not sure how to filter those errors in an elegant way (not string parsing).

examples/vector/vector-insert-select.js
```diff
@@ -0,0 +1,49 @@
...
+    await client.connect();
+
+    await client.execute(
+        "CREATE KEYSPACE IF NOT EXISTS examples WITH replication =" +
+            "{'class': 'NetworkTopologyStrategy', 'replication_factor': '1' }",
+    );
+
+    await client.execute(
+        "CREATE TABLE IF NOT EXISTS examples.vector_comments " +
+            "(id uuid PRIMARY KEY, comment text, comment_vector vector<float, 5>)",
```
QuerthDP: PR cover letter says `vector<float, 3>`

examples/vector/vector-insert-select.js
```diff
@@ -0,0 +1,49 @@
...
+    );
+
+    console.log("Inserting");
+    const id = cassandra.types.Uuid.random();
+    // A Vector can be built from a plain Array (with an explicit subtype) or a Float32Array.
```
QuerthDP: Could you provide both version in example?

examples/vector/vector-search-ann.js
```diff
@@ -0,0 +1,106 @@
...
+        }
+        await new Promise((resolve) => setTimeout(resolve, 250));
```
QuerthDP, line 100: Is it just a sleep in JS?

QuerthDP (approved):
Looks fine to me.
Consider adding both vector representations in working with vector example.

Consider checking the index availability via an  `/api/v1/indexes/{keyspace}/{index}/status` Vector Store endpoint instead of retrying the SELECT.

examples/vector/vector-search-ann.js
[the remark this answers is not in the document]
hmarszalek: Good catch — the retry was swallowing genuine schema/auth/connection failures.

On filtering without string parsing: there is no way to do it today. These all arrive as a napi `GenericFailure` (`src/errors.rs:178`) with a flat message, so a "still building the index" failure and a real CQL error are indistinguishable to the client — there is no structured server error code to switch on.

So instead of filtering, nothing is discarded: the last error is kept and attached as the `cause` of the timeout error.

```
There was an error Error: Vector search did not become available - is the Vector Store service running?
  [cause]: [Error [ExecutionError]: Database returned an error: ... Vector Store service is unavailable] {
    code: 'GenericFailure'
  }
```

examples/vector/vector-insert-select.js
hmarszalek: Done — the example now inserts and reads back both forms: a `Vector` built from a plain `Array` with an explicit `"float"` subtype, and a `Float32Array`.

```
Inserting
Retrieved comment: "Great product, fast shipping!"
Retrieved vector: [0.11999999731779099,0.3400000035762787,...]
Retrieved comment: "Exactly as described"
Retrieved vector: [0.10999999940395355,0.2199999988079071,...]
```

examples/vector/vector-search-ann.js
hmarszalek: Yes — `new Promise((resolve) => setTimeout(resolve, ms))` is the usual way to sleep in JS, since there is no built-in blocking sleep.

Node ships a promise-based version though, so I switched to it:

```js
const { setTimeout: sleep } = require("node:timers/promises");
...
await sleep(250);
```

wprzytula:
> `examples/vector/vector-search-ann.js` stores comment embeddings in a `vector<float, 3>` column, creates a vector index, and runs an ANN similarity query. To make this runnable in CI, `.github/docker-compose.yml` now also starts the Vector Store indexing service and points the nodes at it with `--vector-store-primary-uri`.

What's the impact of this on CI's execution time?

wprzytula:
Isn't this a blocker for considering vector type properly supported?
https://github.com/scylladb/nodejs-rs-driver/issues/380

wprzytula:
Rust Driver includes a full integration test for vector type with all possible cql types inside: https://github.com/scylladb/scylla-rust-driver/blob/2c8539fc3c65b018d98630b4e8a137208070c303/scylla/tests/integration/types/cql_collections.rs#L608
Can you port it to Node.js RS Driver too? We need to verify all corner cases, because the vector type has weird and confusing de/ser, which might be broken in the JS code inherited from the old DSx driver.

wprzytula (changes requested):
What you added looks good. It might be not enough though - see my other comments.

Akvear:
> Rust Driver includes a full integration test for vector type with all possible cql types inside

Both C# drivers also test all types nested in a vector:
https://github.com/scylladb/csharp-driver/blob/36ded54dedbe6e3a023ca0178d91368f12923b85/src/Cassandra.IntegrationTests/Core/TypeSerializersTests.cs#L313

hmarszalek:
> What's the impact of this on CI's execution time? 

The time needed for the examples job to run went from 1:40 to around 3:40.

wprzytula:
> > What's the impact of this on CI's execution time?
> 
> The time needed for the examples job to run went from 1:40 to around 3:40.

I don't find such big increase acceptable for just one example, not strictly related to driver's functionality (just ScyllaDB itself with ANN queries).
Let's either not have this example (worse) or disable running this example in the CI (better).

hmarszalek:
> Isn't this a blocker for considering vector type properly supported? #380

Vectors with primitive types and collections, and vector nested inside list/set pass just fine. What breaks is any parameterized type inside map or tuple.

wprzytula:
> > Isn't this a blocker for considering vector type properly supported? #380
> 
> Vectors with primitive types and collections, and vector nested inside list/set pass just fine. What breaks is any parameterized type inside map or tuple.

We need to fix this as part of our effort to support vector type.

hmarszalek:
> > > Isn't this a blocker for considering vector type properly supported? #380
> > 
> > 
> > Vectors with primitive types and collections, and vector nested inside list/set pass just fine. What breaks is any parameterized type inside map or tuple.
> 
> We need to fix this as part of our effort to support vector type.

Addressed in #521.

test/test-helper.js
```diff
@@ -1872,8 +1870,8 @@ const dataProviderWithCollections = dataProvider
                 ],
                 customTypeName: "vector",
             },
-            value: data.value.map((value) => [value, value, value]),
-        }, */
+            value: data.value.map((value) => [value]),
```
wprzytula, line 1895: commit `Enable vector<map> and vector<set> in dataProviderWithCollections`:
❓What's _the provider_? Why was it put like this in the code (duplicated thrice) - why did _the provider_ build it this way?

> The provider built each set as [value, value, value], but sets drop
> duplicates — so the server returned one element while the test expected
> three. I changed it to [value]: one element per set, nothing to
> deduplicate.

test/test-helper.js
```diff
@@ -1637,6 +1637,28 @@ const dataProvider = [
...
+            code: types.dataTypes.custom,
+            customTypeName: "vector",
+            info: [{ code: types.dataTypes.varint }, 3],
+        },
+        value: [
+            types.Integer.fromString("-1"),
+            types.Integer.fromString("0"),
+            types.Integer.fromString("12345678901234567890"),
+        ],
+    },
```
wprzytula, line 1661: ❓ Any idea why these two were missing from _primitives_?
❓ Are now all primitives included? What about decimal?

hmarszalek: > ❓ Any idea why these two were missing from primitives?

Probably an oversight :woman_shrugging: .

> ❓ Are now all primitives included? What about decimal?

The following types ARE present (I think that's all of them):
"ascii", "bigint", "blob", "boolean", "counter", "date", "decimal", "double", "duration", "float", "inet", "int", "smallint", "text", "time", "timestamp", "timeuuid", "tinyint", "uuid", "varchar", "varint".

wprzytula: I confirm these are all of them.

test/test-helper.js
hmarszalek: > What's the provider?

`dataProvider` (in test-helper.js) lists one entry per CQL primitive: its name, the type info for vector<subtype, 3>, and three sample values. `dataProviderWithCollections` expands each of those into extra entries – list<T>, set<T>, map<int, T>, tuple<T, T>, vector<T, 3> and my_udt.

> Why was it put like this in the code (duplicated thrice) - why did the provider build it this way?

Dunno, but it was incorrect. The bug wasn't caught earlier because the tests weren't enabled.

wprzytula (approved):

wprzytula (approved):

PR #521, opened 2026-08-20
Fix type hints parsing

### LINKED ISSUE #380: Ensure TypeHints validity for custom types

hmarszalek:
Text type hints are resolved by `dataTypes.getByName()`, which split the arguments of a parameterized type without regard for nesting. A parameterized type used as an argument was therefore torn apart, and hints like these were rejected as invalid:
- `map<int, vector<float, 3>>`
- `tuple<vector<float, 3>, int>`
- `map<int, map<int, text>>`
- `tuple<map<int, text>, int>`
- `map<int, tuple<int, text>>`

The arguments are now split by a shared helper that tracks angle bracket depth and only breaks on the commas separating the arguments themselves. Vector finds its dimension the same way, and validates its own argument count rather than relying on the regular expression to do it.

`dataTypes.getByName()` had no direct test coverage at all. Every type hint exercised elsewhere was either a scalar or a parameterized type whose arguments were scalars, so the failing shape was never parsed in a test. The second commit covers every scalar type name, the partial hints, case and whitespace handling, each container holding a parameterized argument, and the names that must be rejected.

Fixes: #380

## Pre-review checklist

- ~[ ] I have made sure that the type stubs / TS definitions match the JS public API.~

### DIFF
--- a/lib/types/index.js
+++ b/lib/types/index.js
@@ -90,6 +90,33 @@ consistencyToString[consistencies.localOne] = "LOCAL_ONE";
  * @property {Number} udt User-defined type.
  * @property {Number} tuple A sequence of values.
  */
+
+/**
+ * Splits the comma separated arguments of a parameterized type name, ignoring
+ * the commas that belong to the arguments themselves.
+ *
+ * @param {String} args Everything between the outermost angle brackets.
+ * @returns {Array<String>} The arguments, trimmed.
+ */
+function splitTypeArgs(args) {
+    const result = [];
+    let depth = 0;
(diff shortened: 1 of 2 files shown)


wprzytula (approved):
🚀

PR #525, opened 2026-08-27
Run unit and integration tests on ARM64

### LINKED ISSUE #524: Run tests for arm64

dkropachev:
## Why

ARM64 release artifacts need native unit and integration coverage. Repeating the full Node.js matrix on both architectures would add cost without changing the executed suites.

## What

- Add native Ubuntu ARM64 unit and integration jobs on Node.js 20.
- Configure CCM to download the matching ARM64 ScyllaDB package.
- Keep the full Node.js version matrix on x86_64.
- Update maintenance and release documentation.

Fixes #524

### DIFF
--- a/.github/common/integration/action.yml
+++ b/.github/common/integration/action.yml
@@ -4,7 +4,9 @@ description: "Runs integration tests on a single architecture and node version"
 inputs:
   build-command:
     required: true
-    type: string
+  scylla-arch:
+    description: "Architecture of the ScyllaDB package installed by CCM"
+    required: true
 
 runs:
   using: "composite"
@@ -29,11 +31,13 @@ runs:
       env:
         CCM_IS_SCYLLA: "true"
         CCM_PATH: "./scylla-ccm"
+        SCYLLA_ARCH: ${{ inputs.scylla-arch }}
       shell: bash
       run: npm run integration
     - name: Run integration tests on scylla with expose-gc
       env:
         CCM_IS_SCYLLA: "true"
         CCM_PATH: "./scylla-ccm"
+        SCYLLA_ARCH: ${{ inputs.scylla-arch }}
       shell: bash
       run: npm run integration-gc
--- a/.github/workflows/extended-ci.yml
+++ b/.github/workflows/extended-ci.yml
@@ -46,21 +46,35 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        settings:
+        host:
+          - ubuntu-latest
+          - ubuntu-24.04-arm
+        node: [ 20, 22, 24, current ]
+        include:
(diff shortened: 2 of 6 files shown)


.github/common/integration/action.yml
```diff
@@ -5,6 +5,10 @@ inputs:
   build-command:
     required: true
     type: string
+  scylla-arch:
+    description: "Architecture of the ScyllaDB package installed by CCM"
+    required: true
+    type: string
```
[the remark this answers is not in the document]
dkropachev: Fixed. Removed unsupported `type` metadata from both composite-action inputs; inputs remain strings.

.github/workflows/extended-ci.yml
wprzytula, line 1: 📝 You've fixed a pre-existing bug introduced in #515: changed
`npm run build ...` into `npm run build:test`. This merits a mention in the commit message, or rather a separate commit. cc @hmarszalek

.github/workflows/unit-tests.yml
```diff
@@ -19,16 +19,22 @@ jobs:
...
-          host: ubuntu-latest
-          target: x86_64-unknown-linux-gnu
+          host: ${{ matrix.settings.host }}
+          target: ${{ matrix.settings.target }}
       - name: Build
-        run: npm run build -- --target x86_64-unknown-linux-gnu
+        run: npm run build -- --target ${{ matrix.settings.target }}
         shell: bash
       - name: Run unit tests
         run: npm run unit
```
wprzytula, line 40: 📌 @hmarszalek as part of #66, let's remember to run `npm run build:test` here too.

.github/workflows/unit-tests.yml
wprzytula, line 45: ❓ @hmarszalek What's this?

hmarszalek: These are unit tests for classes present in `lib/` as plain JS, but not yet exposed to Rust backend.

wprzytula: Why do we run them in the CI? Do they actually test anything meaningful? Why are they called "not supported"?

hmarszalek: > Why do we run them in the CI?

Technically, the code they test isn't dead – it is still reachable even though it is not fully functional. Users can currently import and call them. 

> Why are they called "not supported"?

I assume the label is about driver feature completeness (unit broken – not exposed to package.json at all, unit not supported – tests that pass, but haven't yet been integrated with Rust core yet, unit – fully funtioncal).

wprzytula (approved):
@hmarszalek Please merge once you approve.

hmarszalek (approved):

PR #527, opened 2026-08-28
Cancel stale CI workflow runs

dkropachev:
## Why

Older CI runs consume runner capacity after newer commits make their results obsolete.

## What

- Group each workflow by pull request or pushed ref.
- Cancel an in-progress run when a newer run supersedes it.
- Apply the policy to docs, code quality, examples, integration, TypeScript, and unit workflows.

### DIFF
--- a/.github/workflows/check-docs.yml
+++ b/.github/workflows/check-docs.yml
@@ -12,6 +12,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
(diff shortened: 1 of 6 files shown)


wprzytula (approved):
@Lorak-mmk, WDYT about enabling this in other drivers too?

hmarszalek (approved):

Issue #244, opened 2025-10-02
CCM timeouts


adespawn:
According to [CI logs](https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/actions/runs/18190788172/job/51785163370#step:10:32), the ccm fails with timeout on the `ccm start --wait-for-binary-proto` command. 

## Problematic list of commands:
```sh
ccm create arbitraryname -v 3.11.4
ccm populate -n 1:0
ccm start --wait-for-binary-proto
```

## Steps to reproduce:
1. Modify the integration test task, by removing build and run integration tests step.
2. Add a script with offending commands as a CI step (see [this file](https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/64c885903fe500957e671467e963b932068523cc/.github/workflows/integration-tests.yml#L60-L61))
2. Alternatively you can add [ssh step](https://github.com/scylladb-zpp-2024-javascript-driver/scylladb-javascript-driver/blob/e24bbbe345d202138c601eca7fd3f95b0d7182d2/.github/workflows/integration-tests.yml#L62-L67) to the CI instead of the script, and run the offending commands manually after connecting with ssh.   
3. Run the CI

## Problem:
With some probability (around 20% observed before summer, around 80% now), the last command (`ccm start --wait-for-binary-proto` will run indefinitely. This happens only on the first launch of this command on a CI VM - meaning that once this command finishes successfully, you need to obtain a new VM to observe this issue.

### Taken debugging steps  
The ccm does not provide any logging.
When run with strace, I was able to determine, that a Cassandra java process is started, but I was unable to find any other useful information in that log

[strace-logs.zip](https://github.com/user-attachments/files/22677932/strace-logs.zip)

## ~Proposed~ Implemented workaround
With enabling of the integration tests on ScyllaDB we can for now skip testing on a Cassandra.

wprzytula:
@adespawn What's the status of this?

PR #529, opened 2026-08-28
Token and token ranges for Metadata and Host

### LINKED ISSUE #519: Support for Token and TokenRange
Implement the following Metadata class methods (#261):
- [x] getTokenRanges
- ~[ ] getTokenRangesForHost~
- [x] newToken
- [x] newTokenRange

Remove the `tokens` field from `Host`.

hmarszalek:
Implements the token-related endpoints of the `Metadata` class (`#getTokenRanges`, `#newToken`, `#newTokenRange`).

- `lib/token.ts` is rewritten on top of bigint token values instead of the old Long representation. Only the main `Token` class was exported to the user. The `RadomToken` and `ByteOrderedToken` subclasses are dead code and cannot be brought back to life on top of the Rust driver, because it does not support the partitioners they stand for. `MurMur3Token` doesn’t need to be represented as a separate subclass.
- `Token`'s constructor is registered with Rust.
- `lib/metadata/index.ts`: `newToken()` computes and constructs a JS class Token through the registered `Token` constructor. `newTokenRange()` constructs a `TokenRange` class from given tokens. `getTokenRanges()` builds the ring's ranges from the tokens Rust reports.

Testing

Unit tests cover the token/range arithmetic in isolation (splitting, wrapping, containment). Integration tests run against real clusters and check:
- `newToken` agrees with what the server itself reports via `SELECT token(...)`, for both single-column and composite partition keys
- the ring's ranges are contiguous and cover the whole ring.

Fixes: #519 

## Pre-review checklist


lib/token.ts
```diff
@@ -3,17 +3,22 @@
...
+ */
+const _Murmur3TokenType = { code: types.dataTypes.bigint, info: null };
 
 /**
  * Represents a token on the Cassandra ring.
  */
 class Token {
+    // What a token's value is depends on the partitioner that computed it, so the base class
+    // leaves it open and every subclass narrows it.
     #value: any;
```
wprzytula: ♻️  This explanatory comment has nothing to do with either of:
- _Rewrite Murmur3Token values as bigint_
- _correct TokenType_

lib/token.ts
```diff
@@ -61,16 +66,28 @@ class Token {
...
         return _Murmur3TokenType;
     }
+
+    compare(other: Token): number {
+        const otherValue = other.getValue();
+        if (this.getValue() < otherValue) {
+            return -1;
+        }
+        return this.getValue() > otherValue ? 1 : 0;
+    }
```
wprzytula: ♻️ The commit message doesn't mention at all nor justify addition of these methods.

src/metadata/token.rs
```diff
@@ -0,0 +1,117 @@
...
+/// Computes the Murmur3 token of a partition key, given the already-serialized values of its columns.
+/// Murmur3 is the only partitioner the Rust driver computes tokens with, so the partitioner of the
+/// cluster does not need to be consulted here.
+#[napi(ts_return_type = "import('./lib/token').Murmur3Token")]
+pub fn compute_murmur3_token<'env>(
+    env: &'env Env,
+    components: Vec<Buffer>,
+) -> JsResult<JsInstance<'env, js_constructible_class::Murmur3Token>> {
+    with_custom_error_sync(|| {
+        let mut hasher = Murmur3Partitioner.build_hasher();
```
wprzytula: ⚠️ ❗  False! There's also `CDCPartitioner`.

src/metadata/token.rs
```diff
@@ -0,0 +1,117 @@
...
+                            "Partition key component is too long: {} bytes",
+                            component.len()
+                        )))
+                    })?;
+                    hasher.write(&length.to_be_bytes());
+                    hasher.write(component);
+                    hasher.write(&[0u8]);
+                }
+            }
+        }
```
wprzytula: ❓ What is this logic taken from? Add comments.

src/metadata/token.rs
```diff
@@ -0,0 +1,117 @@
...
+                    .map(|(token, _)| {
+                        build_murmur3_token(env, FnArgs::from((BigInt::from(token.value()),)))
+                            .map_err(ConvertedError::from)
+                    })
+                    .collect::<ConvertedResult<Vec<_>>>()?;
+                ConvertedResult::Ok(Some(tokens))
+            })
+        })
+    }
+}
```
wprzytula: 💭 This has a large cost: it goes over the whole token ring and, for each token range upper bound, computes all replicas for that token and strategy.
❓ Why does user even need such endpoint? Rust Driver has been doing perfectly without it.

lib/tokenizer.ts
```diff
@@ -0,0 +1,137 @@
...
+     */
+    abstract hash(components: Buffer | Buffer[]): Token;
+
+    /**
+     * Parses the representation the server reports a token with.
+     *
+     * @param value The representation of the token.
+     * @returns The parsed token.
+     */
+    abstract parse(value: string): Token;
```
wprzytula: 🤔 I don't see why we need this in the wrapper.

lib/tokenizer.ts
```diff
@@ -0,0 +1,137 @@
...
+
+import { Murmur3Token, Token, TokenRange } from "./token";
+import { computeMurmur3Token } from "../index";
+
+/**
+ * Represents a set of methods that are able to generate and parse tokens for the partitioner of
+ * the cluster, and to reason about the ring those tokens are placed on.
+ * @ignore
+ */
+abstract class Tokenizer {
```
wprzytula: `Reintroduce the tokenizer` — when and why was the tokenizer removed?

lib/tokenizer.ts
```diff
@@ -0,0 +1,137 @@
...
+     * Splits the range ]start, end] into `numberOfSplits` parts of equal "size" (referring to the
+     * number of tokens, not the actual amount of data), returning the tokens the parts are split
+     * at.
+     *
+     * @param start Starting token.
+     * @param end End token.
+     * @param numberOfSplits Number of splits to make.
+     * @returns The `numberOfSplits - 1` points the range is split at.
+     */
+    abstract split(start: Token, end: Token, numberOfSplits: number): Token[];
```
wprzytula: ❓ What is this useful for? I really don't know such API and rationale for it.

wprzytula (commented):
After reading this PR confused me - in amount of things that were bridged and surprised me - I believe we should have a step back and start with describing the requirements, desired API and proposed implementation in an issue, and continue there.

lib/tokenizer.ts
hmarszalek: DSx allowed passing a string representation of a decimal value and creating a Token with that value.

hmarszalek:
Partially implemented – `Token`/`TokenRange` and `newToken`/`newTokenRange` are implemented; `getTokenRanges()` is not yet.

lib/token.ts
wprzytula, line 1: commit `Remove leftover tokenizer from token internals`:
🔧 It actually adds more new code than it removes old code. The name is misleading.

Cargo.toml
```diff
@@ -11,6 +11,7 @@ crate-type = ["cdylib"]
 napi = { version = "3.6.0", default-features = false, features = ["napi4", "napi6", "async"] }
 napi-derive = "3.4.0"
 scylla = { version = "1.8.0", features = ["num-bigint-03", "openssl-010", "unstable-nodejs-rs", "unstable-client-routes"] }
+scylla-cql-core = "1.8.0"
```
wprzytula: ❓ What for? Commit message doesn't provide any rationale.

Cargo.toml
```diff
@@ -10,7 +10,7 @@ crate-type = ["cdylib"]
 
 napi = { version = "3.6.0", default-features = false, features = ["napi4", "napi6", "async"] }
 napi-derive = "3.4.0"
-scylla = { version = "1.8.0", features = ["num-bigint-03", "openssl-010", "unstable-nodejs-rs", "unstable-client-routes"] }
+scylla = { version = "1.8.0", features = ["num-bigint-03", "openssl-010", "unstable-nodejs-rs", "unstable-client-routes", "unstable-csharp-rs"] }
```
wprzytula: > TODO: remove once compute_token_preserialized is marked with unstable-nodejs-rs.

Please open a relevant PR to Rust Driver immediately, not to forget.

lib/metadata/index.ts
```diff
@@ -126,13 +126,26 @@ class Metadata {
...
 
     /**
-     * Constructs a Token from the input buffer(s) or string input. If a string is passed in
-     * it is assumed this matches the token representation reported by cassandra.
-     * @param {Buffer[] | Buffer | string} components The token components.
+     * Constructs a Token from the input buffer(s)
+     *
+     * The token of a partition key is computed by the Rust driver, which reads the partitioner
+     * from the table's own metadata, so the components have to be the values of that table's
+     * partition key columns, in order, each already serialized with the CQL type of its column.
```
wprzytula: ❓ In what order? Please specify.

wprzytula: 🔧 Also provide an example for clarity. Given table definition X, you must provide the following components in order: Y, X.

wprzytula (changes requested):

wprzytula (commented):
An example is missing for computing tokens and using token ranges, right?
You could write one for executing range queries, assuming vnodes are used.

Otherwise, LGTM.

lib/metadata/index.ts
```diff
@@ -80,58 +81,86 @@ class Metadata {
...
+     * @returns {Set<TokenRange> | null} Ranges for the keyspace on this host or null if keyspace
+     * isn't found or hasn't been loaded.
      */
     getTokenRangesForHost(
         keyspaceName: string,
         host: Host,
-    ): Set<token.TokenRange> | null {
+    ): Set<TokenRange> | null {
         throw new Error("TODO: Not implemented");
     }
```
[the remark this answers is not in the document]
wprzytula: This is intended AFAIK, @hmarszalek?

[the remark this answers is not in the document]
hmarszalek: Yes.

examples/metadata/metadata-tokens.js
```diff
@@ -0,0 +1,74 @@
...
+const keyspace = "examples_tokens";
+const table = "readings";
+
+async function example() {
+    await client.connect();
+
+    await client.execute(
+        `CREATE KEYSPACE IF NOT EXISTS ${keyspace} WITH replication = {'class': 'NetworkTopologyStrategy',
+            'replication_factor': 1} AND tablets = {'enabled': false}`,
+    );
```
wprzytula, line 44: 🔧 The clause `AND tablets = {'enabled': false}` makes the example fail on Cassandra and older ScyllaDBs that do not support tablets!

⚠️  This hasn't been caught by the CI, because we still don't run it against Cassandra.

📌 Please add testing against Cassandra to the CI: both ITs and examples.

wprzytula: The proper solution should perform ifology based on tablets being supported by the cluster or not. See https://github.com/scylladb/scylla-rust-driver/blob/2bd506f18df74bbcb7a1afbc2b91574985e7dfd5/scylla/src/utils/test_utils.rs#L100 and https://github.com/scylladb/scylla-rust-driver/blob/2bd506f18df74bbcb7a1afbc2b91574985e7dfd5/scylla/src/utils/test_utils.rs#L44.

examples/metadata/metadata-tokens.js
```diff
@@ -0,0 +1,74 @@
...
+        keyspace,
+        table,
+    );
+    console.log("'%s' hashes to token %s", users[0], token.toString());
+
+    // With vnodes, the ring is divided once for the whole cluster: every keyspace and table sees
+    // the same ranges, and only who replicates each range depends on the keyspace's replication
+    // strategy.
+    const ranges = client.metadata.getTokenRanges();
+    console.log("the ring is divided into %d ranges", ranges.size);
```
wprzytula, line 73: 🔧 Describe that the example purposely uses vnodes. The top of the file seems a good fit for that.

examples/metadata/metadata-tokens.js
```diff
@@ -0,0 +1,74 @@
...
+                    " WHERE token(user_id) > ? AND token(user_id) <= ?",
+                [piece.start.getValue(), piece.end.getValue()],
+                { prepare: true },
+            );
+            if (result.rowLength > 0) {
+                console.log("  %s: %d row(s)", piece, result.rowLength);
+            }
+            totalRows += result.rowLength;
+        }
+    }
```
wprzytula: 🤔 Is it advised to perform range queries sequentially or in parallel? Here you're doing it sequentially, and I believe we could do that in parallel by using `executeConcurrent`.

wprzytula (changes requested):

examples/metadata/metadata-tokens.js
hmarszalek: Fixed: the keyspace creation now checks `supported_features` on `system.local` (via a `supportsTablets()` helper, same idea as the linked Rust driver test util) and only adds `AND tablets = {'enabled': false}` when the cluster actually reports tablet support. On Cassandra or older ScyllaDB without tablets, the clause is omitted entirely and the keyspace is vnode-based by default anyway.

examples/metadata/metadata-tokens.js
hmarszalek: Fixed: added a doc comment at the top of the file, right after the requires, explaining that the example assumes the keyspace uses vnodes on purpose, and that tablet-based keyspaces divide each table into its own ranges instead, which this example does not cover.

examples/metadata/metadata-tokens.js
hmarszalek: Good point - switched to `executeConcurrent()` (the driver's own helper, already used elsewhere in examples/): the ranges are flattened (after `unwrap()`) into one parameter array and scanned concurrently with `collectResults: true`, instead of one `await client.execute()` at a time in a loop. `resultItems` stays index-aligned with the input parameters despite the concurrency, so the per-range row counts are still correct.

examples/metadata/metadata-tokens.js
hmarszalek: Added `supportsTablets()` check.

hmarszalek: > Please add testing against Cassandra to the CI: both ITs and examples.

Is there an issue for that?

wprzytula: https://github.com/scylladb/nodejs-rs-driver/issues/244

wprzytula (approved):

PR #533, opened 2026-09-01
Schema agreement

### LINKED ISSUE #530: Support for checking schema agreement
Implement:

- `checkSchemaAgreement()` method in the `Metadata` class,
- `info.isSchemaInAgreement` in `ResultSet`,
- `maxSchemaAgreementWaitSeconds` option in `ClientOptions`.

hmarszalek:
Implements the schema agreement API: `Metadata.checkSchemaAgreement()`, a new `Metadata.waitForSchemaAgreement()`, and the two client options controling the driver's own automatic waiting.

## Public API

`Metadata`
- `checkSchemaAgreement(callback?)` – one-time check, no retry. Resolves `true` when the reachable nodes agree, `false` when they don't or when the check could not be performed. Never rejects; a failed check is reported as a `warning` log event.
- `waitForSchemaAgreement(callback?)` **(new)** – actively retries until agreement is reached or `maxSchemaAgreementWaitSeconds` elapses, then **rejects**.

`ClientOptions.protocolOptions`
- `maxSchemaAgreementWaitSeconds` – documented and defaulted since forever, but never actually forwarded anywhere. Now wired to the Rust driver's `SessionBuilder::schema_agreement_timeout`, and validated.
- `autoAwaitSchemaAgreement` **(new)** – when `false`, DDL statements return as soon as the coordinator acknowledges them instead of blocking. Callers can then check or wait explicitly. Defaults to `true`, matching both the Rust driver's default and existing behaviour.

`ResultSet`
- **Removed** `info.isSchemaInAgreement`. It was hardcoded `false` on every result. With `autoAwaitSchemaAgreement` on, the Rust driver rejects the query outright if agreement isn't reached, so a successful `ResultSet` carrying `false` is unreachable. With it off, nothing in the returned `QueryResult` records whether the statement was a schema change or whether versions converged — that lives on an internal `NonErrorQueryResponse` discarded before the public type is built.

## Tests

Integration tests run against a live 3-node cluster. Covered for both methods: agreement with all nodes reachable; agreement after DDL with one node stopped (exercising the exclusion behaviour above); and the unreachable-cluster path — where `checkSchemaAgreement()` resolves `false` with a warning log and `waitForSchemaAgreement()` rejects — each with promise and callback.

Unit tests cover option validation and the JS <-> Rust options round-trip for both new fields.

Fixes: #530 

## Pre-review checklist


lib/client-options.js
```diff
@@ -140,6 +140,10 @@ const { ExecutionProfile } = require("./execution-profile.js");
  * [TODO: Add support for this field]
  * @property {Number} [protocolOptions.maxSchemaAgreementWaitSeconds] The maximum time in seconds to wait for schema
  * agreement between nodes before returning from a DDL query. Default: 10.
+ * @property {Boolean} [protocolOptions.autoAwaitSchemaAgreement] Determines whether the driver automatically waits
+ * for schema agreement after a DDL query, retrying for up to `maxSchemaAgreementWaitSeconds`. When `false`, DDL
+ * queries return as soon as the coordinator acknowledges them, without waiting for the rest of the cluster to
+ * catch up. Default: true.
```
wprzytula, line 146: 🔧 DDL _statements_, not DDL _queries_. Queries are SELECTs. SELECTs are not DDLs.

test/integration/supported/metadata/schema-agreement-tests.js
```diff
@@ -0,0 +1,108 @@
...
+        client.metadata.checkSchemaAgreement().then((agreement) => {
+            assert.strictEqual(agreement, true);
+        }));
+
+    context("with one node stopped", function () {
+        before((done) => helper.ccmHelper.stopNode(3, done));
+        // Give the connection pool a moment to notice the node went down, so the DDL
+        // statement below is coordinated by a still-up node rather than retried against
+        // the one that was just stopped.
+        before(() => helper.delayAsync(300));
```
wprzytula: 🔧 Such timing-based tests are unacceptable. What you can instead do is periodically check if the driver has already noticed the node going down, and wait for that in a (time-bounded) loop.

wprzytula: `Node::is_connected()` in Rust Driver is a function you may find useful for this purpose.

test/integration/supported/metadata/schema-agreement-tests.js
wprzytula, line 1: @Lorak-mmk I'm summoning you because you have experience in testing schema agreement. We'd appreciate your opinion on whether this way of testing it is correct.

Lorak-mmk: >  I'm summoning you because you have experience in testing schema agreement.

I only have experience in writing flaky tests for schema agreement. Are you sure you want my opinion here?

wprzytula: Yes, definitely! You also have experience in **removing** flakiness from schema agreement tests.

wprzytula: Also, your experience will protects us from following your past mistakes.

test/integration/supported/metadata/schema-agreement-tests.js
```diff
@@ -0,0 +1,108 @@
...
+        it("should return true after a DDL statement, ignoring the stopped node", () =>
+            client
+                .execute(
+                    "CREATE KEYSPACE ks_schema_agreement_test" +
+                        " WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1}",
+                )
+                .then(() => client.metadata.checkSchemaAgreement())
+                .then((agreement) => {
+                    assert.strictEqual(agreement, true);
+                }));
```
Lorak-mmk, line 33: `check_schema_agreement` returning true immediately after a DDL depends on the implicit schema agreement wait being enabled.
Perhaps this is what you actually want, I'm just making sure you're aware of that.

test/integration/supported/metadata/schema-agreement-tests.js
hmarszalek: It would be useful to finally implement the `Host.isUp()` method. To do that, I would have to add a private `#node` field to each `Host` that holds an `Arc<Node>` to the `Node` this `Host` represents and implement `is_connected` for it. WDYT?

hmarszalek: This would allow us to use `helper.wait.forNodeDown`.

hmarszalek: Added it here, but if you ~think~ _agree, that_ we should implement the `Host.isUp()`, I would probably do it in a separate PR.

lib/host.ts
```diff
@@ -87,12 +94,10 @@ class Host extends events.EventEmitter {
...
 
     /**
-     * This endpoint is not yet implemented, and its usage will throw an error
-     *
      * Determines if the node is UP now (seen as UP by the driver).
      */
     isUp(): boolean {
-        throw new Error(`TODO: Not implemented`);
+        return this.#node.isConnected();
     }
```
wprzytula, line 103: I suggest the following docstring:

```js
    /**
     * Determines if the node is UP now and the driver has working connections to it.
     * 
     * In particular, if there's a network partition and the node is seen as UP by other nodes
     * but driver has no connections to it, then this will return `false`. 
     */
    isUp(): boolean {
        return this.#node.isConnected();
    }
```

test/integration/supported/metadata/schema-agreement-tests.js
wprzytula: > Added it here, but if you think we should implement the `Host.isUp()`, I would probably do it in a separate PR.

I don't get it. You did implement `Host.isUp()`.

And I think it merits a separate PR, too.

hmarszalek: What about `canBeConsideredAsUp`? The comment claims it is deprecated, but the error thrown is "TODO: Not implemented".

lib/host.ts
hmarszalek: Addressed in #534

test/integration/supported/metadata/schema-agreement-tests.js
wprzytula: It's your job to understand the intended difference between those two methods. Then we'll discuss its future.

hmarszalek: `canBeConsideredAsUp` was already marked as deprecated in DSx. It determined whether a specific host should be treated as online and healthy by the driver's internal connection pool, even if it was previously marked as down or had not fully completed its reconnection attempts.

hmarszalek: Continued [here](https://github.com/scylladb/nodejs-rs-driver/pull/534#discussion_r3915501381).

docs/source/migration-guide/migration-guide.md
```diff
@@ -398,6 +404,28 @@ const trace = await client.metadata.getTrace(traceId, types.consistencies.all);
...
+
+#### `ResultSet.info.isSchemaInAgreement` is removed
+
+The `cassandra-driver` reports whether the automatic wait after a DDL statement converged, as
+`isSchemaInAgreement` on `result.info`. This driver has no equivalent field: when
+`protocolOptions.autoAwaitSchemaAgreement` is enabled (the default), the driver rejects the query
+outright if agreement is not reached in time, so a successful `ResultSet` with
+`isSchemaInAgreement: false` can never exist to read; and when it is disabled, nothing in the
+result identifies whether the cluster has converged. Use `waitForSchemaAgreement()` after the query instead.
+
```
wprzytula, line 429: 🤔 I dislike that the _auto schema agreement awaiting flag_ is mentioned in a place distant from the remaining parts of schema agreement API. Can you think of a better way to describe those? I'm not saying it's possible.

hmarszalek: I could add a link to the Client options subsection in the Schema agreement subsection.

hmarszalek: And vice versa.

wprzytula: And the other direction too, please.

wprzytula (approved):

PR #534, opened 2026-09-02
Implement Host.isUp

hmarszalek:
Everything a `Host` exposed was a snapshot: the address, datacenter, rack and host id were copied out of the `Node` when the `Host` is built. Liveness cannot be snapshotted the same way. It requires holding the `Arc<Node>` for the value to be read afresh on every call.

Adds a `nodeHandle` napi struct, holding the `Arc<Node>` inside, and implements `Host.isUp` by adding a private `#node` field to the `Host`.

Remove the already deprecated `canBeConsideredAsUp` method.

### Tests

- Unit tests (test/unit/host-tests.js):
  - Updated test helpers to accept and pass node parameters
  - Added tests verifying `isUp()` correctly reflects node connectivity state
- Integration tests (test/integration/supported/metadata-host-tests.js):
  - Added test suite that stops/restarts a node in a 3-node cluster
  - Verifies `isUp()` returns false when a node is stopped
  - Verifies `isUp()` returns true when the node comes back online
  - Uses `helper.wait.forNodeDown()` and `helper.wait.forNodeUp()` for reliable state detection

## Pre-review checklist

- ~[ ] I added appropriate `Fixes:` annotations to PR description.~

### DIFF
--- a/docs/source/migration-guide/migration-guide.md
+++ b/docs/source/migration-guide/migration-guide.md
@@ -200,6 +200,11 @@ when using `client.hosts.keys()` - see issue [#282](https://github.com/scylladb/
 (they were checked in the driver tests). Those assumptions no longer hold true,
 the hosts returned from `client.hosts.keys()` may be in a random order, that may vary from run to run.
 
+### UP/DOWN distinction
+
+Previously, the Host was set as UP or DOWN as a result of received events. Now, whether a Host is considered
+as UP depends on whether it has an established open connection with the driver.
+
 #### `Host.address` is now a `net.SocketAddress`
 
 In the `cassandra-driver`, `host.address` was a plain string in the `ip:port` format.
@@ -230,6 +235,7 @@ raw `Buffer`.
   open-source Cassandra and ScyllaDB clusters, which have no DSE version to report.
 - `workloads` (getter/setter): also DSE-specific (e.g. the Search/Analytics/Graph workloads DSE
   assigns to nodes); not applicable outside DSE.
+- `canBeConsideredAsUp()`: determines if the host can be considered as UP – use `Host.isUp()` instead.
 
 `HostMap`:
 
--- a/lib/host.ts
+++ b/lib/host.ts
@@ -65,6 +65,11 @@ class Host extends events.EventEmitter {
     hostId: Uuid;
 
     /**
+     * Live handle to this node in the Rust driver's cluster state.
+     */
+    #node: rust.NodeHandle;
+
+    /**
      * Creates a new Host instance.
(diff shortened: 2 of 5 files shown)


test/unit/host-tests.js
```diff
@@ -4,13 +4,28 @@ const { SocketAddress } = require("net");
...
 const Uuid = require("../../lib/types/uuid");
 
+/**
+ * Creates a mock NodeHandle with an isConnected method that returns a fixed value.
+ */
+function makeMockNode(isConnected) {
+    return {
+        isConnected: () => isConnected,
+    };
+}
```
wprzytula: 💭 I don't really see value in these unit tests. You're merely asserting that the implementation forwards whatever `isConnected` method returns.

hmarszalek: Fair, so should I remove the unit tests?

hmarszalek: I will remove them.

lib/host.ts
hmarszalek: > What about `canBeConsideredAsUp`? The comment claims it is deprecated, but the error thrown is "TODO: Not implemented".

`canBeConsideredAsUp` was already marked as deprecated in DSx. It determined whether a specific host should be treated as online and healthy by the driver's internal connection pool, even if it was previously marked as down or had not fully completed its reconnection attempts.

wprzytula: Please illustrate cases when `canBeConsideredAsUp` and `isUp` would yield different results.

hmarszalek: When the host is still marked as down, but a retry is due, then `isUp` will report false (the Host is not up), but `canBeConsideredAsUp` will return true (the Host could be resurrected any moment).

wprzytula: What retry? Why do we believe a resurrection could happen in such case?

docs/source/migration-guide/migration-guide.md
```diff
@@ -200,6 +200,11 @@ when using `client.hosts.keys()` - see issue [#282](https://github.com/scylladb/
 (they were checked in the driver tests). Those assumptions no longer hold true,
 the hosts returned from `client.hosts.keys()` may be in a random order, that may vary from run to run.
 
+### UP/DOWN distinction
+
+Previously, the Host was set as UP or DOWN as a result of recieved events. Now, whether a Host is considered
+as UP depends on whether it has an established open connection with the driver.
```
wprzytula, line 206: typo: recieved -> received

wprzytula (changes requested):

wprzytula (approved):

PR #535, opened 2026-09-02
Add ddl helper for schema changes in tests

### LINKED ISSUE #528: Integration test flakes on concurrent Group 0 schema modification
## Why

The integration suite intermittently fails when ScyllaDB rejects a schema change with:

`Failed to apply group 0 change due to concurrent modification`

Observed in [PR #526 CI](https://github.com/scylladb/nodejs-rs-driver/actions/runs/33134710035/job/98731986243) while running `Metadata#getKeyspace() > altering a keyspace should result in a cache refresh`.

## What

Track and mitigate this flaky integration failure. The underlying ScyllaDB problem is tracked by [scylladb/scylladb#21779](https://github.com/scylladb/scylladb/issues/21779) and `SCYLLADB-3447`.

hmarszalek:
Add and use ddl helper, which retries the schema altering statement when a conflict happens.

Heavily inspired by:
https://github.com/scylladb/python-rs-driver/pull/170
and
https://github.com/scylladb/scylla-rust-driver/blob/ab3d56eab36c0d251a872fb162af55f5b67c0843/scylla/tests/integration/utils.rs#L217-L331

Fixes: #528 

## Pre-review checklist

### DIFF
--- a/test/integration/broken/client-metadata-tests-long.js
+++ b/test/integration/broken/client-metadata-tests-long.js
@@ -185,8 +185,9 @@ describe("Client", function () {
                         ));
 
                 it("should return false when executing DDL queries", () =>
-                    client
-                        .execute(
+                    helper
+                        .ddl(
+                            client,
                             "CREATE KEYSPACE ks_rs_is_schema_in_agreement" +
                                 " WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor' : 1}",
                         )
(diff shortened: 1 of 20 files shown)


wprzytula (changes requested):
1. This only implement retries, which is not enough. Both Rust Driver and Python RS Driver implement special LBP, too, to guarantee stability of the queried targets.
2. This implement retries out of the client. Both reference drivers use a custom Retry Policy for that.

❓ Why didn't you follow the reference impls?

test/integration/broken/client-metadata-tests-long.js
wprzytula, line 1: ❓ `test/integration/broken/client-metadata-tests-long.js` and `test/integration/broken/metadata-tests.js`:
What's still missing in the driver to enable these tests? Perhaps it's possible to already enable them at least partly?

wprzytula (approved):

PR #536, opened 2026-09-03
Add Tests feature

hmarszalek:
Hide all rust test helpers by the tests feature (previously the client-routes-proxy-tests feature).

Update `unit-tests.yml` to build for testing. 

On release, we verify everything works by running examples instead of the full test suite, which requires the `tests` feature to be enabled.

## Pre-review checklist


wprzytula (approved):
Are we sure that CI fails if some error is encountered when running examples?

hmarszalek:
> Are we sure that CI fails if some error is encountered when running examples?

It will, after #538 is merged.

PR #539, opened 2026-09-03
Update the Rust driver version to 1.8.0

hmarszalek:
Update the Rust driver version to 1.8.0

## Pre-review checklist


wprzytula (changes requested):
You only bumped `Cargo.toml`. What about `Cargo.lock`?

1. If we have it versioned, but you forgot to do that, please do.
2. If we don't version it, please add it to the repo in a new commit.
    Cargo.lock should be versioned unless the project is a library used in another Rust project — and Rust wrapper over Rust Driver is used a plugin to Node, which is **not** another Rust project.

hmarszalek:
> You only bumped Cargo.toml. What about Cargo.lock?

Cargo.lock is in .gitignore.

wprzytula (approved):

PR #541, opened 2026-09-07
Support for getReplicas

### LINKED ISSUE #531: Support for getReplicas
Reimplement the `getReplicas(keyspaceName, token)` method in the `Metadata` class. Change the signature to be tablet-aware (tablets have per-table replica placement). 

```ts
getReplicas(keyspaceName: string, token: Buffer | token.Token | token.TokenRange): Host[];
getReplicas(keyspaceName: string, tableName: string, token: Buffer | token.Token | token.TokenRange): Host[];
```

hmarszalek:
Implements the `getReplicas(keyspaceName, tableName, token)` endpoint to retrieve which nodes replicate a given partition, along with comprehensive documentation for cluster topology features.

## Changes

**Replicas endpoint**: Returns `Replica[]` (host + shard pairs) for a partition
  - Throws an error if the keyspace or table is not found (instead of returning null)
  - Accepts three input types: `Buffer`, `Token`, or `TokenRange`
  - Respects both vnode-based and tablet-based routing

### Documentation
- **Replicas guide** (`metadata.md`): How to look up replicas, three ways to specify partitions, replica object structure, host identity comparison
- **Topology guide** (NEW `topology.md`): Host properties, HostMap collection, datacenter/rack awareness
- **Migration guide**: Explains that `getReplicas()` key differences

### Testing
- 9 comprehensive integration tests covering:
  - Replica count matches replication factor
  - No duplicate nodes in replicas
  - Host identity consistency with `client.hosts`
  - Consistent placement across repeated calls
  - Vnode and tablet routing modes

Fixes: #531 

## Pre-review checklist

### DIFF
--- a/docs/source/migration-guide/migration-guide.md
+++ b/docs/source/migration-guide/migration-guide.md
@@ -298,6 +298,25 @@ const token = client.metadata.newToken(Buffer.from("key"), "ks", "tbl");
 const token = new (require("@scylladb/driver").token.Token)(12345n); // wrap a raw value directly
 ```
 
+### Replicas
+
+`getReplicas(keyspaceName, tableName, token)` returns an array of `Replica` objects (node + shard
+pairs) that replicate a partition, given a keyspace, table, and token. See the
+[Replicas](../metadata/replicas.md) page for the full description of the new API.
+
+```javascript
+const replicas = client.metadata.getReplicas("my_keyspace", "my_table", token);
+
+for (const replica of replicas) {
+  console.log(replica.host.address, "shard", replica.shard);
+}
+```
+
+**Key differences from similar methods in `cassandra-driver`:**
+- The token parameter can be a `Token` or a `TokenRange`
+- Each replica includes both the host and its shard number
+- The replica's `host` is the same object instance as `client.hosts.get(hostId)`, so you can compare by identity
+
 ### Schema
 
 The schema metadata API was rewritten from scratch, and almost nothing carries over unchanged.
--- a/lib/metadata/index.ts
+++ b/lib/metadata/index.ts
@@ -10,7 +10,7 @@ import { EmptyCallback, ValueCallback } from "../..";
 // @ts-ignore
 import promiseUtils = require("../promise-utils");
 import { Token, TokenRange, minTokenRange } from "../token";
-import { Host } from "../host";
+import { Host, Replica } from "../host";
 import types = require("../types");
 import { ColumnInfo } from "../types/cql-utils";
 
@@ -76,19 +76,18 @@ class Metadata {
     }
 
     /**
-     * Gets the host list representing the replicas that contain the given partition key, token or token range.
+     * Gets the replicas that contain the given partition key, token or token range.
      *
-     * It uses the pre-loaded keyspace metadata to retrieve the replicas for a token for a given keyspace.
-     * When the keyspace metadata has not been loaded, it returns null.
+     * A replica is the shard of a node the partition lives on, paired with that node.
      * @param {string} keyspaceName Name of the keyspace.
      * @param {Buffer | Token | TokenRange} token Can be Buffer (serialized partition key),
      * Token or TokenRange.
-     * @returns {Host[]} The replicas.
+     * @returns {Replica[]} The replicas.
      */
     getReplicas(
         keyspaceName: string,
         token: Buffer | Token | TokenRange,
-    ): Host[] {
+    ): Replica[] {
         throw new Error("TODO: Not implemented");
     }
 
--- a/lib/metadata/index.ts
+++ b/lib/metadata/index.ts
@@ -79,15 +79,29 @@ class Metadata {
      * Gets the replicas that contain the given token or token range.
      *
      * A replica is the shard of a node the partition lives on, paired with that node.
+     * The hosts returned are the very same objects the {@link Client#hosts} map holds,
+     * so a replica's node can be compared against a host of the cluster by identity.
      * @param {string} keyspaceName Name of the keyspace.
+     * @param {string} tableName Name of the table the token belongs to or empty string if
+     * the table is irrelevant.
      * @param {Token | TokenRange} token Token or TokenRange.
      * @returns {Replica[]} The replicas.
      */
     getReplicas(
         keyspaceName: string,
+        tableName: string,
         token: Token | TokenRange,
     ): Replica[] {
-        throw new Error("TODO: Not implemented");
+        if (token instanceof TokenRange) {
+            // A range is replicated as a whole, so any token it contains identifies its replicas.
+            // The end token is the one it always contains, being its inclusive end.
+            token = token.end;
+        }
+        return this.#rustClient.getReplicas(
+            keyspaceName,
+            tableName,
+            token.getValue(),
+        );
     }
 
     /**
--- a/lib/metadata/index.ts
+++ b/lib/metadata/index.ts
@@ -76,17 +76,16 @@ class Metadata {
     }
 
     /**
-     * Gets the replicas that contain the given partition key, token or token range.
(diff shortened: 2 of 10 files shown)


lib/metadata/index.ts
```diff
@@ -76,19 +76,18 @@ class Metadata {
...
      * @param {Buffer | Token | TokenRange} token Can be Buffer (serialized partition key),
      * Token or TokenRange.
-     * @returns {Host[]} The replicas.
+     * @returns {Replica[] | null} The replicas or `null` if the keyspace or table does not exist.
      */
     getReplicas(
         keyspaceName: string,
         token: Buffer | Token | TokenRange,
-    ): Host[] {
+    ): Replica[] | null {
```
wprzytula: 🔧 We discussed that `getReplicas` should throw on unknown keyspace/table; null should never be returned.

test/typescript/metadata-tests.ts
```diff
@@ -16,10 +16,11 @@ async function myTest(): Promise<any> {
     let promise: Promise<void>;
     let n: number;
     let hosts: Host[];
+    let replicas: Replica[] | null;
 
     promise = client.connect();
 
-    hosts = client.metadata.getReplicas("ks1", Buffer.from([0]));
+    replicas = client.metadata.getReplicas("ks1", Buffer.from([0]));
```
wprzytula: ♻️ `hosts` variable looks stray.

src/metadata/state.rs
```diff
@@ -76,6 +93,47 @@ impl ClusterSnapshot {
...
+                    })?.ok_or_else(|| {
+                        ConvertedError::from(make_js_error(format!(
+                            "Node {} replicates {token:?} but is not part of the cluster state it was read from",
+                            node.host_id
+                        )))
+                    })?;
+                build_replica(env, host, shard).map_err(ConvertedError::from)
+            })
+            .collect::<ConvertedResult<Vec<_>>>()
+    }
```
wprzytula, line 129: ⚠️ TOCTOU here: `get_keyspace()` may return `Some`, but in the meantime a metadata refresh may swap the `Arc<ClusterState>`, resulting in `get_token_endpoints` not finding the looked up keyspace/table.

src/metadata/state.rs
```diff
@@ -76,6 +93,41 @@ impl ClusterSnapshot {
...
+    /// The replicas of `token` of the given table: each the shard of a node the partition lives
+    /// on, paired with the very same JS `Host` object this snapshot's `HostMap` holds.
+    ///
+    /// If passed keyspace is unknown, this falls back to Strategy::SimpleStrategy with
+    /// replication_factor = 1, walks the global ring from token and returns the first node it finds.
+    ///
+    /// If keyspace exists and uses tablets, but the table doesn't exist, this uses keyspace's real
+    /// replication strategy over the vnode ring (tablets aren't considered since they can't be
+    /// looked up).
+    pub(crate) fn replicas<'env>(
```
wprzytula, line 105: 💭 I'm not sure if this description is factually correct, perhaps it is. But what I am sure of is that this must be eventually improved — both in Rust Driver and here.

1. Make sure the Rust Driver's relevant issue mentions all problems computing replicas there has wrt unknown keyspace/table.
2. Open an issue in this driver's repo to cover the same shortcomings.

---------------------------------

Nonsense behaviour:
- > If passed keyspace is unknown, this falls back to Strategy::SimpleStrategy with replication_factor = 1, walks the global ring from token and returns the first node it finds.

    This could have had sense when vnodes where the only option, because any replication strategy with any replication factor agreed about the first replica. This is no longer true for tablets, so this trick doesn't work — yet confuses and misleads!

- > If keyspace exists and uses tablets, but the table doesn't exist, this uses keyspace's real replication strategy over the vnode ring

    We already **know** that we use tablets, not vnodes, so we consciously return incorrect results.


If the Rust Driver's issue doesn't yet have these observations, please put them there.

lib/metadata/index.ts
```diff
@@ -79,15 +79,29 @@ class Metadata {
      * Gets the replicas that contain the given token or token range.
      *
      * A replica is the shard of a node the partition lives on, paired with that node.
+     * The hosts returned are the very same objects the {@link Client#hosts} map holds,
+     * so a replica's node can be compared against a host of the cluster by identity.
      * @param {string} keyspaceName Name of the keyspace.
+     * @param {string} tableName Name of the table the token belongs to or empty string if
+     * the table is irrelevant.
```
wprzytula, line 86: :recycle: 

> the table is irrelevant

This is enigmatic. Better be explicit: _For tablets, this is absolutely necessary. For vnodes, the table is irrelevant, so you can pass an empty string as well_.

docs/source/migration-guide/migration-guide.md
```diff
@@ -298,6 +298,25 @@ const token = client.metadata.newToken(Buffer.from("key"), "ks", "tbl");
...
+
+for (const replica of replicas) {
+  console.log(replica.host.address, "shard", replica.shard);
+}
+```
+
+**Key differences from similar methods in `cassandra-driver`:**
+- The token parameter can be a `Token` or a `TokenRange`
+- Each replica includes both the host and its shard number
+- The replica's `host` is the same object instance as `client.hosts.get(hostId)`, so you can compare by identity
```
wprzytula, line 318: ⛏️ Missing punctuation.

wprzytula (changes requested):
So close!

src/metadata/state.rs
```diff
@@ -76,6 +93,41 @@ impl ClusterSnapshot {
...
 
+    /// The replicas of `token` of the given table: each the shard of a node the partition lives
+    /// on, paired with the very same JS `Host` object this snapshot's `HostMap` holds.
+    ///
+    /// If passed keyspace is unknown, this falls back to Strategy::SimpleStrategy with
+    /// replication_factor = 1, walks the global ring from token and returns the first node it finds.
+    ///
+    /// If keyspace exists and uses tablets, but the table doesn't exist, this uses keyspace's real
+    /// replication strategy over the vnode ring (tablets aren't considered since they can't be
+    /// looked up).
```
Lorak-mmk, line 104: Would be good to add a `TODO` to update this comment when Rust Driver is fixed.

lib/metadata/index.ts
```diff
@@ -79,15 +79,30 @@ class Metadata {
...
         keyspaceName: string,
+        tableName: string,
         token: Token | TokenRange,
     ): Replica[] {
-        throw new Error("TODO: Not implemented");
+        if (token instanceof TokenRange) {
+            // A range is replicated as a whole, so any token it contains identifies its replicas.
+            // The end token is the one it always contains, being its inclusive end.
+            token = token.end;
+        }
```
Lorak-mmk, line 99: Is there a guarantee that `TokenRange` belongs to a single partition? If so, how do you guarantee it?
If not, won't this silently return wrong results?

What did the original driver do in this situation?

Lorak-mmk (commented):
Generally looks good, I don't think Wojciech will have many more comments.

lib/metadata/index.ts
hmarszalek: There is no unconditional guarantee – `TokenRange` is just a `{start, end}` pair. This mirrors the original driver, which did the exact same thing.

src/metadata/state.rs
```diff
@@ -58,16 +78,58 @@ pub(crate) struct ClusterSnapshot {
...
+    ///
+    /// TODO: update this comment when Rust Driver is fixed.
```
Lorak-mmk: Another person reading this in the future won't know what issue you are talking about. Please link it.

Lorak-mmk (approved):
I see no more problems, but lets wait for Wojciech with merging

PR #542, opened 2026-09-10
Metadata documentation

hmarszalek:
Rename `metadata/metadata.md` to `metadata/schema.md`. Add a documentation page for topology under `metadata.topology.md`

## Pre-review checklist

- [ ] I have split my patch into logically separate commits.
- [ ] All commit messages clearly explain what they change and why.
- [ ] All commits compile, pass static checks and pass tests.
- [ ] PR description sums up the changes and reasons why they should be introduced.


docs/source/metadata/topology.md
```diff
@@ -0,0 +1,86 @@
...
+```javascript
+// All three are equivalent:
+const host = client.hosts.get(hostId);
+const host = client.hosts.get(socketAddress);
+const host = client.hosts.get("192.168.1.1:9042");
+```
+
+### Staleness
+
+`client.hosts` is a snapshot of the cluster's topology at the moment it was retrieved. The actual set of hosts and their properties (address, datacenter, rack, connectivity status) can change at any time as the driver detects topology changes, so always read it fresh from the client when you need current information. Hosts returned from a single `client.hosts` call are mutually consistent, but separate calls may see different cluster states if a topology change occurs between them.
```
wprzytula: 🔧 
```suggestion
`client.hosts` is a snapshot of the cluster's topology at the moment it was retrieved. The actual set of hosts and their properties (address, datacenter, rack, connectivity status) can change at any time as the driver detects topology changes, so always read it fresh from the client when you need current information. Hosts returned from a single `client.hosts` access are mutually consistent, but separate accesses may see different cluster states if a topology change occurs between them.
```

Functions are called, fields are accessed. Language surface-related remark.

wprzytula (dismissed):

wprzytula (dismissed):

wprzytula (approved):

PR #547, opened 2026-09-17
Bump the test Scylla version with Renovate

hmarszalek:
Pin the ScyllaDB version used by the docker compose cluster
- Introduce `scylla_version.env` in the repository root as the single source of truth for the version used in testing.
- The cluster definition moves to `.github/scylla-cluster.yml`, and `.github/docker-compose.yml` becomes an include of it with that env file.

Take the default CCM version from `scylla_version.env` in `test-helper.js`
- The version CCM creates its clusters with was hardcoded in `test-helper.js`, separately from the one the docker compose cluster used, so the two drifted apart with nothing to keep them in sync.
- Read it from `scylla_version.env` instead, prefixed with `release:`.
- CCM_VERSION still overrides it.
- The value is validated to be a full, three-component version.

Keep the ScyllaDB version up to date with Renovate

## Pre-review checklist


scylla_version.env
```diff
@@ -0,0 +1,15 @@
...
+# This file is kept up to date automatically by Renovate.
+#
+# The value must be a full, three-component version (e.g. 2026.2.2), not a
+# truncated one (e.g. 2026.2). A truncated version is accepted by CCM, but it
+# makes every single ccm invocation - even against an already-running cluster -
+# query AWS to resolve the latest patch number, which slows the tests down a lot.
+#
+# This file may only contain `#` comments and bare KEY=VALUE lines: no `export`,
+# no shell syntax and no quotes around the value.
+SCYLLA_VERSION=2026.2.2
```
Lorak-mmk: 2026.3.0 is already released.

test/test-helper.js
```diff
@@ -1023,7 +1056,7 @@ const helper = {
             isScylla: isScylla,
             version:
                 process.env["CCM_VERSION"] ||
-                (isScylla ? "release:2026.2.2" : "3.11.4"),
+                (isScylla ? `release:${getDefaultScyllaVersion()}` : "3.11.4"),
```
Lorak-mmk, line 1059: Nit: if this is called often (for example once per test instead of once per whole test run) you could consider caching the result somehow to avoid reading the same file hundreds of times for no reason.

hmarszalek: `getDefaultScyllaVersion()` already caches into the module-level `defaultScyllaVersion` variable and only calls `fs.readFileSync` once per process.

Lorak-mmk (approved):

PR #548, opened 2026-09-18
Update the Rust drvier version to 1.9.0

hmarszalek:
Update the Rust drvier version to 1.9.0.

`configure_ssl` now returns an `OpenSsl010Config` rather than a built `SslContext`. The `SslContext` variant of `TlsContext` is deprecated since 1.9.0 and cannot support TLS session tickets.

The function uses `SslConnector::builder` rather than a bare `SslContextBuilder`, for its safer defaults. It derefs to `SslContextBuilder`, so it has the same methods.

## Pre-review checklist


Cargo.toml
```diff
@@ -10,7 +10,7 @@ crate-type = ["cdylib"]
 
 napi = { version = "3.6.0", default-features = false, features = ["napi4", "napi6", "async"] }
 napi-derive = "3.4.0"
-scylla = { version = "1.8.0", features = ["num-bigint-03", "openssl-010", "unstable-nodejs-rs", "unstable-client-routes"] }
+scylla = { version = "1.9.0", features = ["num-bigint-03", "openssl-010", "unstable-nodejs-rs", "unstable-client-routes"] }
 tokio = { version = "1.34", features = ["full"] }
 futures = "0.3"
 uuid = "1"
```
Lorak-mmk, line 16: In 1.9.0 we deprecated `TlsContext::Openssl010`.
Even if the deprecation warning doesn't show automatically (because you use `Into<TlsContext> for SslContext`), you should update the code.

The change is small: your `configure_ssl` needs to return `ConvertedResult<Option<SslContextBuilder>>` instead of `ConvertedResult<Option<SslContext>>`. Then you can use https://docs.rs/scylla/latest/scylla/client/session/struct.OpenSsl010Config.html#method.from_dangerous_builder to convert it to `OpenSsl010Config`, which you can pass to `builder.tls_context`.

A better solution would be to use `SslConnectorBuilder` from openssl. It has the same methods as SslContextBuilder (it derefs to it), but ut has safer defaults - it verifies certs by default and sets a reasonable cipher list. You'd need to make sure that its usage is ok with this driver.

Lorak-mmk (changes requested):

Cargo.toml
hmarszalek: Should I do all changes in one PR or fix it in a follow up?

Lorak-mmk: Imo this is strongly connected to the update of Rust Driver, so it should be done here.

In both cases (minimal changes, or migrationto SslconnectorBuilder) the changes will be small (in terms of changed lines of code).

hmarszalek:
Does this change resolve https://scylladb.atlassian.net/browse/DRIVER-443? It references https://github.com/scylladb/scylla-drivers/issues/56.

Lorak-mmk (approved):

Lorak-mmk:
> Does this change resolve https://scylladb.atlassian.net/browse/DRIVER-443? It references https://github.com/scylladb/scylla-drivers/issues/56.

In general yes. However, current Rust Driver implementation of TLS tickets does not work very well with TLS1.3: https://rust-driver.docs.scylladb.com/stable/connecting/tls.html#tls-tickets-limitations
This limitation should be mentioned in the docs, and an issue opened to address it in the future after we address it in Rust Driver.

